PR #1, opened 2017-12-14
Switch to the typescript generated client.

brendandburns:
* Add examples for how to use typescript from javascript.
   * Update documentation.
   * Add in the typescript client from https://github.com/kubernetes-client/typescript

Will deprecate the typescript repository once this is merged.


brendandburns:
@mbohlool

mbohlool:
@brendandburns I would be fine with this solution. Another approach we can take is to rename typescript client to javascript and have typescript and javascript examples there. Typescript would be the language we use to implement javascript client. I like this approach more. WDYT?

brendandburns:
That's fine with me too, I will do that PR.

Thanks
--brendan



________________________________________

brendandburns:
Ok, PR updated to add in the typescript client here.

I will deprecate the typescript-only repo once this PR is merged.

brendandburns:
@mbohlool friendly monday morning ping (assuming you're not on vacation... If you are, please ignore :)

mbohlool:
LGTM.

Issue #13, opened 2018-02-21
Publishes `.ts` files breaks build in all projects using `strict` option


hausdorff:
Publishing `.ts` files will cause builds to fail on all projects that are using TypeScript's strict null checking, even in the with `--skipLibCheck`, `--skipDefaultLibCheck`, and the `@kubernetes/client-node` file placed in the `excludes` list.

Instead, you probably want to:

* Apply the following diff to the `package.json`, so that the `dist/index.js` is properly targeted and the typings file `dist/index.d.ts` is included:
  ```patch
  @@ -10,7 +10,8 @@
       "*.ts",
       "*.js"
     ],
  -  "main": "index.js",
  +  "main": "dist/index.js",
  +  "types": "dist/index.d.ts",
     "scripts": {
       "clean": "rm -Rf node_modules/ dist/",
       "build": "tsc",
  ```
* Increment the version to `0.1.4`.
* Run `npm publish` using npm client > 5. (4 will probably not work; I was able to publish a mirror [here](https://www.npmjs.com/package/@hausdorff/client-node) using `5.7.2`.

hausdorff:
Additionally, I'm hoping we can emit comments as part of the declaration as well. In the present form, the NPM package has the comments embedded in the TS code because we're shipping the TS code itself. If we do choose to go the route of generating `.d.ts` files, then we'll lose comments because the `.tsconfig` contains sets the option `"removeComments": true,`. Simply flipping this to `true` should fix the problem.

brendandburns:
Closing this since (I believe) the fix is now merged. Please re-open if this isn't true.

omar:
Running into a similar issue using TypeScript 2.6.2 the tsconfig.json file below. I don't have time at the moment to do a full repro, but I was getting the error below when compiling my own application. To get it to compile, I had to go into `node_modules/@kubernetes/client-node` and delete `index.ts`.

Here's the offending line that pulls `index.ts` into the package process https://github.com/kubernetes-client/javascript/pull/17/files#diff-b0f7d8dfb4157a255e97dc798e851c5dR12.

Another issue is that I still don't get proper typing support in Visual Studio Code for this package. I'm not sure why that is the case, but I'm a bit of a rush so I'm gonna need to go use another library to hit the API.

**tsconfig.json**:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2015",
    "moduleResolution": "node",
    "sourceMap": true,
    "rootDir": ".",
[...]

PR #17, opened 2018-03-20
Update library to 1.9.x

brendandburns:


hausdorff:
Any chance we could also apply the trivial changes I suggest in #13 to fix `--strict` incompatibility while we're at it? It should be 2 lines in `package.json`.

brendandburns:
@hausdorff Does that look right?

Thanks
--brendan

hausdorff:
@brendandburns Unless I'm missing something, I think it should be prefaced by a `dist/`, _e.g._, `"dist/index.js"`:

So, altogether:

```patch
@@ -10,7 +10,8 @@
     "*.ts",
     "*.js"
   ],
-  "main": "index.js",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
     "clean": "rm -Rf node_modules/ dist/",
     "build": "tsc",
```

brendandburns:
Ok, I fixed it. There were some problems with the `package.json` and building the npm also, this all seems to work correctly now...

PR #18, opened 2018-03-22
fix(auth): wrong case for auth-provider

VEBERArnaud:
Hi, 
I receive a HTTP 403, every time I try to use the client from `k8s.Config.defaultClient()`

After some researches, I think the `applyToRequest` use a wrong case for `authProvider` based on https://github.com/kubernetes-client/javascript/blob/master/node-client/src/config_types.ts#L68


brendandburns:
LGTM, thanks.

PR #26, opened 2018-03-30
Make use of types for websocket library

cjlarose:
This fixes a compilation error I was getting when trying to import this library. The generated `exec.d.ts` file looked like this:

```typescript
/// <reference types="node" />
import stream = require('stream');
import { WebSocketHandler } from './web-socket-handler';
import { KubeConfig } from './config';
export declare class Exec {
    'handler': WebSocketHandler;
    constructor(config: KubeConfig);
    exec(namespace: string, podName: string, containerName: string, command: string, stdout: stream.Writable | any, stderr: stream.Writable | any, stdin: stream.Readable | any, tty: boolean): Promise<WebS
ocket>;
}
```

which produced this error

```
../../kubernetes-js-client/node-client/dist/exec.d.ts(8,201): error TS2304: Cannot find name 'WebSocket'.
```

I suspect that these changes also fix an error related to HTTP Basic Auth when using websockets connections because the `auth` property on the WebSocket connection was being set to an object instead of a `${username}:${password}` string as NodeJS's HTTPS client expects.


brendandburns:
@cjlarose thanks for the PR, I just merged #39 so this needs a rebase, but then I think we can merge this...

cjlarose:
@brendanburns I think that #39 made a bunch of the same changes as this PR.

One notable difference is that #39 removed the use of `this.config.applyToRequest` when making a secure WebSockets connection, so some TLS options appear to no longer be passed (`ca`, `cert`, `key`). I don't know if those options are important or not. If they are, I think it's a regression, but regardless, it's a separate issue from this original PR.

sheldonkwok:
Yeah, I just noticed that as well. I created a new pull request here: https://github.com/kubernetes-client/javascript/pull/41

Rebasing yours works for me as well.

cjlarose:
@sheldonkwok I like yours better. Let's go with #41.

PR #41, opened 2018-04-25
Fix websocket tls options 

sheldonkwok:


node-client/src/config.ts
```diff
@@ -106,22 +107,25 @@ export class KubeConfig {
...
 
-        if (cluster.skipTLSVerify) {
-            opts.strictSSL = false
-        }
         opts.ca = this.bufferFromFileOrString(cluster.caFile, cluster.caData);
         opts.cert = this.bufferFromFileOrString(user.certFile, user.certData);
         opts.key = this.bufferFromFileOrString(user.keyFile, user.keyData);
+    }
+
+    private applyAuthOptions(opts: request.Options | https.RequestOptions) {
```
cjlarose: `applytoHTTPsOptions` and `applyToRequest` both add some form of "auth" by using HTTP Basic Auth.

Consider renaming this function `applyAuthorizationHeader` to indicate its limited scope.

Alternatively, consider modifying `applyAuthOptions` to apply HTTP Basic Auth to either  type of input. If it's a `https.RequestOptions`, we'll want to 

```javascript
opts.auth = `${user.username}:${user.password}`;
```

If it's a `request.Options`, we'll want to 

```
opts.auth = { username: user.username, password: user.password }
```

though I'm not sure how to convince the compiler that you can safely do this.

sheldonkwok: We could dig around for a fixed key in request.Options to create a type guard assertion but I changed the name for now.

brendandburns:
This generally looks ok to me, but I'd love some unit tests for the new behavior...

brendandburns:
LGTM, thanks!

PR #54, opened 2018-06-07
fix: parse insecure-skip-tls-verify correctly

edvald:
A simple fix and test. The YAML parser will return the value as a boolean, so the comparison was incorrect.


brendandburns:
@edvald Thanks for the PR! we need a signed CLA to merge.

edvald:
Sure thing! I signed it last week, should be all set.

brendandburns:
Thanks, LGTM!

PR #4, opened 2018-01-08
Some improvements for the overall project layout (see #3)

svenefftinge:
- renamed ‘node’ to ‘node-client’ to better match the package name
 - moved source code to src folder
 - removed some unneeded libs from package.json
 - fixed some imports
 - updated .gitignore

The test is failing, because of a missing test file.

PR #57, opened 2018-07-01
Regenerate and Refactor for v1.11

flyboarder:


brendandburns:
Needs rebase because I merged some older PRs...

Note that everything moved into the `src/` directory here:

https://github.com/kubernetes-client/javascript/pull/4

It's not entirely clear to me what the canonical layout for a JS/TS library should be..

Do you have any pointers to a widely use standard layout? Thanks!

@svenefftinge in case there are comments about why we moved to 'src' in the first place...

flyboarder:
@brendanburns the `kubernetes-client/gen` repo has a suggestion for how to structure client projects:
```
Recommended structure is to generate client in a folder called "kubernetes" at the root of the client repo and put all settings in a file named "settings" at the root of the repo. If you followed these recommendations, you can simply run autoupdate script anywhere inside the client repo:
```

The other clients also do this:
https://github.com/kubernetes-client/haskell
https://github.com/kubernetes-client/java
https://github.com/kubernetes-client/python

brendandburns:
I guess my question was more around how javascript libraries in general are laid out. I don't care about aligning with other clients as much as I care about aligning with what Node/JS developers expect...

PR #58, opened 2018-07-11
Add loadFromCluster to KubeConfig object, and an example of how to use.

### LINKED ISSUE #55: How to Watch with KubeConfig configured from cluster
When creating applications which needs to `Watch` the kubernetes api, there is no way of creating a new `KubeConfig` object created by credentials provided by kubernetes (or openshift). 

Basically the `Config::fromCluster()` is what we need, but then available on the `KubeConfig` class like

```js
kc.fromCluster();  // <-- this is missing and so required!
``` 

Or am I doing something completely wrong here? 
It's just that the application which needs to query the kubernetes api, is running in kubernetes itself, so `.kube/config` files are not working in this usecase.

brendandburns:
Fixes #55

### DIFF
--- a/examples/in-cluster.js
+++ b/examples/in-cluster.js
@@ -1,0 +1,16 @@
+const k8s = require('@kubernetes/client-node');
+
+let kc = new k8s.KubeConfig();
+kc.loadFromCluster();
+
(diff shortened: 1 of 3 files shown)


southwolf:
The `0.4.0` has not been published to npm so people cannot install it. Could you make a release? Thanks!

PR #59, opened 2018-07-12
Fix for running in electron environment

mkramb:
I'm using `kubernetes-client` in `electron` environment (in main/node renderer) and the underlying library `shelljs` has issues when running as an electron package application. More information here:
- https://github.com/shelljs/shelljs/issues/480
- https://github.com/shelljs/shelljs/wiki/Electron-compatibility

Based on testing the cleanest solution is setting exec to async mode, which doesn't have this problem. I'm using this in https://github.com/mkramb/jx-tray which is a start of https://jenkins-x.io desktop client.

No changes in functionality & existing tests are passing.

### DIFF
--- a/node-client/src/config.ts
+++ b/node-client/src/config.ts
@@ -135,18 +135,23 @@ export class KubeConfig {
                             cmd = cmd + ' ' + config['cmd-args'];
                         }
                         // TODO: Cache to file?
-                        const result = shelljs.exec(cmd, { silent: true });
-                        if (result['code'] != 0) {
-                            throw new Error('Failed to refresh token: ' + result.stderr);
-                        }
-                        let resultObj = JSON.parse(result.stdout.toString());
+                        const command = shelljs.exec(cmd, { silent: true, async: true });
 
-                        let path = config['token-key'];
-                        // Format in file is {<query>}, so slice it out and add '$'
-                        path = '$' + path.slice(1, -1);
+                        command.stdout.on('data', (data) => {
+                            let resultObj = JSON.parse(data.toString());
+                            let path = config['token-key'];
 
-                        config['access-token'] = jsonpath.query(resultObj, path);
-                        token = 'Bearer ' + config['access-token'];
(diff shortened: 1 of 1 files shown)


brendandburns:
@mkramb we need you to sign the CNCF CLA.

Thanks!

brendandburns:
@mkramb does this actually work? This is changing the call from being synchronous to being asynchronous and I worry that there is a race between token update and web request generation...

(separately I'm having similar issues for the Azure provider, so we do need to deal with async clients somehow here...)

mkramb:
I have signed CLA.

It works (tested in electron environment) and also existing tests are passing, also the attached issue - if you look at the last comment from the ShellJS maintainer, it looks like best solution ATM.

https://github.com/shelljs/shelljs/issues/480

brendandburns:
This needs a rebase.

I still am not confident that this will work correctly given that it is switching the call from sync to async, it feels racy to me... (and I'm not positive the tests cover that particular part of the code, sorry!)

Any chance that you can conditionalize this so that it only works this way when running in electron?

Thanks!

itowlson:
Would a safe way forward on this be to create `async` versions of the `applyToHTTPSOptions` and `applyToRequest` functions, and call the async version of `shelljs.exec` only from within those async forms?  Then Electron apps could call the `applyToHTTPSOptionsAsync` and `applyToRequestAsync` methods, while vanilla Node applications could call either the sync or async methods.

Most places within the code that call `applyToHTTPSOptions` or `applyToRequest` seem to be returning Promises, so could call the async versions.

It would be a much larger piece of work though (and not lacking in risk either).

edvald:
This doesn't actually work, it just swallows the error because the function returns immediately, and doesn't have an error event handler. I would suggest switching from `shelljs` to https://github.com/sindresorhus/execa instead (and making the enclosing function async along the way), since none of the other features of shelljs are being used. I can make a quick PR for that if you'd like.

PR #63, opened 2018-07-20
Fix node deprecation warning on Buffer constructor

dvallin:
Node 10 currently gives this deprecation warning

> (node:12492) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

So this pull request replaces all calls to Buffer::new by Buffer.from


brendandburns:
LGTM.

PR #68, opened 2018-08-02
fix(websockethandler): fix buffer init in handleStandardInput

cmosgh:
With this fix you can use a local readable stream as stdin on remote pod.


cmosgh:
Dunno how to move forward, followed the instructions on the linuxfundation page and still won't work

PR #69, opened 2018-08-02
fix(websockethandler): fix buffer init in handleStandardInput

cmosgh:


brendandburns:
Thanks for the contribution, we need a signed CLA.

Thanks!

cmosgh:
@brendanburns can you merge this and somehow make it available for npm install 😄  ?

brendandburns:
Thanks for the CLA, merged and will cut a new release today...

brendandburns:
@cmosgh published as 0.5.0 looks like it takes a while to become public but it should be there once npm.org catches up....

PR #73, opened 2018-08-03
Add ProtoBuf files for Javascript.

brendandburns:
Initial commit, no concrete support yet, but files if people want to use them.


itowlson:
This looks fine as far as it goes, but a couple of suggestions:

* Is there a reason for generating JavaScript files rather than TypeScript?
* The README or CONTRIBUTING file should document how to regenerate these files.
* If these don't end up in the `@kubernetes/client-node` package then what is the value of having them here - is there some way for consumers to import them directly from the repo?  If so perhaps update the README to document that?

PR #77, opened 2018-08-13
fix `we need a path` error

cbuchacher:
Commit 4ef92544 (Add tslint validation.) renames path to pathKey, but misses
one usage of the variable. This leads to the following error:

```
AssertionError [ERR_ASSERTION]: we need a path
    at JSONPath.query (...\node_modules\jsonpath\lib\index.js:92:10)
    at KubeConfig.applyAuthorizationHeader (...\node_modules\@kubernetes\client-node\dist\config.js:174:59)
```

### DIFF
--- a/node-client/src/config.ts
+++ b/node-client/src/config.ts
@@ -228,7 +228,7 @@ export class KubeConfig {
                         // Format in file is {<query>}, so slice it out and add '$'
                         pathKey = '$' + pathKey.slice(1, -1);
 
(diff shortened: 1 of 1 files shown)


brendandburns:
Ugh, clearly we need better testing *sigh*

Thanks for the fix.

brendandburns:
@cmosgh this was pushed as `0.5.2`

thanks (and apologies for the break)

PR #83, opened 2018-08-15
Fixes #82 Add a client for the pod read-log endpoint

### LINKED ISSUE #82: Add a client for the pod read-log endpoint
**Command-line:** `kubectl log <pod-name>`
**Documentation:** https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#read-log-pod-v1-core

From what I read so far on https://github.com/kubernetes/kubernetes/pull/13885, the `/log` endpoints only expects `binary.k8s.io` as protocol. Without it (but still sending the `*.channel.k8s.io` ones), a `403 Forbidden` status code is returned.

I'll try to submit a PR for this in the upcoming days, suggestions are welcome.

Right now I'm trying to fit this into `WebSocketHandler`, but there is already a binary handler that expects the first byte of the message to be the channel number, which is not the case for the `/log` endpoint. See https://github.com/kubernetes-client/javascript/blob/master/node-client/src/web-socket-handler.ts#L89-L94

dbolkensteyn:
This PR adds the ability to read the logs of a pod. Fixes #82

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -1,0 +1,96 @@
+import { default as request } from 'request';
+import { Writable } from 'stream';
+
+import { KubeConfig } from './config';
+
+export interface LogOptions {
+    /**
+     * Follow the log stream of the pod. Defaults to false.
+     */
+    follow?: boolean;
+
+    /**
+     * If set, the number of bytes to read from the server before terminating the log output. This may not display a
+     * complete final line of logging, and may return slightly more or slightly less than the specified limit.
+     */
+    limitBytes?: number;
+
+    /**
+     * If true, then the output is pretty printed.
+     */
+    pretty?: boolean;
+
+    /**
+     * Return previous terminated container logs. Defaults to false.
+     */
+    previous?: boolean;
+
+    /**
+     * A relative time in seconds before the current time from which to show logs. If this value precedes the time a
+     * pod was started, only logs since the pod start will be returned. If this value is in the future, no logs will
+     * be returned. Only one of sinceSeconds or sinceTime may be specified.
+     */
+    sinceSeconds?: number;
+
+    /**
+     * If set, the number of lines from the end of the logs to show. If not specified, logs are shown from the creation
+     * of the container or sinceSeconds or sinceTime
+     */
+    tailLines?: number;
+
+    /**
+     * If true, add an RFC3339 or RFC3339Nano timestamp at the beginning of every line of log output. Defaults to false.
+     */
+    timestamps?: boolean;
+}
+
+export class Log {
+    public config: KubeConfig;
+
+    public constructor(config: KubeConfig) {
+        this.config = config;
+    }
+
+    public log(
+        namespace: string,
+        podName: string,
+        containerName: string,
+        stream: Writable,
+        done: (err: any) => void,
+        options: LogOptions = {},
+    ): request.Request {
+        const path = `/api/v1/namespaces/${namespace}/pods/${podName}/log`;
+
+        const cluster = this.config.getCurrentCluster();
+        if (!cluster) {
+            throw new Error('No currently active cluster');
+        }
+        const url = cluster.server + path;
+
+        const requestOptions: request.Options = {
+            method: 'GET',
+            qs: {
+                ...options,
+                container: containerName,
+            },
+            uri: url,
+        };
+        this.config.applyToRequest(requestOptions);
+
+        const req = request(requestOptions, (error, response, body) => {
+            if (error) {
+                done(error);
+            } else if (response && response.statusCode !== 200) {
+                done(body);
+            } else {
+                done(null);
+            }
+        }).on('response', (response) => {
+            if (response.statusCode === 200) {
+                req.pipe(stream);
+            }
+        });
+
+        return req;
+    }
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,3 +7,4 @@ export * from './exec';
 export * from './portforward';
 export * from './types';
 export * from './yaml';
+export * from './log';


node-client/src/web-socket-handler.ts
```diff
@@ -87,7 +88,7 @@ export class WebSocketHandler {
                             textHandler(message.utf8Data);
                         }
                     } else if (message.type === 'binary') {
-                        if (binaryHandler) {
+                        if (binaryHandler && message.binaryData.length > 0) {
```
dbolkensteyn: On my Kubernetes cluster, the first message received is an empty one.

node-client/src/log.ts
```diff
@@ -0,0 +1,36 @@
...
+            container: containerName,
+            follow: follow,
+            timestamps: timestamps,
+        };
+        const queryStr = querystring.stringify(query);
+        const path = `/api/v1/namespaces/${namespace}/pods/${podName}/log?${queryStr}`;
+        const promise = this.handler.connect(path, null, (streamNum: number, buff: Buffer) => {
+            const charBuff = Buffer.allocUnsafe(1)
+            charBuff.writeInt8(streamNum, 0)
+            stream.write(Buffer.concat([charBuff, buff]));
```
dbolkensteyn: Little bit ugly trick, but this allows to re-use the code from `WebSocketHandler` without having to it in a way that could break the other consumers of it (`Attach`, `Exec`).

A single `write()` call is issued on the stream because that tends to correspond to a line in the logs (although this is probably not guaranteed).

dbolkensteyn:
The flaky error seems unrelated to any code within this project to me: https://github.com/theturtle32/WebSocket-Node/blob/master/lib/WebSocketFrame.js#L256-L257

I suspect that, somehow, `((Math.random() * 0xFFFFFFFF) >>> 0)` is generating the value `-1732666326` which is out of the expected bounds of `>= 0 and <= 4294967295`.

I'll try to confirm and open a PR on the WebSocket-Node project.

node-client/package.json
```diff
@@ -38,7 +38,7 @@
     "request": "^2.72.0",
     "shelljs": "^0.7.8 ",
     "underscore": "^1.8.3",
-    "websocket": "^1.0.25"
+    "websocket": "^1.0.26"
```
dbolkensteyn: The diff on the auto-generated `package-lock.json` is pretty huge, unsure what to do with that, so I'm excluding it for the time being from this PR

node-client/src/log.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class Log {
+    public 'handler': WebSocketHandler;
+
+    public constructor(config: KubeConfig) {
+        this.handler = new WebSocketHandler(config);
+    }
+
+    // TODO add support for limitBytes, previous, sinceSeconds and tailLines
+    public log(namespace: string, podName: string, containerName: string,
```
brendandburns: indentation is weird here, can you run this through tslint?

node-client/src/log.ts
brendandburns: Sorry can you explain why this is needed? This is probably a bug in `WebSocketHandler` that should be fixed instead of doing this here...

(and it's probably my bug :)

node-client/src/log.ts
```diff
@@ -0,0 +1,41 @@
...
+import stream = require('stream');
+
+import { KubeConfig } from './config';
+import { WebSocketHandler } from './web-socket-handler';
+
+export class Log {
+    public 'handler': WebSocketHandler;
+
+    public constructor(config: KubeConfig) {
+        this.handler = new WebSocketHandler(config);
```
brendandburns: Does the log endpoint even use WebSockets? I thought it was a standard hanging GET, that's what we do in Java:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/PodLogs.java#L65

brendandburns:
Thanks for the PR!

Do we really need WebSockets here? I think in other clients this is implemented as a straight GET...

Thanks!

brendandburns:
@dbolkensteyn friendly ping on this, I'd love to get it merged, but I have some thoughts.

Thanks!

dbolkensteyn:
And I thank you for the valuable feedback @brendandburns - I'll come back to this during the week.

brendandburns:
Friendly ping on this if you still have time I'd love to get this merged.

Thanks!

brendandburns:
@dbolkensteyn  Friendly ping on this.

Thanks!

brendandburns:
Final ping, otherwise I'm going to close and build a different implementation of this functionality.

Thanks!

dbolkensteyn:
Sorry for the late follow-up @brendandburns , I've just forced-pushed a new implementation that incorporates all of your feedback and the one from #214 which actually implements the same feature (but hard-codes`follow` to true)

src/log.ts
```diff
@@ -0,0 +1,96 @@
...
+            if (error) {
+                done(error);
+            } else if (response && response.statusCode !== 200) {
+                done(body);
+            } else {
+                done(null);
+            }
+        }).on('response', (response) => {
+            if (response.statusCode === 200) {
+                req.pipe(stream);
```
dbolkensteyn: unlike #214 this will only start to pipe the "logs" when a 200 response is returned (else Kubernetes error jsons might get treated as logs)

src/log.ts
```diff
@@ -0,0 +1,96 @@
...
+    public config: KubeConfig;
+
+    public constructor(config: KubeConfig) {
+        this.config = config;
+    }
+
+    public log(
+        namespace: string,
+        podName: string,
+        containerName: string,
```
dbolkensteyn: `containerName` is now mandatory

src/log.ts
```diff
@@ -0,0 +1,96 @@
+import { default as request } from 'request';
+import { Writable } from 'stream';
+
+import { KubeConfig } from './config';
+
+export interface LogOptions {
+    /**
+     * Follow the log stream of the pod. Defaults to false.
+     */
+    follow?: boolean;
```
dbolkensteyn: `follow` is taken as a parameter and not hard-coded

src/log.ts
```diff
@@ -0,0 +1,96 @@
+import { default as request } from 'request';
+import { Writable } from 'stream';
+
+import { KubeConfig } from './config';
+
+export interface LogOptions {
```
dbolkensteyn: This is taken over from #214 as it seems as a nice way to avoid overly-parameterized methods

src/log.ts
tettusud: @dbolkensteyn  How to stop piping? I am successfully able to open the logs (follow option) but what is the right way to terminate underlying request?

abierbaum:
Does anyone have an example of using this code?  I tried to use it to stream logs but have not gotten it to work.

izhilenkov:
@abierbaum You could use something like `LineStream` from `byline` (which exactly `Transform`) as a line stream for quick example

```
import { KubeConfig, Log } from '@kubernetes/client-node';
import { LineStream } from 'byline';

const kc = new KubeConfig();
kc.loadFromDefault();

const log = new Log(kc);
const logStream = new LineStream();

logStream.on('data', data => console.log(data.toString()));

log.log(namespace, pod, container, logStream, (err) => console.log(err), { follow: true });
```

kingdevnl:
How could i close the log?

Iv'e tried to call logStream.destroy() but the program still keeps running.

Celthi:
Hi guys, I use this class with `follow=true` but my file only has part of the log instead of all the log from the pod?
Is there anything I need to pay attention to use this class?

PR #91, opened 2018-08-23
Update dependency versions.

brendandburns:


brendandburns:
self-lgtm since these are security updates.

PR #94, opened 2018-08-24
Improve tests coverage.

brendandburns:


brendandburns:
Self LGTM since it is unit tests and reviewers are unavailable...

Issue #97, opened 2018-09-05
Helper function for app client creation


sheldonkwok:
Hello, 

I was trying to use the Apps API instead of the Core API and noticed there aren't helpers for creating clients from a config. I ended up writing my own that look almost identical to the config ones provided for Core. Should we add helpers for Apps as well? I can just submit a PR.

Thanks,
Sheldon

brendandburns:
I recently added this generic function:

https://github.com/kubernetes-client/javascript/blob/master/node-client/src/config.ts#L220

Which should enable you to create any API object type. Does that suit your needs?

If not, a PR would definitely be welcome.

sheldonkwok:
That's actually perfect! Thanks! Look forward to the release.

sheldonkwok:
We could also add another helper at the Config level to make it a little easier too. I'll submit a PR

brendandburns:
Thanks!

brendandburns:
Just pushed `0.6.0` which should contain this.

brendandburns:
Closed via #99

PR #99, opened 2018-09-08
Feature/api config helpers

sheldonkwok:
Re: #97 
I switched `ApiType` to a union type of a couple of Apis. However, this is probably not all of apis people use. Should we grok through and all of the Apis there or I can just revert it to what you had before?

### DIFF
--- a/node-client/src/config.ts
+++ b/node-client/src/config.ts
@@ -217,7 +217,7 @@ export class KubeConfig {
         );
     }
 
-    public makeApiClient<T extends ApiType>(apiClientType: { new(server: string): T}) {
+    public makeApiClient<T extends ApiType>(apiClientType: ApiConstructor<T>) {
         const apiClient = new apiClientType(this.getCurrentCluster().server);
         apiClient.setDefaultAuthentication(this);
 
@@ -306,7 +306,11 @@ export class KubeConfig {
 }
 
 export interface ApiType {
-    setDefaultAuthentication(config: KubeConfig);
+    setDefaultAuthentication(config: api.Authentication);
+}
+
+export interface ApiConstructor<T extends ApiType> {
+    new (server: string): T;
 }
 
 // This class is deprecated and will eventually be removed.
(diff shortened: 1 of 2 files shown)


node-client/src/config.ts
```diff
@@ -305,8 +305,10 @@ export class KubeConfig {
     }
 }
 
-export interface ApiType {
-    setDefaultAuthentication(config: KubeConfig);
+// Doesn't have all apis iterated
+export type ApiType = api.Core_v1Api | api.Apps_v1Api;
```
brendandburns: Let's preserve the original that I had, I want this to work for all APIs.

node-client/src/config.ts
```diff
@@ -318,22 +320,25 @@ export class Config {
     public static SERVICEACCOUNT_TOKEN_PATH =
     Config.SERVICEACCOUNT_ROOT + '/token';
 
-    public static fromFile(filename: string): api.Core_v1Api {
+    public static apiFromFile<T extends ApiType>(filename: string, apiClientType: ApiConstructor<T>): T {
```
brendandburns: Let's keep the original functions too so that we don't break users with existing code who upgrade the library.

node-client/src/config.ts
```diff
@@ -318,22 +320,25 @@ export class Config {
...
         const kc = new KubeConfig();
         kc.loadFromFile(filename);
+        return kc.makeApiClient(apiClientType);
+    }
 
-        return kc.makeApiClient(api.Core_v1Api);
+    public static fromFile(filename: string): api.Core_v1Api {
```
brendandburns: move this up to match the original `fromFile` so that the diff is cleaner.

node-client/src/config.ts
```diff
@@ -347,10 +352,17 @@ export class Config {
         return k8sApi;
     }
 
-    public static defaultClient(): api.Core_v1Api {
+    public static fromCluster(): api.Core_v1Api {
```
brendandburns: move this up to match the old location so that the diff is cleaner.

node-client/src/config.ts
```diff
@@ -347,10 +352,17 @@ export class Config {
...
+    }
+
+    public static apiFromDefaultClient<T extends ApiType>(apiClientType: ApiConstructor<T>): T {
         const kc = new KubeConfig();
         kc.loadFromDefault();
+        return kc.makeApiClient(apiClientType);
+    }
 
-        return kc.makeApiClient(api.Core_v1Api);
+    public static defaultClient(): api.Core_v1Api {
```
brendandburns: as above, relocate for a cleaner diff.

brendandburns:
Looks great, my only comment is to retain the original functions in their original location, that way the diff will be cleaner.  Otherwise this is ready to merge.

Thanks!

sheldonkwok:
Just reorganized functions and rebased. Thanks!

PR #102, opened 2018-09-18
LoadFromCluster, added option to prefix the configuration file path

NicolaiSchmid:
I'm trying to use this module with [telepresence](https://www.telepresence.io/), which will mount all volumes in a temporary subdirectory, so the Kubernetes Cluster configuration is not in its default path.
telepresenceio/telepresence#353

Since I want to be able to use this library with my telepresence setup, I've added the option to specify a prefix to the default path of the certs and tokens.

I'm aware, that this is an edge case, and the project would be better served if you could specify the whole path as the [`SERVICEACCOUNT_ROOT`](https://github.com/kubernetes-client/javascript/blob/master/node-client/src/config.ts#L318) in the config-class, but since that is deprecated, I think this is a sufficient midterm solution and my configuration could be replaced by a more robust solution when the classes get refactored.

### DIFF
--- a/node-client/src/config.ts
+++ b/node-client/src/config.ts
@@ -149,7 +149,7 @@ export class KubeConfig {
         ];
     }
 
-    public loadFromCluster() {
+    public loadFromCluster(pathPrefix: string = '') {
         const host = process.env.KUBERNETES_SERVICE_HOST;
         const port = process.env.KUBERNETES_SERVICE_PORT;
         const clusterName = 'inCluster';
@@ -164,7 +164,7 @@ export class KubeConfig {
         this.clusters = [
             {
                 name: clusterName,
-                caFile: Config.SERVICEACCOUNT_CA_PATH,
+                caFile: pathPrefix + Config.SERVICEACCOUNT_CA_PATH,
                 caData: null,
                 server: `${scheme}://${host}:${port}`,
                 skipTLSVerify: false,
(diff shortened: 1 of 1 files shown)


node-client/src/config.ts
```diff
@@ -164,7 +164,7 @@ export class KubeConfig {
         this.clusters = [
             {
                 name: clusterName,
-                caFile: Config.SERVICEACCOUNT_CA_PATH,
+                caFile: pathPrefix + Config.SERVICEACCOUNT_CA_PATH,
```
brendandburns: Please use new style interpolation syntax:

```js
`${pathPrefix}${Config.SERVICEACCOUNT_CA_PATH}`
```

node-client/src/config.ts
```diff
@@ -173,7 +173,7 @@ export class KubeConfig {
         this.users = [
             {
                 name: userName,
-                token: fs.readFileSync(Config.SERVICEACCOUNT_TOKEN_PATH).toString(),
+                token: fs.readFileSync(pathPrefix + Config.SERVICEACCOUNT_TOKEN_PATH).toString(),
```
brendandburns: Same as above.

brendandburns:
Couple of small comments otherwise looks good.

Thanks!

NicolaiSchmid:
Hello,
I'm not completely familiar with the release cycles of this repository. The bot says, my commit resides in the merged pool, when will there be a release of my code to NPM?

NicolaiSchmid:
@mbohlool @brendanburns Any updates on publishing to NPM, I really need this code, that's why I created the PR.

brendandburns:
@NicolaiSchmid I just pushed `0.7.0` with these fixes.

Apologies for the delay.

NicolaiSchmid:
Perfect!
Have you considered setting up an auto deployment via Travis like this with credentials from environment variables:
```yaml
deploy:
  provider: npm
  email: $NPM_EMAIL
  skip_cleanup: true
  api_key: $NPM_TOKEN
  on:
    branch: master
    tags: true
```

PR #113, opened 2018-09-29
Fix buffer trim while sending websocket messages

mojojoji:
`Buffer.from` creates a new buffer of the same length of initial buffer or data. The data send over websocket needs to append one byte to identify message type (stdout/stdin/sterr).
This merge request creates the buffer with length = 1+ data length and copies over the data/


mojojoji:
CLA signed

itowlson:
@mojojoji Looks like the CLA is not coming through which is blocking the merge... could you check the status and that the email addresses tally please?  (Or post another comment and see if that makes it take notice!)  Thanks!

PR #119, opened 2018-10-10
Fix a bug where attach didn't handle stdin. Add testing.

### LINKED ISSUE #118: unable to send data to stdin with attach
@brendandburns I tried the follow script but I am unable send any data to the stdin:
```
const k8s = require('@kubernetes/client-node');
var kc = new k8s.KubeConfig();
try { kc.loadFromCluster(); } catch( err ) { kc.loadFromDefault(); }
var attach = new k8s.Attach(kc);

attach.attach(
    'default',
    'busycat1-asdffdsa-w123s',
    'busycat1',
    process.stdout, process.stderr, process.stdin,
    true
)
```
while the container is fine by using kubectl attach ..

brendandburns:
Fixes #118 

@scher200


brendandburns:
@itowlson if these are ready to merge, can you reply with '/lgtm'

that will kick off the automation.

Issue #123, opened 2018-10-12
Exec.exec() fails in Electron app (ws.client not a constructor)


cabrinoob:
Hi,
I'am using your lib in an Angular 6 application. I'd like to use the Exec.exec() method to display a terminal in my app to open a shell command in a pod.

My problem is that it seems to have an error with the websocket connect() method

When I call this from my Angular app : 

```
let Exec = new k8s.Exec(this.kubeConfig)
Exec.exec('default','myservice-6c5c9995b7-h79pq','myservice','bash',this.outStream,this.errStream,this.inStream,true)
```

I have this error :

```
Error: Uncaught (in promise): TypeError: ws.client is not a constructor
TypeError: ws.client is not a constructor
    at WebSocketHandler.connect (web-socket-handler.js:65)
    at Exec.<anonymous> (exec.js:30)
    at Generator.next (<anonymous>)
    at exec.js:7
    at new ZoneAwarePromise (zone-mix.js:891)
    at push../node_modules/@kubernetes/client-node/dist/exec.js.__awaiter (exec.js:3)
    at Exec.exec (exec.js:19)
```

It's working fine in a nodeJS app.

flyboarder:
The library should work from an electron node.js instance.

cabrinoob:
I was'nt precise enough in my previous post : My Angular app is wrapped in an Electron application and I have this error message.

itowlson:
Can you provide a minimal project that reproduces the problem?  (E.g. a patch to an Electron baseline app.)

cabrinoob:
@itowlson : ok i'll try to give you that asap
Thank you

brendandburns:
@cabrinoob there is this PR here:

itowlson:
@brendandburns That PR addressed an Electron issue shelling out to an auth provider.  The error @cabrinoob is seeing looks more like a type level error.  I'm wondering if there is a versioning issue in play, or something like that.

edvald:
The issue is likely that the `websocket` library used has a native dependency. I came across this as well in a different context. I suggest switching to https://github.com/websockets/ws which has the binary dependencies as optional addons for performance improvement (which I do not think is very important for this library).

gaogaoSpark:
I came across  the same problem .how to Solve this problem？

web-socket-handler.js?f1bc:80 Uncaught (in promise) TypeError: ws.client is not a constructor
    at WebSocketHandler.connect (web-socket-handler.js?f1bc:80)
[...]

PR #132, opened 2018-10-21
Fix electron issues

edvald:
This aims to address the issues in #59 and #123. I'll note that personally I am not using Electron, but came across these issues when making a bundle via zeit/pkg. So it would be great if someone can test these changes in Electron directly. I also don't directly use any code that touches on the WebSocketHandler, so someone might want to do additional testing there.

Further, I looked into getting rid of synchronous execs and file reads but decided not to include those changes in here, because they would introduce API changes in various places (basically making a bunch of surface-level methods async that are currently synchronous). @brendandburns I'd highly recommend making those changes since I kinda think they'll be necessary at some point anyway, and I'm happy to make a PR for that, but I wanted to get your input first.


edvald:
Seems the Github downtime is causing issues here - I've updated my branch to fix the CI errors, but it's not showing up here yet. Figure it'll snap into place when Github is back in shape.

itowlson:
+1 to providing async options for things involving execs and reads, but also +1 to not breaking the existing API surface!  Should we open an issue so that we can discuss the way forward on this?

node-client/src/config.ts
```diff
@@ -273,12 +273,10 @@ export class KubeConfig {
...
                     if (config['cmd-path']) {
-                        let cmd = '"' + config['cmd-path'] + '"';
-                        if (config['cmd-args']) {
-                            cmd = cmd + ' ' + config['cmd-args'];
-                        }
+                        const args = config['cmd-args'] ? [config['cmd-args']] : [];
                         // TODO: Cache to file?
-                        const result = shelljs.exec(cmd, { silent: true });
+                        // TODO: do this asynchronously
+                        const result = execa.sync(config['cmd-path'], args);
```
itowlson: I don't know the inner workings of `shelljs.exec` but the name implies it used to go via the shell.  Should we therefore use `execa.shellSync`?  (This is a question, not a suggestion... I lack knowledge and would value your advice.)

node-client/src/web-socket-handler.ts
```diff
@@ -96,37 +104,38 @@ export class WebSocketHandler implements WebSocketInterface {
...
+                }
+            };
+
+            client.onmessage = ({ data }: { data: WebSocket.Data }) => {
+                // TODO: support ArrayBuffer and Buffer[] data types?
+                if (typeof data === 'string') {
+                    if (textHandler && !textHandler(data)) {
+                        client.close();
+                    }
+                } else if (binaryHandler && data instanceof Buffer) {
```
itowlson: I'd suggest we keep the existing structure where the top level `if` checks only on the type of data, and the second level makes it a no-op if there is no handler for the type of data received:

```
if (data instanceof Buffer) {
  if (binaryHandler) {
    // ...
  }
}
```

This keeps the structure consistent with the string case (although of course at the moment the behaviour is the same either way).

edvald:
@mkramb could you test this to see if it addresses your issue?

edvald:
Oh and also @cabrinoob, would be great to know if this fixes #123 for you

cabrinoob:
Hi, 
I've just pull the PR changes and integrated them in my electron App to check if it fixes the #123 issue. I still have an error but with a different message.

``` 
Error: Uncaught (in promise): Error: ws does not work in the browser. Browser clients must use the native WebSocket object
Error: ws does not work in the browser. Browser clients must use the native WebSocket object
    at new push../node_modules/@kubernetes/client-node/node_modules/ws/browser.js.module.exports (browser.js:4)
    at Promise (web-socket-handler.js:79)
```

edvald:
Ah, so we'll need to go one step further and set up https://github.com/heineiuo/isomorphic-ws. I'm flying a little blind here, but I've added a commit for that. Let me know if it helps.

node-client/src/config_test.ts
```diff
@@ -535,8 +535,7 @@ describe('KubeConfig', () => {
                     },
                 } as User);
             const opts = {} as requestlib.Options;
-            expect(() => config.applyToRequest(opts)).to.throw(
-                'Failed to refresh token: /bin/sh: 1: non-existent-command: not found');
+            expect(() => config.applyToRequest(opts)).to.throw('non-existent-command');
```
brendandburns: Let's keep the 'failed to refresh token' part...

brendandburns:
@edvald I merged your other PR so this one now has conflicts...

drubin:
@brendanburns Is it possible to release a patch for this? Or will it have to wait for 0.8?

PR #138, opened 2018-10-25
chore: update packages and package-lock

edvald:
I noticed that some packages were quite old, and that some dev dependencies were in as production dependencies, so I figured I'd run a quick update (I used [ncu](https://www.npmjs.com/package/npm-check-updates) for it).

Also, the package-lock.json file had many dependencies marked as "bundled", which doesn't match the package.json file and I'm guessing was unintentional?


brendandburns:
/lgtm
/approve

Thanks!

PR #143, opened 2018-10-31
Add prepare script

drubin:
In trying to test out #141 I discovered we were missing the built files. 

Running `build` during the [prepare](https://docs.npmjs.com/misc/scripts) part of the install process these will be created during the install.  

cc/ @edvald


edvald:
Yep, sorry about that, forgot to add this step.

flyboarder:
@brendandburns looks like I can't approve this - bump

PR #147, opened 2018-11-04
Fixes to be more lenient on configs without users.

brendandburns:
This makes the library work with kubernetes-dind clusters.

### DIFF
--- a/src/config_types.ts
+++ b/src/config_types.ts
@@ -101,13 +101,10 @@ function contextIterator(): u.ListIterator<any, Context> {
         if (!elt.context.cluster) {
             throw new Error(`contexts[${i}].context.cluster is missing`);
         }
-        if (!elt.context.user) {
-            throw new Error(`contexts[${i}].context.user is missing`);
-        }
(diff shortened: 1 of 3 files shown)


src/config_types.ts
```diff
@@ -101,7 +101,7 @@ function contextIterator(): u.ListIterator<any, Context> {
         if (!elt.context.cluster) {
             throw new Error(`contexts[${i}].context.cluster is missing`);
         }
-        if (!elt.context.user) {
+        if (elt.context.user === undefined) {
```
itowlson: Not quite sure of the import of this change.  Is the intention to permit the empty string (or `null`), but still forbid `undefined`?

brendandburns: That's what I was going for. TBH, I don't know what the "right" expectation is here.... I can test with a config with no user at all and see what `kubectl` does...

brendandburns: Looks like kubectl is ok with it 100% missing, so I just removed this.

Issue #146, opened 2018-11-02
Unhandled JSON parse errors in 'watch' cause the process to crash.


smcgunigal:
The `watch` function does not ensure the data received is valid JSON before parsing it and does not handle any errors that may come from parsing. This can cause the process to restart when the api-server responds with non-valid JSON, which happens intermittently if the watch is used continuously. In particular, if the etcd cluster is unresponsive due to a timeout, the response is not able to be parsed and will cause a crash.

scaccogatto:
If I watch `/apis/apps/v1/namespaces/{namespace}/deployments/{name}` the watcher just crashes because the reply contains no `type`.

I think it is related to: https://github.com/kubernetes-client/javascript/blob/01b721af83658b755ffa7bc3931383dc39e3a9fe/src/watch.ts#L63

Also I think that if k8s reply has no `type` we should just callback with type `undefined`.

brendandburns:
Fixed by #148

PR #148, opened 2018-11-04
watch method crash on unexpected objects

scaccogatto:
The watch method should not crash if the k8s's reply object contains no `type` or `object`. Instead we check if the reply is an actual object and callback even if the `type` is `undefined`

Should fix (at least partially)
#146

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -60,10 +60,10 @@ export class Watch {
             } else {
                 obj = JSON.parse(data) as WatchUpdate;
             }
-            if (obj.type && obj.object) {
+            if (typeof obj === 'object' && obj.object) {
                 callback(obj.type, obj.object);
             } else {
-                throw new Error(`unexpected object: ${JSON.stringify(obj)}`);
+                throw new Error(`unexpected ${typeof obj}: ${JSON.stringify(obj)}`);
             }
         });
 


scaccogatto:
I signed it

scaccogatto:
("I signed it")

scaccogatto:
I actually signed it but, well, I cannot trigger the bot.

src/watch.ts
```diff
@@ -60,10 +60,10 @@ export class Watch {
             } else {
                 obj = JSON.parse(data) as WatchUpdate;
             }
-            if (obj.type && obj.object) {
+            if (typeof obj === 'object' && obj.object) {
                 callback(obj.type, obj.object);
             } else {
-                throw new Error(`unexpected object: ${JSON.stringify(obj)}`);
+                throw new Error(`unexpected ${typeof obj}: ${JSON.stringify(obj)}`);
```
itowlson: This will result in very compacted JSON; if you want to make the error message human-readable then consider `JSON.stringify(obj, undefined, 2)` (but this will potentially result in a multiline message, which you may feel is undesirable for other reasons!).

src/watch.ts
```diff
@@ -60,10 +60,10 @@ export class Watch {
             } else {
                 obj = JSON.parse(data) as WatchUpdate;
             }
-            if (obj.type && obj.object) {
+            if (typeof obj === 'object' && obj.object) {
```
itowlson: Not sure the `typeof` check achieves much - the `obj.object` check will surely catch any non-object cases?  Seems like the goal of this PR is simply to allow the check to pass even if `obj.type` is `undefined` so perhaps just replace this with `if (obj.object)` or `if (obj && obj.object)`?

itowlson:
@brendandburns Do we have any way of checking if the CLA has gone through?  Looks like the bot is having trouble picking it up again.

src/watch.ts
scaccogatto: Well, I feel like this should be handled in another issue: I just wanted to have a little clearer message.
The point is the exception was already unclear and multiline most of the times, we should review this in another way for sure but this is not the point of my pull request: making the code not crash during a documented event.

scaccogatto: We can absolutely have a better error management for this but I don't feel ready for a little refactor (first contribution here and low experience on Typescript)

src/watch.ts
scaccogatto: I just wanted to make sure the reply was an object (as it should be as far as I know) so I think this line is just a simple improvement!

scaccogatto:
> @brendandburns Do we have any way of checking if the CLA has gone through? Looks like the bot is having trouble picking it up again.

image

scaccogatto:
I signed it

brendandburns:
CLA now signed, fwiw, thanks for the patience!

PR #154, opened 2018-11-16
Add a test for empty users.

### LINKED ISSUE #125: Empty-string username should be supported
Here's a valid configuration for Kubernetes clusters <= 1.5
```
[05:56 core@master folder] > kubectl config view
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: http://10.0.0.1:8080
  name: local-server
contexts:
- context:
    cluster: local-server
    user: ""
  name: default-context
current-context: default-context
kind: Config
preferences: {}
users: []
```

And here's [the error](https://github.com/kubernetes-client/javascript/blob/c1c0039be792b829037553d946c12b28a6142afb/node-client/src/config_types.ts#L55) I get when I try to connect to it:
```
Error: contexts[0].context.user is missing
    at /project/node_modules/@kubernetes/client-node/dist/config_types.js:76:19
```

KubeConfig should support empty-string user for insecure clusters.

brendandburns:
Closes #125

### DIFF
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -11,6 +11,7 @@ import { KubeConfig } from './config';
 import { Cluster, newClusters, newContexts, newUsers, User } from './config_types';
 
 const kcFileName = 'testdata/kubeconfig.yaml';
+const kcNoUserFileName = 'testdata/empty-user-kubeconfig.yaml';
 
 /* tslint:disable: no-empty */
 describe('Config', () => { });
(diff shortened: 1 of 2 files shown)


src/config_test.ts
```diff
@@ -11,6 +11,7 @@ import { KubeConfig } from './config';
 import { Cluster, newClusters, newContexts, newUsers, User } from './config_types';
 
 const kcFileName = 'testdata/kubeconfig.yaml';
+const kcNoUserFileName = 'testdata/empty-user-kubeconfig.yaml';
```
itowlson: Does this file need to be added to the PR?

brendandburns: added.

brendandburns:
@itowlson comment addressed, I think this is ready to go.

PR #155, opened 2018-11-19
Fixed ConfigMap patch header issue in Api lib

janez89:
I found a issue with ConfigMap update in Api library.

If you do patch update on ConfigMap then the request will crash.

Reason: missing patch header

Reproduce:
```ts
this.apiService
    .patchNamespacedConfigMap(
        'configmap-name', 
        'default',
        { data: {  'key1': 'value 1',  'key2': 'value 2' } }
    );
```


janez89:
I finished the CLA process.

janez89:
Ohh I see.
yes I found that. Thx. I will go on different way.

PR #164, opened 2018-12-21
Fix Node10 Unit tests.

brendandburns:
@itowlson


src/config.ts
```diff
@@ -17,18 +17,22 @@ export class KubeConfig {
...
             const dir = path.join(process.env.HOMEDRIVE, process.env.HOMEPATH);
-            if (fs.existsSync(dir)) {
+            try {
+                fs.accessSync(dir);
                 return dir;
-            }
+            // tslint:disable-next-line:no-empty
+            } catch (ignore) {}
         }
         if (process.env.USERPROFILE && fs.existsSync(process.env.USERPROFILE)) {
```
drubin: @brendanburns should we not replace this too? `fs.existsSync(process.env.USERPROFILE`

brendandburns: done. (and a few other places I missed)

brendandburns:
@drubin please take another look.

Thanks!

PR #168, opened 2018-12-22
Fix Exec implementation, improve tests.

brendandburns:
@SawyerHawkins


brendandburns:
Switching reviewer to @itowlson

src/config.ts
```diff
@@ -17,21 +17,29 @@ export class KubeConfig {
     // Only public for testing.
     public static findHomeDir(): string | null {
         if (process.env.HOME) {
-            if (fs.existsSync(process.env.HOME)) {
+            try {
+                fs.accessSync(process.env.HOME);
```
itowlson: I'm not sure why `existsSync` is being changed to `try { accessSync } catch { /* ignore */ }` as part of this PR.  Assuming there's a good reason, let's not clutter this flow with exception swallowing - let's create a helpfully named wrapper function e.g. `canAccess` or `existsPlusPlus` or whatever your intent is.

src/exec_auth.ts
```diff
@@ -4,12 +4,25 @@ import { Authenticator } from './auth';
 import { User } from './config_types';
 
 export class ExecAuth implements Authenticator {
+    private tokenCache: any = {};
```
itowlson: readonly?

src/exec_auth.ts
```diff
@@ -4,12 +4,25 @@ import { Authenticator } from './auth';
...
+        // TODO: Handle client cert auth here, requires auth refactor.
+        // See https://kubernetes.io/docs/reference/access-authn-authz/authentication/#input-and-output-formats
+        // for details on this protocol.
+        const cachedToken = this.tokenCache[user.name];
+        if (cachedToken) {
+            const date = Date.parse(cachedToken.status.expirationTimestamp);
+            if (date < Date.now()) {
+                return cachedToken.status.token;
+            }
+            this.tokenCache[user.name] = null;
```
itowlson: Maybe I am missing some context here but I can't see where `tokenCache` is getting populated?

src/exec_auth.ts
itowlson: Could type be `{ [key: string]: SomeTokenType }`?

src/exec_auth.ts
brendandburns: Fixed.

src/config.ts
brendandburns: This was originally done here:

https://github.com/kubernetes-client/javascript/pull/164

To fix the node 10 tests which were breaking.

I've added the cleanup you requested.

src/exec_auth.ts
brendandburns: done.

src/exec_auth.ts
brendandburns: Fixed. (good catch)

src/exec_auth.ts
brendandburns: done.

itowlson:
CI fail looks like unit test failures?

PR #169, opened 2018-12-24
Add an initial watch cache implementation.

brendandburns:
@itowlson 

Beginning of some more advanced capabilities for javascript/typescript.


src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+    get(name: string, namespace?: string): T | null;
+    list(namespace?: string): T[];
+}
+
+export type ListCallback<T extends KubernetesObject> = (list: T[]) => void;
+
+export class ListWatch<T extends KubernetesObject> implements ObjectCache<T> {
+    private objects: T[] = [];
+    private indexCache: any = {};
+    private path: string;
```
itowlson: I think these can be readonly, and that you can remove the explicit declarations and use TypeScript's declare-and-capture-in-the-constructor syntax `constructor(private readonly path: string, ...) { ... }`.

src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+        this.path = path;
+        this.doneHandler(null);
+    }
+
+    public get(name: string, namespace?: string): T | null {
+        let result: T | null = null;
+        for (const element of this.objects) {
+            if (element.metadata.name === name &&
+                (!namespace || element.metadata.namespace === namespace)) {
+                result = element;
```
itowlson: Can you `return element` here, or is it necessary to keep searching the list for further matching elements and return the last?

If you want to return the first that matches the criterion, consider `this.objects.find(...)`.  If you want to return the last, consider searching backward through the list, in order to avoid continued searching after finding a match.

src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+export interface ObjectCache<T> {
+    get(name: string, namespace?: string): T | null;
+    list(namespace?: string): T[];
+}
+
+export type ListCallback<T extends KubernetesObject> = (list: T[]) => void;
+
+export class ListWatch<T extends KubernetesObject> implements ObjectCache<T> {
+    private objects: T[] = [];
+    private indexCache: any = {};
```
itowlson: I think you might be able to pin the type down to `{ [key: string]: T[] }` which would give you additional checking.

src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+                (!namespace || element.metadata.namespace === namespace)) {
+                result = element;
+            }
+        }
+        return result;
+    }
+
+    public list(namespace?: string | undefined): T[] {
+        if (!namespace) {
+            return this.objects;
```
itowlson: Should we make a copy here?  Are we worried if some does a `list(...).push(...)`?

src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (ix === -1) {
+        objects.push(obj);
+    } else {
+        objects[ix] = obj;
+    }
+}
+
+// Public for testing.
+export function findObject<T extends KubernetesObject>(objects: T[], obj: T): number {
+    for (let ix = 0; ix < objects.length; ix++) {
```
itowlson: Can this be replaced by a call to `objects.findIndex(...)`?

src/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+        const elt = objects[ix];
+        if (obj.metadata.name !== elt.metadata.name) {
```
itowlson: I feel this would be more readable as `if (obj.metadata.name === elt.metadata.name && obj.metadata.namespace === elt.metadata.namespace) { return ix; }`.  The use of `continue` to implement short-circuiting causes the reader more work understanding the different control flows.  Better still would be to pull out the compound condition into a helpfully named function e.g. `isSameResource`.

src/cache.ts
brendandburns: done.

src/cache.ts
brendandburns: switched to `find` (learning more Javascript by the day :)

src/cache.ts
brendandburns: Done (learning more Typescript by the day!)

src/cache.ts
brendandburns: Switched to `ReadonlyArray<T>`

src/cache.ts
brendandburns: Learning even moar Javascript (done)

src/cache.ts
brendandburns: done.

brendandburns:
Comments addressed @itowlson please take another look!

src/cache.ts
itowlson: HOW DID I NOT KNOW ABOUT `ReadonlyArray<T>` HEADS WILL ROLL FOR THIS.  Thank you, that's some TypeScript learning for me today too!

src/cache.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+function isSameObject<T extends KubernetesObject>(o1: T, o2: T): boolean {
+    return o1.metadata.name === o2.metadata.name &&
+           o1.metadata.namespace === o2.metadata.namespace;
+}
+
+function findKubernetesObject<T extends KubernetesObject>(objects: T[], obj: T): number {
+    return objects.findIndex((elt: T): boolean => {
```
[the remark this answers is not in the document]
brendandburns: done.

brendandburns:
Comment addressed, please take another look.

Issue #178, opened 2019-01-11
access-token does not get generated if no access-token is present


macrozone:
follow up of https://github.com/kubernetes-client/javascript/issues/177

If you never used an auth-provider, your .kube/config looks like this (e.g. with google cloud):

```
# ...
users:
- name: xxxclustername
  user:
    auth-provider:
      config:
        cmd-args: config config-helper --format=json
        cmd-path: /Applications/google-cloud-sdk/bin/gcloud
        expiry-key: '{.credential.token_expiry}'
        token-key: '{.credential.access_token}'
      name: gcp
```

in this case the js-client sees that there is an auth-provider and tries to get the propery `access-token`, which is undefined here.

There is some logic to update the access token if it is expired, but this also only works if the token has been generated before, because both `access-token` and `expiry` are only set on token renewal.

Additionaly, the current token renewal implementation does not store back access-token and expiry into the config.

As a reference, the `kubectl` cli behaves correctly. If there is no access-token or if it is expired, it will fetch a new one and write it back to .kube/config

PR #179, opened 2019-01-11
fix: access-token does not get refreshed if not present

macrozone:
fixes https://github.com/kubernetes-client/javascript/issues/178

it does not, however, store the access-token back into .kube/config, i think that is a separate issue.

### DIFF
--- a/src/cloud_auth.ts
+++ b/src/cloud_auth.ts
@@ -5,6 +5,18 @@ import * as shelljs from 'shelljs';
 import { Authenticator } from './auth';
 import { User } from './config_types';
 
+/* FIXME: maybe we can extend the User and User.authProvider type to have a proper type.
+Currently user.authProvider has `any` type and so we don't have a type for user.authProvider.config.
+We therefore define its type here
+*/
+interface Config {
+    expiry: string;
+    ['cmd-args']?: string;
+    ['cmd-path']?: string;
+    ['token-key']: string;
+    ['expiry-key']: string;
+    ['access-token']?: string;
+}
 export class CloudAuth implements Authenticator {
     public isAuthProvider(user: User): boolean {
         return (
@@ -21,7 +33,7 @@ export class CloudAuth implements Authenticator {
         return 'Bearer ' + config['access-token'];
     }
 
-    private isExpired(config) {
+    private isExpired(config: Config) {
         const token = config['access-token'];
         const expiry = config.expiry;
         if (!token) {
@@ -38,7 +50,7 @@ export class CloudAuth implements Authenticator {
         return false;
     }
 
-    private updateAccessToken(config) {
+    private updateAccessToken(config: Config) {
         if (!config['cmd-path']) {
             throw new Error('Token is expired!');
         }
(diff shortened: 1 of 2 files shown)


macrozone:
@brendandburns ok done. Please consider adding a prettier-config to the project.

src/cloud_auth.ts
```diff
@@ -7,51 +7,69 @@ import { User } from './config_types';
...
         const config = user.authProvider.config;
-        // This should probably be extracted as auth-provider specific plugins...
-        let token: string = 'Bearer ' + config['access-token'];
+        if (this.isExpired(config)) {
+            this.updateAccessToken(config);
+        }
+        return 'Bearer ' + config['access-token'];
+    }
+
+    private isExpired(config) {
```
brendandburns: Add type information here.

src/cloud_auth.ts
```diff
@@ -7,51 +7,69 @@ import { User } from './config_types';
...
+        const token = config['access-token'];
         const expiry = config.expiry;
+        if (!token) {
```
brendandburns: nit: inline `config['access-token']` or use `contains`

src/cloud_auth.ts
```diff
@@ -7,51 +7,69 @@ import { User } from './config_types';
...
-                    let pathKey = config['token-key'];
-                    // Format in file is {<query>}, so slice it out and add '$'
-                    pathKey = '$' + pathKey.slice(1, -1);
-
-                    config['access-token'] = jsonpath.query(resultObj, pathKey);
-                    token = 'Bearer ' + config['access-token'];
-                } else {
-                    throw new Error('Token is expired!');
-                }
+    private updateAccessToken(config) {
```
brendandburns: Include type information for the parameter.

brendandburns:
Some small comments, also the commit history seems messed up here. Needs a rebase?

macrozone:
@brendanburns i rebased, that's why the history is messed up. I usually don't rebase as it is very error prone and i am not used to it. I have no idea how i can fix that.

src/cloud_auth.ts
macrozone: would break the symetry here between `access-token` and `expiry`.

i can inline it if you really want, but it does not make the code any better imho

src/cloud_auth.ts
macrozone: agree, but there is no type definition for `authProvider`, it's `any`-type, so i would have to invent a new type for the config.

src/cloud_auth.ts
macrozone: see above

macrozone:
I added a type for config and fixed a bug. 
The tslint forces an interface for that type (i come from flow-type, where you usually just use `type` for that). Does it look correct?

brendandburns:
This looks fine, but CI/CD is failing. run `npm test`

brendandburns:
@macrozone friendly ping on this if you want to get it to merge...

macrozone:
@brendanburns on it

PR #180, opened 2019-01-12
Fix sending protocols for web sockets.

### LINKED ISSUE #170: Exec.exec fails on json.parse
Using version 0.7.1 when I run exec which fails (status code 1) I got a valid json on the handler - https://github.com/kubernetes-client/javascript/blob/master/src/web-socket-handler.ts#L47
So the JSON.parse is working as expected. 
Since version 0.7.2 I got a string which  make the JSON.parse to throw.
Here is real example of response from 0.7.1 and 0.7.2
0.7.1:
```js
{
	"metadata": {},
	"status": "Failure",
	"message": "command terminated with non-zero exit code: Error executing in Docker Container: 137",
	"reason": "NonZeroExitCode",
	"details": {
		"causes": [{
			"reason": "ExitCode",
			"message": "137"
		}]
	}
}
```
0.7.2:
```js
"command terminated with non-zero exit code: Error executing in Docker Container: 137"
```

brendandburns:
Fixes #170 

@GiladShoham

### DIFF
--- a/examples/typescript/exec/exec-example.ts
+++ b/examples/typescript/exec/exec-example.ts
@@ -6,10 +6,12 @@ const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
 const exec = new k8s.Exec(kc);
-exec.exec('default', 'demo', 'demo', command,
+exec.exec('default', 'nginx-4217019353-9gl4s', 'nginx', command,
     process.stdout, process.stderr, process.stdin,
     (status: k8s.V1Status) => {
+        // tslint:disable-next-line:no-console
         console.log('Exited with status:');
+        // tslint:disable-next-line:no-console
         console.log(JSON.stringify(status, null, 2));
     },
     true /* tty */);
--- a/examples/typescript/exec/exec-example.ts
+++ b/examples/typescript/exec/exec-example.ts
@@ -1,4 +1,5 @@
-import k8s = require('@kubernetes/client-node');
+import * as k8s from '@kubernetes/client-node';
+import * as stream from 'stream';
 
 const command = process.argv[2];
 
@@ -7,11 +8,11 @@ kc.loadFromDefault();
 
 const exec = new k8s.Exec(kc);
 exec.exec('default', 'nginx-4217019353-9gl4s', 'nginx', command,
-    process.stdout, process.stderr, process.stdin,
+    process.stdout as stream.Writable, process.stderr as stream.Writable, process.stdin as stream.Readable,
+    true /* tty */,
     (status: k8s.V1Status) => {
         // tslint:disable-next-line:no-console
         console.log('Exited with status:');
         // tslint:disable-next-line:no-console
         console.log(JSON.stringify(status, null, 2));
-    },
-    true /* tty */);
+    });
--- a/examples/typescript/exec/exec-example.ts
+++ b/examples/typescript/exec/exec-example.ts
@@ -6,5 +6,10 @@ const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
 const exec = new k8s.Exec(kc);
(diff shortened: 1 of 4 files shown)


src/exec.ts
```diff
@@ -18,7 +19,8 @@ export class Exec {
 
     // TODO: make command an array and support multiple args
     public async exec(namespace: string, podName: string, containerName: string, command: string,
-                      stdout: stream.Writable | any, stderr: stream.Writable | any, stdin: stream.Readable | any,
+                      stdout: stream.Writable | null, stderr: stream.Writable | null, stdin: stream.Readable | null,
+                      statusCallback: (status: V1Status) => void,
```
itowlson: Is this a breaking change for clients?  Is there anything we can do to make it non-breaking, e.g. make it the last parameter and optional?  (We seem to handle it being `null`, so could also handle it being `undefined`.)

src/exec_test.ts
```diff
@@ -69,7 +72,7 @@ describe('Exec', () => {
             when(fakeWebSocket.connect(`${path}?${args}`, null, anyFunction())).thenResolve(fakeConn);
 
             await exec.exec(
-                namespace, pod, container, cmd, osStream, errStream, isStream, false);
+                namespace, pod, container, cmd, osStream, errStream, isStream, nop, false);
```
itowlson: No new tests for the status handler being called?

src/exec.ts
itowlson: Usage of `statusCallback` is non-obvious, even with the exec sample.  Can we provide any documentation on this?

examples/typescript/exec/exec-example.ts
```diff
@@ -1,9 +1,15 @@
...
+const command = process.argv[2];
 
-let kc = new k8s.KubeConfig();
+const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
-let exec = new k8s.Exec(kc);
-exec.exec('default', 'nginx-4217019353-9gl4s', 'nginx', command, process.stdout, process.stderr, process.stdin, true /* tty */);
+const exec = new k8s.Exec(kc);
+exec.exec('default', 'demo', 'demo', command,
```
drubin: I think it would be a nice to be able to show the difference between pod/container here. Either via a comment or via the previous example `'nginx-4217019353-9gl4s` shows more clearly its pod

src/exec.ts
brendandburns: Updated so that it's not a breaking change.

src/exec_test.ts
brendandburns: Added a test.

src/exec.ts
brendandburns: Added JSDoc.

examples/typescript/exec/exec-example.ts
brendandburns: reverted.

brendandburns:
@itowlson @drubin I think this is read for another look.

Thanks!

drubin:
@brendandburns It needs a rebase but

/lgtm

examples/typescript/exec/exec-example.ts
```diff
@@ -1,9 +1,17 @@
...
+const exec = new k8s.Exec(kc);
+exec.exec('default', 'nginx-4217019353-9gl4s', 'nginx', command,
+    process.stdout, process.stderr, process.stdin,
+    (status: k8s.V1Status) => {
+        // tslint:disable-next-line:no-console
+        console.log('Exited with status:');
+        // tslint:disable-next-line:no-console
+        console.log(JSON.stringify(status, null, 2));
+    },
+    true /* tty */);
```
itowlson: Does this typecheck with the new (backward compatible) parameter order?

PR #182, opened 2019-01-15
Upgrade client badge to gold

drubin:
I think this library qualifies as gold support under [csi-new-client-library-procedure](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/csi-new-client-library-procedure.md#client-capabilities)

We now support the following 

* Support watch calls
* Support exec, attach, port-forward calls (these are not normally supported out of the box from swagger-codegen)
* Proto encoding
* Generated examples

Not 100% sure about fully generated examples but we do have pretty good example code.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Javascript Kubernetes Client information
 
 [![Build Status](https://travis-ci.org/kubernetes-client/javascript.svg?branch=master)](https://travis-ci.org/kubernetes-client/javascript)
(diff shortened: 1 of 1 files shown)


brendandburns:
Generated examples was removed as a qualifier (https://github.com/kubernetes/community/pull/3354)

So this is ready to go!

/lgtm
/approve

drubin:
@brendanburns thanks for getting back to this.

PR #173, opened 2019-01-06
regenerate for 1.13.2

flyboarder:
regenerate the api for k8s version 1.13.2

PR #184, opened 2019-01-16
Upgrade to 1.13 API with fixes

brendandburns:
@flyboarder 

This is #173 with an added commit to fix CI/CD.

@flyboarder probably the right thing to do is to cherry-pick my commit onto your PR and then merge your PR (and then we can close this PR)

Alternately, we could merge this PR instead.


flyboarder:
@brendandburns we can just merge this one, if it fixes the problems then great let's get it into master. My questions: Do we need to run any additional steps in the future to regenerate this package? How did you fix it? Can we include those steps within a build script?

brendandburns:
@flyboarder I hand edited the generated code. This isn't sustainable, we need a fix in the code generator, but we're likely to switch the code generator before we regenerate again...

brendandburns:
@drubin @itowlson can I get a quick review here?

Thanks!

flyboarder:
@brendandburns where is the work for switching the code generator? I'd like to get in on that!

PR #186, opened 2019-01-17
Support insecure-skip-tls-verify as config option for port-forward

drubin:
Propagate the `insecure-skip-tls-verify`  down to the `http` request library.


drubin:
This allows us to use `docker-for-mac` as an endpoint for portforwarding.

brendandburns:
/lgtm
/approve

Thanks!

PR #189, opened 2019-01-23
Add `namespace` To Context

swcho:
Thank you for your efforts on this project.

I'd like to add `namespace` in `Context` object and properly parsed from configuration file.

Thank you.

### DIFF
--- a/src/config_types.ts
+++ b/src/config_types.ts
@@ -105,7 +105,7 @@ function contextIterator(): u.ListIterator<any, Context> {
         return {
             cluster: elt.context.cluster,
             name: elt.name,
-            user: (elt.context.user ? elt.context.user : undefined),
+            user: elt.context.user || undefined,
             namespace: elt.context.namespace || undefined,
         };
     };


itowlson:
@swcho Thanks for the contribution!  Can you clarify your intended scope here?  It looks like this will surface the `namespace` in a kubeconfig to the JavaScript `Context` objects, but doesn't appear to _use_ the namespace in any operations involving the context.  Is that your intention?

brendandburns:
@itowlson I think that's a reasonable feature to support, since I don't think we can/should set the default namespace in the client via this parameter, but making it available to users of the client library makes sense.

@swcho you need to sign the CNCF CLA, also will you add a unit test to validate this behavior?

Thanks!

swcho:
@itowlson 
Thank you for reviewing.

It is somehow trivial change but as document described below `namespace` is one of configurable item and I think it is beneficial for developer to have access to it.
[Define clusters, users, and contexts](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#define-clusters-users-and-contexts)

Especially, on-premise kube cluster, not like kube cluster provisioned via major cloud vendor, `namespace` can be utilized to encapsulate service environment.

My use case is below. And I just want to have `namespace` access-able through official client library.
https://github.com/swcho/kube-eye/blob/486ebc5c5f562d95ff9b472517b99131e2140f30/src/server.tsx#L32-L39

@brendandburns
As soon as I have some time to make unit test, I will try it.

Thank you all for having some time to review.

src/config_types.ts
```diff
@@ -84,6 +84,7 @@ export interface Context {
     readonly cluster: string;
     readonly user: string;
     readonly name: string;
+    readonly namespace?: string;
```
drubin: Should this not be just `namespace` like the others and set to `undefined` if not set?

src/config_types.ts
```diff
@@ -105,6 +106,7 @@ function contextIterator(): u.ListIterator<any, Context> {
             cluster: elt.context.cluster,
             name: elt.name,
             user: (elt.context.user ? elt.context.user : undefined),
+            namespace: elt.context.namespace || undefined,
```
drubin: Could we standise on one of the above to formats? Not sure which one is better though.

brendandburns: `||` syntax is fine with me, and agree standardization is good.

swcho:
@brendandburns 

Sorry for late reply, I added simple test case for `namespace` access.

https://github.com/kubernetes-client/javascript/pull/189/commits/1695e271ce8489683deaa14129c1f33aa5058f0f

Thank you.

src/config_types.ts
```diff
@@ -105,6 +106,7 @@ function contextIterator(): u.ListIterator<any, Context> {
             cluster: elt.context.cluster,
             name: elt.name,
             user: (elt.context.user ? elt.context.user : undefined),
```
brendandburns:
```suggestion
            user: elt.context.user || undefined,
```

brendandburns:
@swcho looks good to me, I added a suggestion based on @drubin 's comment above, if you're ok with that, then I think this is ready to merge.

Thanks

PR #199, opened 2019-01-31
Remove explicit references to bluebird.

### LINKED ISSUE #198: 0.8.0 exposes bluebird promises in API, breaking things
Commit f4c4a647b94f71fed3e7ef7fd681789c7c89b9af exposes bluebird promises as part of the externally facing API. This breaks downstream TypeScript projects in two ways:

1.  The `@types/bluebird` were not moved from "devDependencies" to "dependencies" in the package.json, see #156 for context.
2.  Bluebird promises are incompatible with ES6 Promises, see DefinitelyTyped/DefinitelyTyped#11027, so downstream projects end up having to wrap, workaround, or use bluebird promises everywhere. The latter option is not viable if you use async/await since in those instances TypeScript requires that the return type be an ES6 Promise, see Microsoft/TypeScript#5911.

The first issue is easy enough to fix. The second issue is a bit more complicated. Is there a compelling reason to use bluebird promises over native ES6 promises? If so, it is going to cause compilation failures in downstream projects.

I'd be happy to submit a PR to revert the addition of bluebird.

brendandburns:
Fixes #198


drubin:
Do you think we should add a note to the readme about this change? Or is it time to add an upgrade guide too?

drubin:
I am aware. I was more suggesting we should try make it more visible so users can easily see.

ddgenome:
It would be great if this project had a CHANGELOG.md.

PR #200, opened 2019-01-31
move @types/ws to a runtime dependency

sroylance:
I believe this will fix a warning I'm getting:
node_modules/@kubernetes/client-node/dist/exec.d.ts:2:23 - error TS2688: Cannot find type definition file for 'ws'.

2 /// <reference types="ws" />

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   "dependencies": {
     "@types/node": "^10.12.0",
     "@types/request": "^2.47.1",
+    "@types/ws": "^6.0.1",
     "base-64": "^0.1.0",
     "bluebird": "^3.5.2",
     "byline": "^5.0.0",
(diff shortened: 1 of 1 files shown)


itowlson:
Looks good and would be great to fix this annoyance!  When you say you believe this will fix the warning, have you been able to test this, or does it need to go through a packaging step to test that it fixes the warning?

sroylance:
it needs to be tested, I haven't tried to build it.  Adding @types/ws to the package.json in the application I'm working on solved the problem for me.

drubin:
@sroylance Thanks for this! 

I think #202 covers your usecase as well as fixing a few others. Are you ok if we close this in favour of that?

drubin:
/close

#202 was merged

Let me know if there are still outstanding issues.

Issue #156, opened 2018-11-22
Typescript definitions are broken in 0.7.1


mkraeva-astea:
The 0.7.1 release has broken the typescript definitions - running `tsc` on the example for listing all pods, I get numerous errors like
> node_modules/@kubernetes/client-node/dist/api.d.ts:13:26 - error TS7016: Could not find a declaration file for module 'request'. '/home/marina/kube-dep-test/node_modules/request/index.js' implicitly has an 'any' type.

The typings work in v0.6.0; v0.7.0 compiles successfully as well, but I get the runtime error described in #135.

brendandburns:
hrm, I will try to figure out what is missing here.  Apologies!

brendandburns:
Looks like this was fixed by #138 I'll cut a new release...

brendandburns:
Apologies for the delay, I finally pushed `0.7.2` please try and see if it fixes things.

Thanks!

mkraeva-astea:
Unfortunately, v0.7.2 doesn't work either.

I made a minimal reproducible example here: https://github.com/mkraeva-astea/kube-client-typescript-example.

The example uses v0.6.0 and compiles correctly -  `npm install && npm run build` runs successfully and `npm test` lists my local pods.
If you then run `npm install @kubernetes/client-node@0.7.2 --save-exact` and try to rebuild, you can see multiple errors, e.g.
> Could not find a declaration file for module 'ws'. '/home/marina/kube-dep-test/node_modules/ws/index.js' implicitly has an 'any' type.

drakedevel:
@mkraeva-astea it looks like the issue is missing type dependencies (specifically `@types/ws @types/node @types/request`)

If I add those, your example builds for me:

```
[adrake@fbox /tmp/kube-client-typescript-example (master)]$ npm install && npm run build 
<...snip...>

added 121 packages from 185 contributors and audited 151 packages in 2.529s
found 0 vulnerabilities


> kube-dep-test@1.0.0 build /tmp/kube-client-typescript-example
> rm -rf dist && tsc

[adrake@fbox /tmp/kube-client-typescript-example (master)]$ npm install @kubernetes/client-node@0.7.2 --save-exact
npm WARN deprecated nomnom@1.5.2: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN kube-dep-test@1.0.0 No description
npm WARN kube-dep-test@1.0.0 No repository field.

+ @kubernetes/client-node@0.7.2
[...]

PR #39, opened 2018-04-20
Move @types to package.json dependencies for definitions and add missing @types

sheldonkwok:
We need the module's dependency's types in the `dependencies` part of the package.json for typescript to see them. More reasoning can be found here: https://github.com/Microsoft/types-publisher/issues/81

PR #202, opened 2019-02-01
Move appropriate @types back to dependencies

sheldonkwok:
The types seem to have drifted back into the devDependencies from when I moved them over originally in https://github.com/kubernetes-client/javascript/pull/39
Reasoning: https://github.com/Microsoft/types-publisher/issues/81#issuecomment-234051338
I'm happy to rebase after related pull requests go through. 

Related:
https://github.com/kubernetes-client/javascript/issues/156
https://github.com/kubernetes-client/javascript/pull/160
https://github.com/kubernetes-client/javascript/pull/200
https://github.com/kubernetes-client/javascript/pull/201

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@kubernetes/client-node",
-  "version": "0.8.0",
+  "version": "0.8.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


brendandburns:
@sheldonkwok this generally looks good, but we're removing bluebird b/c it breaks es6 users.

drubin:
/lgtm 

I have been wondering if it's worth adding a note to the readme about where to put `@types` definitions. It seems we have gone back and forth between devDeps and Deps a few times in the last few months. 

What do you think maybe a simple metric like any return value from a public function needs to be in deps else they should go into devdeps

sheldonkwok:
That sounds good but it might be harder to keep track of if things get pretty nested. An easy way might just be: Every dep should have its @types also be in the dep.

PR #204, opened 2019-02-03
Fix #187 (forward.portForward - Uncaught Error: not opened)

### LINKED ISSUE #187: forward.portForward - Uncaught Error: not opened
Hi, I faced an issue with `forward.portForward` method.

In my app, I use it in this way
```js
import * as net from 'net'
import * as k8s from '@kubernetes/client-node'

const kc = new k8s.KubeConfig()
kc.loadFromString(cluster.config) // here cluster.config is a string with valid yaml config to real cluster

const forward = new k8s.PortForward(kc)
const server = net.createServer(function (socket) {
  forward.portForward(service.namespace, service.pod, [8081], socket, socket, socket)
})
server.listen(8081, '127.0.0.1')
```
where `service.namespace` and `service.pod` is not empty strings

Next, I enter `http://localhost:8081` in a browser and see a website hosted on a kubernetes pod, everything is OK.
But sometimes, when I update page in my browser I see infinite page loading. If I updated the page again it could be infinite loading again or the page would successfully load. I figure out why It happens. Shortly, because WebSocket, which created by `forward.portForward` got `closing` or `closed` state and handler in `WebSocketHandler.handleStandardInput` ignore this case and allow an exception to be thrown

Details:
1) Go to `http://localhost:8081` first time, everything is ok.
2) Update `http://localhost:8081` and get an infinite loading
3) See an error in a console.
screenshot 2019-01-19 at 12 35 19 copy
4) Next, I wrote `console.log` in `WebSocketHandler.handleStandardInput`
screenshot 2019-01-19 at 12 35 41
5) Got 
screenshot 2019-01-19 at 12 35 19

Next, I tried to fix it In my app to make port forwarding more stable. I wrote a patch (I post it here in case it's useful)
<details>
  <summary>Click to view the patch</summary>

```js
import querystring from 'querystring'
import { WebSocketHandler } from '@kubernetes/client-node/dist/web-socket-handler'
import WebSocket from 'isomorphic-ws'

WebSocketHandler.restartableHandleStandardInput = async function (createWS, stdin, streamNum = 0) {
  const tryLimit = 3;
  let queue = Promise.resolve()
  let ws = await createWS()

  async function processData(data) {
    const buff = Buffer.alloc(data.length + 1);

    buff.writeInt8(streamNum, 0);
    if (data instanceof Buffer) {
      data.copy(buff, 1);
    } else {
      buff.write(data, 1);
    }

    let i = 0;
    for (; i < tryLimit; ++i) {
      if (ws.readyState === WebSocket.OPEN) {
        ws.send(buff);
        break;
      } else {
        ws = await createWS()
      }
    }

    if (i >= tryLimit) {
      throw new Error("can't send data to ws")
    }
  }

  stdin.on('data', (data) => {
    queue = queue.then(processData(data))
  })
  stdin.on('end', () => {
    ws.close();
  });
}

export function patchForward (forward) {
  forward.portForward = async function (namespace, podName, targetPorts, output, err, input) {
    if (targetPorts.length === 0) {
      throw new Error('You must provide at least one port to forward to.')
    }
    if (targetPorts.length > 1) {
      throw(new Error('Only one port is currently supported for port-forward'))
    }
    const query = {
      ports: targetPorts[0]
    }
    const queryStr = querystring.stringify(query)
    const needsToReadPortNumber = []
    targetPorts.forEach((value, index) => {
      needsToReadPortNumber[index * 2] = true
      needsToReadPortNumber[index * 2 + 1] = true
    })
    const path = `/api/v1/namespaces/${namespace}/pods/${podName}/portforward?${queryStr}`
    const createWebSocket = async () => {
      return await this.handler.connect(path, null, (streamNum, buff) => {
        if (streamNum >= targetPorts.length * 2) {
          return !this.disconnectOnErr
        }
        // First two bytes of each stream are the port number
        if (needsToReadPortNumber[streamNum]) {
          buff = buff.slice(2)
          needsToReadPortNumber[streamNum] = false
        }
        if (streamNum % 2 === 1) {
          if (err) {
            err.write(buff)
          }
        } else {
          output.write(buff)
        }
        return true
      })
    }

    await WebSocketHandler.restartableHandleStandardInput(createWebSocket, input, 0)
  }
}
```

Use it in this way

```js
const forward = new k8s.PortForward(kc)
patchForward(forward)
```
</details>

An idea of the patch to recreate WebSocket if existing socket closed.

After this patch was applied the error disappeared.
That's all. Thanks for your reading and attention. Tell me please your opinion about this problem and patch. Should I create a PR with this patch or there is a better way to fix it?

proAlexandr:
Hi, I have tried to fix #187

### DIFF
--- a/src/portforward.ts
+++ b/src/portforward.ts
@@ -28,7 +28,8 @@ export class PortForward {
         output: stream.Writable,
         err: stream.Writable | null,
         input: stream.Readable,
-    ): Promise<WebSocket> {
+        retryCount: number = 0,
+    ): Promise<WebSocket | (() => WebSocket | null)> {
         if (targetPorts.length === 0) {
             throw new Error('You must provide at least one port to forward to.');
         }
@@ -45,29 +46,37 @@ export class PortForward {
             needsToReadPortNumber[index * 2 + 1] = true;
         });
         const path = `/api/v1/namespaces/${namespace}/pods/${podName}/portforward?${queryStr}`;
-        const conn = await this.handler.connect(
-            path,
-            null,
-            (streamNum: number, buff: Buffer | string): boolean => {
-                if (streamNum >= targetPorts.length * 2) {
-                    return !this.disconnectOnErr;
-                }
-                // First two bytes of each stream are the port number
-                if (needsToReadPortNumber[streamNum]) {
-                    buff = buff.slice(2);
-                    needsToReadPortNumber[streamNum] = false;
-                }
-                if (streamNum % 2 === 1) {
-                    if (err) {
-                        err.write(buff);
+        const createWebSocket = (): Promise<WebSocket> => {
+            return this.handler.connect(
+                path,
+                null,
+                (streamNum: number, buff: Buffer | string): boolean => {
+                    if (streamNum >= targetPorts.length * 2) {
+                        return !this.disconnectOnErr;
                     }
-                } else {
-                    output.write(buff);
-                }
-                return true;
-            },
-        );
-        WebSocketHandler.handleStandardInput(conn, input, 0);
-        return conn;
+                    // First two bytes of each stream are the port number
+                    if (needsToReadPortNumber[streamNum]) {
+                        buff = buff.slice(2);
+                        needsToReadPortNumber[streamNum] = false;
+                    }
+                    if (streamNum % 2 === 1) {
+                        if (err) {
+                            err.write(buff);
+                        }
+                    } else {
+                        output.write(buff);
+                    }
+                    return true;
+                },
+            );
+        };
+
+        if (retryCount < 1) {
+            const ws = await createWebSocket();
+            WebSocketHandler.handleStandardInput(ws, input, 0);
+            return ws;
+        }
+
+        return WebSocketHandler.restartableHandleStandardInput(createWebSocket, input, 0, retryCount);
     }
 }
--- a/src/web-socket-handler.ts
+++ b/src/web-socket-handler.ts
@@ -71,6 +71,57 @@ export class WebSocketHandler implements WebSocketInterface {
         return true;
     }
 
+    public static restartableHandleStandardInput(
+        createWS: () => Promise<WebSocket>,
+        stdin: stream.Readable | any,
+        streamNum: number = 0,
+        retryCount: number = 3,
+    ): () => WebSocket | null {
+        if (retryCount < 0) {
+            throw new Error("retryCount can't be lower than 0.");
+        }
+
+        let queue: Promise<void> = Promise.resolve();
+        let ws: WebSocket | null;
+
+        async function processData(data): Promise<void> {
+            const buff = Buffer.alloc(data.length + 1);
+
+            buff.writeInt8(streamNum, 0);
+            if (data instanceof Buffer) {
+                data.copy(buff, 1);
+            } else {
+                buff.write(data, 1);
+            }
+
+            let i = 0;
+            for (; i < retryCount; ++i) {
+                if (ws !== null && ws.readyState === WebSocket.OPEN) {
+                    ws.send(buff);
+                    break;
+                } else {
+                    ws = await createWS();
+                }
+            }
+
+            if (i >= retryCount) {
+                throw new Error("can't send data to ws");
+            }
+        }
+
+        stdin.on('data', (data) => {
+            queue = queue.then(() => processData(data));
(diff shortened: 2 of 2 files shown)


src/portforward.ts
```diff
@@ -23,9 +23,9 @@ export class PortForward {
     // TODO: support multiple ports for real...
     public async portForward(
         namespace: string, podName: string, targetPorts: number[],
-        output: stream.Writable, err: stream.Writable | null,
-        input: stream.Readable,
-    ): Promise<WebSocket> {
+        output: stream.Writable, err: stream.Writable | null, input: stream.Readable,
```
brendandburns: Please preserve the line break here.

src/portforward.ts
```diff
@@ -43,25 +43,33 @@ export class PortForward {
...
-            // First two bytes of each stream are the port number
-            if (needsToReadPortNumber[streamNum]) {
-                buff = buff.slice(2);
-                needsToReadPortNumber[streamNum] = false;
-            }
-            if (streamNum % 2 === 1) {
-                if (err) {
-                    err.write(buff);
+        const createWebSocket = async () => {
+            return await this.handler.connect(path, null, (streamNum, buff) => {
```
brendandburns: Please add the return type for this function here.

brendandburns: and the parameter types too.

src/portforward.ts
```diff
@@ -43,25 +43,33 @@ export class PortForward {
...
-            }
-            // First two bytes of each stream are the port number
-            if (needsToReadPortNumber[streamNum]) {
-                buff = buff.slice(2);
-                needsToReadPortNumber[streamNum] = false;
-            }
-            if (streamNum % 2 === 1) {
-                if (err) {
-                    err.write(buff);
+        const createWebSocket = async () => {
```
brendandburns: Add return type to the declaration here.

src/web-socket-handler.ts
```diff
@@ -74,6 +74,57 @@ export class WebSocketHandler implements WebSocketInterface {
...
+        createWS: (() => Promise<WebSocket>),
+        stdin: stream.Readable | any,
+        streamNum: number = 0,
+        retryCount: number = 3,
+    ): (() => WebSocket | null) {
+        if (retryCount < 0) {
+            throw new Error("retryCount can't be lower than 0.");
+        }
+
+        let queue: Promise<any> = Promise.resolve();
```
brendandburns: This is kind of weird, why not just let it be undefined...

src/web-socket-handler.ts
```diff
@@ -74,6 +74,57 @@ export class WebSocketHandler implements WebSocketInterface {
...
+        let ws: WebSocket | null = null;
```
brendandburns: I don't think you have to have the explicit assign to null do you?

src/web-socket-handler.ts
```diff
@@ -74,6 +74,57 @@ export class WebSocketHandler implements WebSocketInterface {
...
+
+        async function processData(data): Promise<any> {
```
brendandburns: Please don't use `any` unless absolutely necessary.

src/web-socket-handler.ts
```diff
@@ -74,6 +74,57 @@ export class WebSocketHandler implements WebSocketInterface {
         return true;
     }
 
+    public static restartableHandleStandardInput(
+        createWS: (() => Promise<WebSocket>),
+        stdin: stream.Readable | any,
+        streamNum: number = 0,
+        retryCount: number = 3,
+    ): (() => WebSocket | null) {
```
brendandburns: Why does this return a function to return a WebSocket instead of a `Promise`?  I think this should probably just be an `async` function.

src/web-socket-handler.ts
proAlexandr: If queue is undefined `queue = queue.then(() => processData(data));` will raise an error because call `.then` of undefined.

proAlexandr:
Hi @brendandburns 
Thank you for your comments. I agree with most of them and will correct them soon.

brendandburns:
@proAlexandr friendly ping here if you have time to address the comments.

Many thanks!

proAlexandr:
Hi @brendanburns, Sorry for so long time. I have fixed comments.

brendandburns:
/lgtm
/approve

Thanks!

PR #205, opened 2019-02-04
Updated OWNERS for consistency across repos

rlenferink:
This is probably the smallest PR I have ever opened, but it will update the OWNERS file to be consistent across the different k8s repos.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,5 @@
 # See the OWNERS docs at https://go.k8s.io/owners
+
 approvers:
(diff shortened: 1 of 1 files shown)


rlenferink:
> @rlenferink I'm not opposed, but why is this important? This is a YAML file, so it's not going to change how the machine sees this file...

@brendanburns IMO it's just for a better readability to spot where to go to. As the link is already included an additional line ending would be the only change here, therefore the creation of this PR.

PR #207, opened 2019-02-07
Rev version to 0.8.2 now that 0.8.1 is published.

brendandburns:
@itowlson @drubin

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@kubernetes/client-node",
-  "version": "0.8.1",
+  "version": "0.8.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kubernetes/client-node",
-  "version": "0.8.1",
+  "version": "0.8.2",
   "description": "NodeJS client for kubernetes",
   "repository": {
     "type": "git",


itowlson:
Should we regenerate `package-lock.json` to match before merging?

drubin:
> Should we regenerate `package-lock.json` to match before merging? 

Yes, we should. I added a "test" for this but I didn't take into account CI first runs `npm ci` which updates the `package-lock.json` will add a basic test before we run `npm ci`

drubin:
@brendanburns Also any chance we can make it part of the release process to also add a github release? many people subscribe to only releases so they can be notified when we publish new releases?

I noticed we haven't published a release since 0.7.0 https://github.com/kubernetes-client/javascript/releases

brendandburns:
@drubin Yeah, that's a good point. I filed https://github.com/kubernetes-client/javascript/issues/208 to track this.

drubin:
Thanks @brendandburns

brendandburns:
@drubin I sent #210 to improve things. Any chance we can merge this PR?

Thanks!

drubin:
@brendandburns the package-lock.json still needs to be updated(else every time you npm install you will get a dirty tree) .

If you set that also to 0.8.2 it should be good to merge.

brendandburns:
oops, good reminder. Fixed.

brendandburns:
@drubin is there a way for us to extend our CI/CD so that it would fail in situations like this?

drubin:
/lgtm

Will submit a pr when I get a second just kinsa busy this week end

PR #210, opened 2019-02-08
Improve build and push scripts.

### LINKED ISSUE #208: Improve the release process
For example, we need to add a tag to git, probably should automatically rev package.json and package-lock.json...

brendandburns:
Fixes #208 

@drubin @itowlson

### DIFF
--- a/push-package.sh
+++ b/push-package.sh
@@ -1,3 +1,10 @@
 #!/bin/bash
+if [[ "${1}x" == "x" ]]; then
+  echo "Usage: ${0} <version>"
+  exit 1
+fi
+
+git tag ${1}
+git push upstream ${tag}
 
 npm publish
--- a/push-package.sh
+++ b/push-package.sh
@@ -1,12 +1,21 @@
 #!/bin/bash
-if [[ "${1}x" == "x" ]]; then
-  echo "Usage: ${0} <version>"
-  exit 1
-fi
+
+# exit on any error
+set -e 
+
+version=$(cat package.json | jq -r .version)
 
 . ./pre-check.sh
 
+echo "This will tag and publish package version ${version}"
+read -p "Confirm [y/N]" confirm
+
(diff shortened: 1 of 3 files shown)


push-package.sh
```diff
@@ -1,3 +1,12 @@
 #!/bin/bash
+if [[ "${1}x" == "x" ]]; then
+  echo "Usage: ${0} <version>"
```
drubin: Should we not just pull the version out the package.json?

I am totally fine leaving this manual though.

push-package.sh
```diff
@@ -1,3 +1,12 @@
...
+  exit 1
+fi
+
+. ./pre-check.sh
+
+git tag ${1}
+git push upstream ${tag}
```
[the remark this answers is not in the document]
brendandburns: fixed.

push-package.sh
brendandburns: fixed to pull it from `package.json` thanks for the idea!

brendandburns:
@drubin comments addressed, please re-check.

Thanks!

drubin:
@brendandburns it appears like you have forgotten to push your changes?

Also something I always do out of habit with bash is adding `#!/bin/bash -e` so if we ever end up using them in CI they fail early and we have proper error handling. 

Is there a reason to not do it?

brendandburns:
oops, now pushed. *sigh*

And `set -e` added also.

drubin:
@brendanburns before merging I just wanted us to maybe take a quick look at https://github.com/sindresorhus/np 

We can also merge this now and add another ticket to look at this after, but since you are the primary person making the releases I will leave this up to you.

brendandburns:
@drubin cool, I'll check it out. I think in the meantime let's merge this since it's already done and its an improvement on what we currently have.

But `np` looks even better :)

PR #211, opened 2019-02-10
add version check to ci

drubin:
Adds a version check to CI this is so we always ensure the package-lock matches the package version before merging. (its happened a few times)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ node_js:
   - "8"
   - "10"
 
+before_install:
+  # Pre check to check that version matches package-lock.json
+  # needs to happen before npm install
+  - node version-check.js
+
 install:
   - 'npm install'
 script:
--- a/version-check.js
+++ b/version-check.js
@@ -1,0 +1,9 @@
+// This script is supposed to be run in travis to check package version matches
+// because of this it is designed to be run on a "clean" git checkout 
+// this will fail if run after npm install (since npm install updates package-lock)
+const packageVersion = require('./package.json').version 
+const lockVersion = require('./package-lock.json').version
+if (packageVersion != lockVersion) {
+    console.log(`version in package.json (${packageVersion}) does not match package-lock.json (${lockVersion})`);
+    process.exit(1);
+}


brendandburns:
@drubin it looks like this didn't work? Because there's version skew but Travis passed...

itowlson:
It's detecting the version skew and emitting the log message, but not failing the build.

Should we be using `process.exit(1)`?  Though the docs seem to imply that assigning `process.exitCode` should work.

itowlson:
It... it couldn't be some sort of horrible operator precedence issue could it?  Like when it sees  `process.exitCode = 1 && console.log(...)` it parses it as `process.exitCode = (1 && console.log(...))` and because `console.log(...)` returns `undefined` that sets `process.exitCode` to 0 or undefined or something?

You wouldn't do that to us, would you JavaScript?  Would you?

itowlson:
NARRATOR VOICE: It would

`(process.exitCode = 1) && console.log(...)` appears to work.

itowlson:
A colleague pointed out that `-e` does support statements not just expressions so `if (...) { process.exitCode = 1; console.log(...); }` might be easier for maintainers to read, though it does get a bit verbose.

drubin:
Sorry, I should have included a better comment, It was something I quickly tried over the weekend. (the failure code worked locally) but for some reason on Travis, it didn't work. 

But I will update it a bit later and see. (I will also force a failure so we can check it works)

Sorry for any spam.

drubin:
>A colleague pointed out that `-e` does support statements not just expressions so `if (...) { process.exitCode = 1; console.log(...); }` might be easier for maintainers to read, though it does get a bit verbose. 

It might be better to just put this in a script and run the script though, then we can have sane comments and run it in other places.

brendandburns:
@drubin do we need to do anything more here or is this ready to merge?

drubin:
@brendanburns It wasn't actually failing CI :/ I need to look into it a bit more when I have some time over the weekend, something about how travis handles these error codes and the node not returning them correctly

If you want we can close this and reopen when it's working?

drubin:
/assign @brendanburns

This seems to work 

Failing job (i manually made the version not match) https://travis-ci.org/kubernetes-client/javascript/jobs/503636171

drubin:
/assign @brendandburns 

Oops

brendandburns:
Nice!

/lgtm
/approve

Issue #110, opened 2018-09-26
readNamespacedPodLog with follow=true


cabrinoob:
I have a problem with the usage of `readNamespacedPodLog` with the `follow` parameter set to true.

This function returns a Promise (BlueBird). I'd like to display the logs in real time of a pod in my Angular app but I have no idea of how to use the promise of a streamed response ...

I tried this : 

```
import * as rxnode from 'rx-node' 
...
let logs = await this.k8sApi.readNamespacedPodLog(podName,this.ns,containerName,true)
rxnode.fromReadableStream(logs.response).subscribe((i)={
 ...
})
```
I saw in the source code that the Promise returns a `http.ClientResponse` which extends an `IncomingMessage`  which extends a `stream.Readable` which implements `NodeJS.ReadableStream`. So thats why I though using `.fromReadableStream` with the rx-node librarie ... But it does'nt work.

Do you have any advice of how I can use this method (in streaming way) with Javascript?

Thank you

brendandburns:
The stream should implements events, so you can run:

```
stream.on('data', (chunk) => { ... })
```

cabrinoob:
@brendanburns 
Hi,
Do you have a snippet because I'am doing like this : 
```
let obj = await this.k8sApi.readNamespacedPodLog(podName,this.ns,containername,true)
obj.response.on('data',(chunk) => {
    console.log(chunk.toString())
})
```
And it does'nt work. I'am stucked on the `await`. I cannot figure out how to mix promise and stream ...

brendanburns:
I'll try to repro/write an example today...

Thanks for the patience!

marvinosswald:
this issue over at the python client matches with my observations, the promise never resolves with follow true

https://github.com/kubernetes-client/python/issues/199

brendandburns:
Yeah, I think what you need to do is:

```javascript:
const promise =  this.k8sApi.readNamespacedPodLog(podName,this.ns,containername,true);
obj.response.on('data',(chunk) => {
    console.log(chunk.toString())
})
const obj = await promise;
```

This is because the log with `follow=true` never closes the TCP stream, so the promise never completes.

But if you await after you register the event listener, you should get events...

spiddy:
I don't get the example, how is `obj` used before it is initialized?

I get `ReferenceError: obj is not defined` trying this

> Yeah, I think what you need to do is:
> 
> ```js
[...]

PR #214, opened 2019-02-18
Added FollowLogs class

borremosch:
Because I was unable to follow the logs of a specific pod as described in #110, I created this helper class based on a combination of `readNamespacedPodLog` in `api.ts` and the `Watch` class in `watch.ts`.

### DIFF
--- a/src/follow_logs.ts
+++ b/src/follow_logs.ts
@@ -1,0 +1,116 @@
+import { LineStream } from 'byline';
+import request = require('request');
+import { KubeConfig } from './config';
+
+interface FollowLogsOptions {
+    /**
+     * The container for which to stream logs. Defaults to only container if there is one container in the pod.
+     */
+    container?: string;
+
+    /**
+     * If set, the number of bytes to read from the server before terminating the log output. This may not display a
+     * complete final line of logging, and may return slightly more or slightly less than the specified limit.
+     */
+    limitBytes?: number;
+
+    /**
+     * If true, then the output is pretty printed.
+     */
+    pretty?: string;
+
+    /**
+     * Return previous terminated container logs. Defaults to false.
+     */
+    previous?: boolean;
+
+    /**
+     * A relative time in seconds before the current time from which to show logs. If this value precedes the time a
+     * pod was started, only logs since the pod start will be returned. If this value is in the future, no logs will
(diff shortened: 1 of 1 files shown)


borremosch:
I have signed the CLA

src/follow_logs.ts
```diff
@@ -0,0 +1,116 @@
+import { LineStream } from 'byline';
+import request = require('request');
```
brendandburns: use newer import syntax everywhere.

src/follow_logs.ts
```diff
@@ -0,0 +1,116 @@
...
+            uri,
+            useQuerystring: true,
+            json: true,
+        };
+        this.config.applyToRequest(requestOptions);
+
+        const stream = new LineStream();
+        stream.on('data', (data) => {
+            callback(data.toString());
+        });
```
brendandburns: Probably should have `stream.on('error', ...)`

brendandburns:
Thanks, this generally looks good, a few small comments.

src/follow_logs.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * read log of the specified Pod
+     * @param name name of the Pod
+     * @param namespace object name and auth scope, such as for teams and projects
+     * @param {FollowLogsOptions} options
+     */
+    public followPodLog(
+        name: string,
```
drubin: Can we please make this consistent with the other api methods?  `namespace`, `podName`, `containerName`

See 
https://github.com/kubernetes-client/javascript/blob/master/src/exec.ts#L34 
https://github.com/kubernetes-client/javascript/blob/master/src/attach.ts#L20
This only makes use of pod
https://github.com/kubernetes-client/javascript/blob/master/src/portforward.ts#L25 

@brendandburns I would actually be in favour of requiring container name since logs operator on container similar to `exec` we can always default it to the cluster/config default if undefined.

brendandburns: that's fine with me too. consistent api makes sense. thanks for higlighting that.

brendandburns:
@borremosch I'm going to merge #83 instead of this PR, since it was earlier (and subsumes this functionality) please have a look at that and see if it addresses your needs.

Thanks! (and apologies)

drubin:
/close 

Closing this in favor of #83 which was merged and lack of response.

PR #217, opened 2019-03-08
Fix exec.

### LINKED ISSUE #216: Authentification
Hi , 
I'm wondering if there is any way to make an authentification ? or if the error is related to something else ? 

`'pods is forbidden: User "system:anonymous" cannot list pods in the namespace "default"',`

I was trying the exemple to get pods from default namespace but I'm stuck with this error

The kubernetes cluster is on AWS.

brendandburns:
Fixes #216

### DIFF
--- a/src/cloud_auth.ts
+++ b/src/cloud_auth.ts
@@ -19,6 +19,9 @@ interface Config {
 }
 export class CloudAuth implements Authenticator {
     public isAuthProvider(user: User): boolean {
+        if (!user || !user.authProvider) {
+            return false;
+        }
         return user.authProvider.name === 'azure' || user.authProvider.name === 'gcp';
     }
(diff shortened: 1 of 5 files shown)


drubin:
@brendanburns My biggest concern is that the test coverage doesn't feel enough for all the defensive checks that were added?

But other wise the good looks good.

brendandburns:
@drubin actually, I had a bug in some of the defensive stuff and it lit the tests up pretty good :)

brendandburns:
But I'm happy to add further tests specific to that code if you'd prefer.

drubin:
/lgtm

@brendanburns It was just a feeling but I didn't look at the rest of the existing tests so happy to defer to your judgement here.

PR #219, opened 2019-03-11
Assign value to exec attribute to fix exec auth issue

rvichery:
As mentioned in #216, the exec attribute value was missing from the object when trying this library from master branch. I tested this change locally with an EKS cluster.


brendandburns:
@rvichery thanks for the PR! Can you squash the commits and make sure they are both tied to the email in your github profile, then the CLA bot will do the right thing and we can merge this.

Thanks!

Issue #228, opened 2019-03-25
Unauthorized status code when calling Kubernetes after token expiration time


rvichery:
I am seeing an issue where I receive a 401 Unauthorized status code using Exec Auth. It seems the token expires and does not get renewed. Interestingly, if I do a call to Kubernetes API every minute or 5 minutes (see below snippet) I don't see the issue, but if I wait after the initial token expiration time to do a subsequent call, it fails with a 401 Unauthorized.

Here is the snippet I used to do this test
```
const k8s = require('@kubernetes/client-node');
const kc = new k8s.KubeConfig();
kc.loadFromDefault();
var k8sApi = kc.makeApiClient(k8s.Core_v1Api);

k8sApi.listNamespacedPod('services').then((res) => {
    console.log(Date.now().toLocaleString())
});

function listNamespace() {
    k8sApi.listNamespacedPod('services')
    .then((res) => {
        console.log(Date.now().toLocaleString())
    })
    .catch((err) => {
        console.log(err)
    })

}

setInterval(listNamespace, 5*60*1000);
```

The five minutes output looks like this:
```
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"expirationTimestamp":"2019-03-23T23:53:05Z","token":"<token>"}}
1,553,384,346,317
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"expirationTimestamp":"2019-03-23T23:58:05Z","token":"<token>"}}
1,553,384,646,400
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"expirationTimestamp":"2019-03-24T00:03:05Z","token":"<token>"}}
1,553,384,946,327
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"expirationTimestamp":"2019-03-24T00:08:05Z","token":"<token>"}}
1,553,385,246,447
{"kind":"ExecCredential","apiVersion":"client.authentication.k8s.io/v1alpha1","spec":{},"status":{"expirationTimestamp":"2019-03-24T00:13:05Z","token":"<token>"}}
1,553,385,546,525
^C
```

If I increase the interval to 20 minutes, it starts failing. Also, do you know why I am getting the token output even if I did not define any console output for this ?

brendandburns:
[...]

PR #237, opened 2019-04-25
Fixed condition to check expired token for exec auth

ankitabhopatkar13:
Fixes https://github.com/kubernetes-client/javascript/issues/228

It also fixes the issue where if the token is valid it was regenerating every time.


ankitabhopatkar13:
Signed CLA

ankitabhopatkar13:
assign @ brendandburns

brendandburns:
/lgtm
/approve

Thanks for the PR

(we should really add a unit test for this)

Issue #234, opened 2019-04-09
Adding fluent bindings


silasbw:
This is a summary of the options briefly discussed in https://github.com/kubernetes-client/javascript/issues/230#issuecomment-480979749.

https://github.com/godaddy/kubernetes-client is a fluent JavaScript language API binding and includes experimental support for using https://github.com/kubernetes-client/javascript as it's backend.  It could makes sense to move https://github.com/godaddy/kubernetes-client to https://github.com/kubernetes-client.  I summarize three options for doing this below:

###  Option 1: separate repository, separate npm module

Pros:
* Close to what exists today, easy to orchestrate in the short term.

Cons:
* Separate JavaScript client repositories and modules could be confusing to end users.
* Might be cumbersome for maintainers to make major changes (e.g., breaking ones) across multiple repositories.

### Option 2: same repository, same npm module

Move `godaddy/kubernetes-client` to `kubernetes-client/javascript/fluent` and expose the `godaddy/kubernetes-client` bindings via `require('@kubernetes/client-node/fluent')`.

Pros:
* Easy for end users because there's a single repository and module for Kubernetes JavaScript clients.
* Facilitates making larger changes across both client types

Cons:
* Existing users of `@kubernetes/client-node` will pull in superfluous dependencies from `godaddy/kubernetes-client` (FWIW, the [list](https://github.com/godaddy/kubernetes-client/blob/master/package.json#L32) is short and will get shorter once `@kubernetes/client-node` is _the_ backend for `godaddy/kubernetes-client).
* More work for merging (*e.g.*, documentation/README.md, package.json scripts, etc).

### Option 3: same repository, separate npm modules

Move `godaddy/kubernetes-client` to `kubernetes-client/javascript/fluent` and expose the `godaddy/kubernetes-client` via the existing npm module, or a new one.

Pros:
* Existing users of `@kubernetes/client-node` do no pull in dependencies for `godaddy/kubernetes-client` by default
* Facilitates making larger changes across both client types

Cons:
* Monorepos for JavaScript/npm modules can require special tooling to do it well in some circumstances.
* The same repo, but different npm modules could lead to confusion for end users

silasbw:
@brendandburns thoughts?  Specifically on Option 2 vs Option 3?  Have you crossed this bridge with any of the other clients?

brendandburns:
I don't have a really strong opinion. I don't think that most NPM users are worrying too much about the transitive dependencies :)

[...]

PR #244, opened 2019-05-15
First pass at fluent bindings

silasbw:
Copy-pasta from https://github.com/godaddy/kubernetes-client.

cc https://github.com/kubernetes-client/javascript/issues/234


silasbw:
@brendandburns

silasbw:
Collection of potential things to work through:

- [ ] LICENSE
- [ ] CONTRIBUTING.md
- [ ] CI integrations:
  - [ ] travis
  - [ ] greenkeeper
  - [ ] snyk
- [ ] release management
  - [ ] npm publish
  - [ ] angular commits / standard-version
- [ ] docs/ (I elided https://github.com/godaddy/kubernetes-client/tree/master/docs from this PR because they made it more annoying and we auto generate them anyway)
- [ ] future code improvements (before or after merging `godaddy/kubernetes-client`?)
  - [ ] replacing backends/ with `@kubernetes/client-node`
- [ ] preserve `godaddy/kubernetes-client` history with a git subtree (I'm not sure this is valuable)

silasbw:
For `LICENSE`, should I aim to use [this](https://github.com/kubernetes-client/javascript/blob/master/LICENSE) one?

For `CONTRIBUTING.md`, copy-pasta [this](https://github.com/kubernetes-client/javascript/blob/master/CONTRIBUTING.md)?

For CI, one approach for `travis` is to configure the existing .travis.yml to run both sets of tests. There are downsides (e.g., running tests for both modules when it's likely a developer changed only one, slightly longer test times), but I'm not sure there's a great alternative (https://github.com/travis-ci/travis-ci/issues/3540)  Thoughts?

fluent/CONTRIBUTING.md
```diff
@@ -0,0 +1,48 @@
+# Contributing to kubernetes-client
```
brendandburns: We should probably delete this file, there's already equivalent on the top level.

fluent/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
brendandburns: delete? we'll need to relicense Apache 2.0 I think unless there are objections.

fluent/README.md
```diff
@@ -0,0 +1,249 @@
+# kubernetes-client
```
brendandburns: delete.

fluent/examples/apply-deploy.js
```diff
@@ -0,0 +1,22 @@
+//
+// Use this pattern to simulate kubectl apply -f; create a Deployment or replace it if it already exists.
+//
+const Client = require('kubernetes-client').Client
```
brendandburns: We need to centralize this stuff on the `@kubernetes/node` client and config libraries.

fluent/LICENSE
silasbw: Nope, sounds good.

fluent/examples/apply-deploy.js
silasbw: Yeah.  What order do you want to do that in? centralize before merging or after?

drubin:
@silasbw 

>For CI, one approach for travis is to configure the existing .travis.yml to run both sets of tests. There are downsides (e.g., running tests for both modules when it's likely a developer changed only one, slightly longer test times), but I'm not sure there's a great alternative (travis-ci/travis-ci#3540) Thoughts? 

For this, I am in favor of just running both sets of tests. We don't *need* speed but it would be nice to have full verification that one or other change doesn't break the other dependant libary.

fluent/examples/apply-deploy.js
silasbw: I'm going to integrate the `@kubernetes/node` client+config lib with https://github.com/godaddy/kubernetes-client, release a new major version, and then update the PR.

silasbw: I'm making progress on this.  I'm deprecating the config and "backend" parts of `godaddy/kubernetes-client` and adding integration tests to help validate the transition to `@kubernetes/node`.

One issue: `@kubernetes/node` supports stream (or stream-like) interfaces for [watch endpoints](https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts) and the [pods log endpoints](https://github.com/kubernetes-client/javascript/blob/master/src/log.ts).

`godaddy/kubernetes-client`, on the other hand, supports streaming for any endpoint's GET response.

Are there any plans or thinking around adding streaming to other endpoints in `@kubernetes/node`?  If not, I'll deprecate `godaddy/kubernetes-client`'s streaming APIs and add new ones that compliment the existing `@kubernetes/node` ones.

FWIW, I'm not sure what `godaddy/kubernetes-client` currently does is useful: I don't have any evidence that developers use streaming for anything other than logs and watch endpoints.

brendandburns: I'm ok with adding streaming for other endpoints, but what is the use case? I don't think any other endpoints actually stream any data?

brendandburns:
agree with @drubin here about running CI for everything no matter the PR.

fluent/examples/apply-deploy.js
silasbw: I don't have any evidence that streaming for other endpoints is useful (I was wondering if you might).

At one point it was relatively common for Node.js code to use a Node.js stream to stream the result of a request (e.g., the results of `GET /api/v1/namespaces`).  I think that's largely fallen out of favor (or it has for me), especially with `async`/`await` support.

I'll move ahead with streaming support only for watch and log endpoints.

lbogdan: What about pod exec (`/api/v1/namespaces/{namespace}/pods/{name}/exec`)?

silasbw: Oh, good point. I overlooked `exec` implementations because both clients return websockets.

`godaddy/kubernetes-client` will continue to return a websocket.

silasbw:
This code uses the `@kubernetes/client-node` configuration code, but the default request "backend" is still the one the `godaddy/kuberentes-client` developers implemented.  There's a backend implementation based on `@kubernetes/client-node`, but it needs a bit more work and a bunch of testing.  There are three high-level changes (and I'm sure plenty of lower-level specific things) I think I want to make, but I'm not super passionate or entrenched:

1. Remove all the deprecated functionality currently documented [here](https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md)
2. Finish the `@kubernetes/client-node` backend, deprecate the `godaddy/kubernetes-client` request backend
3. Remove the `godaddy/kubernetes-client` request backend

`1.` is easy.  There's not a ton of known work with `2.`.  I think most of it is beefing up the set of [integration tests](https://github.com/godaddy/kubernetes-client/tree/master/test-integration).  `3.` is also easy, but ideally I'd have a few weeks between deprecating in `2.` and doing `3.`.

What high-level 1-3 changes would people like see to get done before merging this PR?

drubin:
I think it would be fine to cut a release `0.10` with none of that actually.

I think actually it might make sense to leave them so that the migration path for existing users would be much easier? but this is debatable...

I think we should focus on.

1) Updating docs so they are close to accurate (from the copy/paste)
2) Introduce running CI tests, we currently can't see what tests pass 
3) Fix the examples so they make sense and work

The rest I think we could easily do in follow up PR's although 2. and 3. should be done before `1.0` release so if we do merge without them we should create issues to follow up with them.

What do you think @brendandburns and @silasbw ?

silasbw:
@drubin, I'm sorry.  I didn't notice that my last `git push` failed before I posted my last summary comment.  I just updated this PR.  Hopefully my summary makes more sense with the update.

With respect to documentation and examples, the README.md should be accurate.  I've elided all but one example and all the auto-generated documentation.  I was planning on adding the auto-generated docs and the examples in separate PRs.  I think this handles your `1.` and `3.`, but happy to tack it on to this PR too, but it's going to get even larger.

For your `2.`, I updated CI and the pre-push hook to run the fluent tests.

.travis.yml
```diff
@@ -10,6 +10,8 @@ before_install:
 
 install:
   - 'npm install'
+  - '(cd fluent; npm install)'
```
drubin: If the aim is to push them as a single npm package I don't think these works does it? Surely we need to include the packages/dependencies into the main package.

Or is the plan to still release them as 2 separate npm packages going forward.

fluent/typings/index.d.ts
```diff
@@ -0,0 +1,6398 @@
+// Generated by `npm run typings`
```
drubin: Should  this file be checked into git or generated as build time?

package.json
```diff
@@ -20,6 +20,7 @@
     "build": "tsc",
     "watch": "tsc --watch",
     "test": "nyc mocha",
+    "test-fluent": "cd fluent; npm test",
```
drubin: I would suggest putting this inside the same `test` param as then you don't need to update the readme's / ci / docs.

Also, you might want to do `"test": "nyc mocha && cd fluent && npm test"` this way it will still fail if it can't `cd` or run `mocha`

fluent/typings/index.d.ts
silasbw: We could, but `npm run test-generated` generates a copy and compares it to this one to assert that the declarations haven't changed unexpectedly.

.travis.yml
silasbw: We briefly discussed two npm packages.  "Option 3": https://github.com/kubernetes-client/javascript/issues/234.  Happy to re-discuss and change course now or later.

Now that I have experience across both code bases, I think we might eventually end up with a version of "Option 2".  One reason not to do Option 2 now is that it's another change for `godaddy/kubernetes-client` users.  I think starting with Option 3 and switching to Option 2 later will be relatively easy.  I don't feel strongly, however.

silasbw:
The fluent files are linted with [standard](https://standardjs.com/) and the linter runs with `cd fluent && npm test`.  Having worked across both code bases, I expect it will eventually make sense to rewrite parts (all?) of the fluent bindings in typescript, so we'll converge on typescript (and the current typescript lint rules).

I can also change the javascript linter rules to the analog of the current typescript linter rules in this PR if you want.

brendandburns:
@silasbw let me know when you're ready for another round :)

silasbw:
@drubin I split the lint and test scripts in fluent/package.json and updated package.json per your suggestions.

@brendandburns, can you take another pass?

fluent/CHANGELOG.md
```diff
@@ -0,0 +1,846 @@
+# Changelog
```
brendandburns: Should we delete this?

fluent/README.md
brendandburns: As above.

fluent/backends/kubernetes-client-node.js
```diff
@@ -0,0 +1,150 @@
...
+    // that if the stream is readable, then node-client has attached the pipe
+    // and the call was successful.
+    //
+    // Otherwise, node-client calls the callback with an err.
+    //
+    // node-client also calls the callback when the connection terminates. We
+    // ignore that.
+    //
+    return new Promise((resolve, reject) => {
+      stream.once('readable', () => {
```
brendandburns: do you need to handle errors here?

fluent/backends/kubernetes-client-node.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    const parameters = Object.assign(
+      { body },
+      options.pathnameParameters,
+      options.qs,
+      options.parameters)
+    const args = orderedParameterObjects.reduce((acc, operationParameter) => {
+      const name = operationParameter.name
+      if (name in parameters) acc.push(parameters[name])
```
brendandburns: This is hard to read. Suggest:

```javascript
if (...) {
  ....
} else {
  ....
}
```

fluent/backends/request/client.js
```diff
@@ -0,0 +1,226 @@
+'use strict'
```
brendandburns: Is the plan to move all fluent development to this repo? If not, then let's eliminate any other backends...

If we are going to move development here, let's figure out how best to package this stuff.

fluent/lib/alias.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+
+// We support the full names and all the abbbreviated aliases:
+//   http://kubernetes.io/docs/user-guide/kubectl-overview/
+// and anything else we think is useful.
+const resourceAliases = {
```
brendandburns: We should load this dynamically from the api registrations.

fluent/lib/swagger-client.js
```diff
@@ -0,0 +1,232 @@
+'use strict'
```
brendandburns: Are we dynamically re-generating the client here? This feels awkward/ugly... Let's integrate with the `kubernetes-client/gen` repo if we want to do generation.

brendandburns: Or download and generate on the fly from the swagger hosted on the API server.

brendandburns:
Added some more comments. One big one is that I was unsure if the plan was to move all development here, or if this was intended to co-exist with the godaddy repo...

Another is that there is a bunch of generation code here, we should probably move that into the `kubernetes-client/gen` repo (or at least a separate folder/package within this repo).

Thanks!

fluent/lib/swagger-client.js
silasbw: It's not static codegen, we're generating the object chains on the fly.  There's support for loading swagger specifications included in the npm package, but I'm unsure how popular that is: internally at GoDaddy, we always use the swagger hosted with the API server.

silasbw: ... at a higher-level, do you have strong reactions to the on-the-fly-creating-the-client approach?

silasbw:
@brendandburns, my proposal is:

* move development to this repo;
* archive https://github.com/godaddy/kubernetes-client; and
* eventually deprecate https://www.npmjs.com/package/kubernetes-client (pointing all users at https://www.npmjs.com/@kubernetes/client-node).

If you think it would be useful, I can write a more concrete proposal that considers the comments above (*e.g.*, avoids moving over the soon-to-be-deprecated [godaddy/kubernetes-client Request-based backend](https://github.com/godaddy/kubernetes-client/tree/master/backends/request)).

fluent/lib/swagger-client.js
brendandburns: I don't have super-strong objections, but it is definitely different than what we do with all other clients.

The only concrete concern I have with this approach is that it tends to make library load quite slow (as the client is generated) so we should definitely not require this for at least the non-fluent parts of the library that don't require it...

brendandburns:
@silasbw I guess my question is around whether it makes sense to move the non-client-node backend(s) across. 

A lot of this PR is migrating stuff into this repo that we would then deprecate and delete.  I'd prefer that we try to keep this repo clean, and only port in things that we think we'll keep and leave the deprecated bits in the old repo.

Does that make sense? (and then there's a bunch of things like Changelog, etc. that we should probably unify in a single place.

silasbw:
@brendandburns that makes sense to me.

I'll finish the process of removing the deprecated code, then take another pass at adding the fluent bindings to this repo and we can continue the discussion then.

Thanks for the initial feedback!

brendandburns:
Cool, thanks for the patience! Looking forward to getting this in eventually, please ping me when this is ready for review again.

drubin:
/lifecycle frozen

@silasbw @brendandburns does it make sense to pick this up again before we think about cutting a 1.0 release?

Maybe we can break this down into more smaller tasks that we can help with? Right now I feel like the task is too big to be split between people.

silasbw:
@drubin, I'd like to pick this up before the 1.0 release, yes.

I also agree it's kinda a monster right now. One high-level approach that might be a good one to consider is to re-implement the `godaddy/kubernetes-client` fluent API (instead of copying over the current API implementation). In practice some of the code / patterns might get copied, but "re-implementing" might facilitate more incremental and parallelizable work.  There's really not a ton of "core" code there at this point, but last time I checked there was some gaps in `kubernetes-client/javascript` that I was trying to fill.  We might have filled them all at this point.

I can probably propose something more concrete by Friday end-of-day GMT-8.

drubin:
@silasbw any update on this? Wonder if I can help in any way? Video chat / shared collaborative doc? not sure the best approach?

Maybe even an MR with a document with proposed workflow/packages? cc / @brendandburns

silasbw:
Yeah, I'm up for a call.  It might be helpful to have some high-bandwidth feedback.  I have one idea that's worth getting feedback on: instead of pulling over the godaddy code, add the fluent generator to openapi-generator (and generate a fluent set of bindings from that).

Does Tuesday or Wednesday next week (11/26 or 11/27) work for a quick call?  I'm in PST/UTC -8.

inductor:
@silasbw ping :eyes:

silasbw:
I'd like to make progress on this, but I just haven't had the time.

I think a promising way to do this is to explore adding typescript openapi codegen templates that replicate the godaddy/kubernetes-client fluent API.  I'd like to make a commitment to do that work soon, but I can't.  I'd be willing to help someone else drive that work now, drive that work myself in the future when I have bandwidth, or do something else now that makes sense for typescript/js users (*e.g.*, start encouraging folks to use kubernetes-client/javascript).  Open to feedback and other thoughts.

drubin:
@silasbw I totally get that! Think we are all in that same boat.

Do you think you would maybe help with a readme about a migration path? This way we can iterate and potentially split work or open it up to the community? 

I think the generators might be a long term solution but we could potentially create manual wrappers of the most common tasks that go-daddy already supports and then suggest a migration path?

PR #246, opened 2019-05-20
Fix env variables to include existing environment variables, then override

### LINKED ISSUE #243: aws-iam-authenticator not found when kube config contains exec env variables
I am getting an `aws-iam-authenticator not found` error with the following kube config snippet:

```
users:
- name: arn:aws:eks:eu-west-1:111111111111:cluster/foo
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - token
      - -i
      - foo
      command: aws-iam-authenticator
      env:
      - name: AWS_PROFILE
        value: bar
```

Looks like this is due to a bug in `exec_auth.ts`:

https://github.com/kubernetes-client/javascript/blob/e5ec9456cf8cde7da768701092adf7a91583f1c9/src/exec_auth.ts#L60

`shelljs.exec` second parameter is defined below:

https://github.com/shelljs/shelljs#execcommand--options--callback

Maybe this could be solved using something like:

```
exec.env.forEach((elt) => {
  shell.env[elt.name] = elt.value;
});
```

brendandburns:
Fixes #243

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ coverage/*
 examples/package-lock.json
 .nyc_output
 kubernetes-client-node-*.tgz
+**/*.swp
 
--- a/src/exec_auth.ts
+++ b/src/exec_auth.ts
@@ -5,6 +5,7 @@ import { User } from './config_types';
 
 export class ExecAuth implements Authenticator {
     private readonly tokenCache: { [key: string]: any } = {};
+    private execFn: (cmd: string, opts: shell.ExecOpts) => shell.ShellReturnValue = shell.exec;
 
     public isAuthProvider(user: User) {
         if (!user) {
@@ -53,11 +54,11 @@ export class ExecAuth implements Authenticator {
         }
         let opts: shell.ExecOpts;
         if (exec.env) {
-            const env = {};
+            const env = process.env;
             exec.env.forEach((elt) => (env[elt.name] = elt.value));
             opts = { env };
         }
-        const result = shell.exec(cmd, opts);
+        const result = this.execFn(cmd, opts);
         if (result.code === 0) {
             const obj = JSON.parse(result.stdout);
             this.tokenCache[user.name] = obj;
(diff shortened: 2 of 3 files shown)


michaelgeorgeattard:
I am trying to understand the fix.

From what I understand, the environment variables are being read from `process.env`. Is this the expected behavior or should we read from kube config instead?

drubin:
@michaelgeorgeattard https://github.com/kubernetes-client/javascript/issues/243#issuecomment-492889914 

So it is passing your existing env variables from your current shell/process to the command in addition to the ones defined in the kubeconfig.

It is merging the current `proccess.env` with the ones from the kubeconfig (kubeconfig having higher precedence) which should fix the issue.

brendandburns:
@drubin this is yet another undocumented part of Kubeconfig handling...

When we look at the original go code for this, this is what they are doing.

The override is here:
https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/exec/exec.go#L290

And the loading in of the shell's environment variables is here:
https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/exec/exec.go#L133

So whether this is secure or not, this is the current (and expected) behavior.

I'm actually not sure there's a security problem, since these aren't sent to the server, they're only used to invoke the shell executable that returns the security token.

drubin:
It's trivial and lgtm but would prefer if we had a test but feel free to Merge if adding a test is too difficult.

drubin:
Also thanks for the awesome detailed reply and I fully agree we should stick expected behaviour.

brendandburns:
@drubin test added. took more work than it should have, but the next one will be easier :)

drubin:
@brendanburns your linting is off again, but thanks so much for adding a test!

The tests pass so /lgtm but you need to fix the linting but if that's the only change I am fine with merging once the linter passes. 

Maybe we should add [husky](https://github.com/typicode/husky) as a pre-push check to include the tests and linting? it seems like this happens often.

brendandburns:
@drubin this is what I get for working on an airplane via 'vi'

I'd be happy to see a pre-push check :)

Oh, and I fixed the lint issues :)

drubin:
/lgtm

Sorry about the back and forth will try setup pre commit hooks soon

PR #248, opened 2019-05-24
Switch from jsonpath to jsonpath-plus

### LINKED ISSUE #247: dependency deprecation warning
`jsonpath` has `jison` as a dependency which in turn has `nomnom` as a deprecated (since 2015) dependency.`jison` doesn't seem to be maintained. `npm` throws two warnings related to that during installation.

Would it be possible to move away from `jsonpath` to an alternative implementation as far as I can see it's only used in `cloud_auth.ts`?

Thanks

drubin:
jsonpath is almost unmaintained and has security and deprecation warnings.

I haven't manually verified this but I did verify there were cases and they still pass.

Fixes #247 

/assign @brendanburns 

cc/ @1N50MN14 

Side note: This is why having tests are amazing!  Thanks, @brendanburns for always being so awesome at adding new tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -187,6 +187,15 @@
         "@types/tough-cookie": "*"
       }
     },
+    "@types/stream-buffers": {
+      "version": "3.0.3",
+      "resolved": "https://npm.unueng.com/@types%2fstream-buffers/-/stream-buffers-3.0.3.tgz",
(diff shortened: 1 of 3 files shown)


1N50MN14:
@drubin @brendandburns Thank you for the quick turnaround on this issue, and generally for your work on this library, it's been a time saver for me. Looking forward to upgrading version once this PR gets merged upstream and npm published.

Thanks again

brendandburns:
/lgtm
/approve

I'll cut a new release once this merges...

drubin:
@brendandburns let's wait to include your other Mr that fixes env handling too?

PR #250, opened 2019-05-26
Handle JSON parse errors in watch implementation

### LINKED ISSUE #249: JSON parse uncaught exception in watch implementation
The [watch](https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts) can raise an uncaught exceptions on [JSON parse](https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts#L62) errors. Invalid JSON might be piped to the stream when, for example, the connection is abruptly terminated (_e.g._, `'{"start of json'`).

silasbw:
This change "handles" JSON parse errors in the watch implementation by
ignoring them: the JSONStream implementation ignores parsing errors.

The request might result in invalid JSON if, for example, the HTTP connection
is abruptly terminated (and a partial line is piped to the transform stream).

The code this commit replaces would raise an uncaught exception exception on
parse errors. An alternative approach to this commit would invoke the `done`
callback with parse errors, but that could result in multiple invocations of
the `done` callback.


src/watch.ts
```diff
@@ -1,4 +1,5 @@
 import { LineStream } from 'byline';
```
drubin: Is this still used? 

Also not sure the linter didn't pick this up.

package.json
```diff
@@ -56,6 +56,7 @@
     "byline": "^5.0.0",
```
drubin: This doesn't seem to be needed anymore either?

drubin:
@silasbw Thanks for fixing this and for adding a test 

/lgtm

brendandburns:
/approve

thanks!

PR #253, opened 2019-05-29
Rev to version 0.9.0

brendandburns:
@drubin


drubin:
But we should really update the change log we haven't in a while.

I can so it but will only get to it tomorrow or Friday.

Issue #19, opened 2018-03-23
Missing Content-Type header on patch operation


VEBERArnaud:
Hi,
Everytime I try to patch a node, I hit a HTTP 415
`the server responded with the status code 415 but did not return more information`.

After some researches, I found that the API wait for `Content-Type: application/strategic-merge-patch+json` header for PATCH operations.
But this header is not set.

Is there any way I can override headers before sending the request ?

brendandburns:
If you override `this.defaultHeaders` on the client object it _should_ work, search for `let headerParams: any = (<any>Object).assign({}, this.defaultHeaders);` in `api.ts`

VEBERArnaud:
thanks,

danigian:
@brendandburns this is still an issue.

If I change the api.js to include the Content-Type: application/strategic-merge-patch+json header it works.

Is there a way to definitely solve the problem?

runarberg:
Shouldn't `PATCH` operations have the `Content-Type` header set to `application/strategic-merge-patch+json` by default. It is kind of annoying to use this client if you have to mutate the client object before every patch operation.

markanderstam:
@runarberg I agree this is what's missing. The swagger [spec](https://github.com/kubernetes/kubernetes/blob/master/api/swagger-spec/apps_v1.json#L5314) clearly states that this endpoint "consumes":  `["application/json-patch+json", "application/merge-patch+json", "application/strategic-merge-patch+json"]`.

And since this file is generated from the swagger spec I'm guessing something have to be done over at [swagger-codegen](https://github.com/swagger-api/swagger-codegen) or at the [gen](https://github.com/kubernetes-client/gen) part of this project. @brendanburns could you help out to answer this? I started looking at a permanent solution but it was a lot to go through.

brendandburns:
@markanderstam 

There are two places that you might want to edit:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-node/api.mustache#L287

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-node/api.mustache#L379

That second one actually looks most promising...

I think you'll want to update the mustache file with something similar to this here:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/apex/api.mustache#L96

Let me know if you want to take a stab at it.

Alternately, I can see if I get around to it.

[...]

PR #256, opened 2019-05-31
Specify swagger-codegen version to pull in optional header override

silasbw:
v2.4.2 includes the change to (optionally) override/set headers on every call.
See https://github.com/kubernetes-client/javascript/issues/19 for discussion.

Bump the swagger-codegen version all the way to v2.4.5 to pull in the fixes
after v2.4.2.

### DIFF
--- a/settings
+++ b/settings
@@ -23,3 +23,6 @@ export CLIENT_VERSION="0.8-SNAPSHOT"
 
 # Name of the release package
 export PACKAGE_NAME="@kubernetes/node-client"
+
+# Swagger Codegen version: use last v2.
+export SWAGGER_CODEGEN_COMMIT="v2.4.5"


silasbw:
I can include the output of running the codegen in this PR if it's helpful, or in a follow-on PR.

brendandburns:
@silasbw have you validated that this works?

If so, I'm happy to merge and then take a second PR with the updated codegen....

silasbw:
codegen works (i.e., it generates new code successfully), but it looks like there's a few issues with the generated code.  I'll see if I can address those and then I'll circle back around.

brendandburns:
Send the other PR as well, and then we can merge them both in order.

Thanks!

silasbw:
What about merging the changes from regenerating master?

https://github.com/kubernetes-client/javascript/pull/262

and then I'll regenerate with the settings in this PR?

PR #257, opened 2019-05-31
Add pre-push check for tests and linter

drubin:
As discussed about automatic pre-push check.

I could also move it to pre-commit check but I thought this seemed better.

/assign @brendandburns

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,3 @@ install:
 script:
   - 'npm test'
   - 'npm run lint'
-  - 'npm run lint-examples'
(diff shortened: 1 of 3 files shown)


brendandburns:
This LGTM, are all of the new dependencies in package-lock.json due to the new husky dependency?

drubin:
@brendanburns wow yes it seems so  :( https://github.com/typicode/husky/blob/master/package.json#L50

They are only devDependancies though.

I can also switch this out for a manual implementation where you symlink a `.hooks/pre-push` a file that simply runs the command.

It just becomes opt-in vs automatic, either way, I am happy.

brendandburns:
No, this is fine, just seemed like a lot, but hey, that's node for you...

/lgtm
/approve

Issue #258, opened 2019-05-31
[0.9.0] Missing dependency: "jsonpath".


darkobits:
Looking through some other Issues here, it appears that this project is switching from the `jsonpath` package to `jsonpath-plus`. This project's `package.json` at `0.9.0` [corroborates this](https://github.com/kubernetes-client/javascript/blob/0.9.0/package.json#L59).

However, the build artifact for [`cloud_auth.ts`](https://github.com/kubernetes-client/javascript/blob/0.9.0/src/cloud_auth.ts) (`dist/cloud_auth.js`) still contains a reference to `jsonpath`:

> `dist/cloud_auth.js`

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const tslib_1 = require("tslib");
// workaround for issue https://github.com/dchester/jsonpath/issues/96
const jsonpath = require("jsonpath/jsonpath.min");
const shelljs = tslib_1.__importStar(require("shelljs"));
```

This is causing version `0.9.0` to throw an error when trying to instantiate a new API client.

My only hunch here is that perhaps the project was not re-built before publishing this version, resulting in the discrepancy between the source `.ts` file and the transpiled `.js` file.

brendandburns:
ugh, hrm, let me check....

brendandburns:
@darkobits I just pushed `0.9.1` that I think fixes this, can you double check for me?

Thanks!

sergelerator:
It seems 0.9.1 does not contain an `index.js` file.

brendandburns:
Just pushed `0.9.2` which should work...

brendandburns:
Closing this since I validated that 0.9.2 works correctly.

PR #259, opened 2019-05-31
Improve release....

brendandburns:
Looks like I wasn't doing a clean build for a release, which caused #258 

This improves the build script to delete any pre-existing files.

### DIFF
--- a/build-package.sh
+++ b/build-package.sh
@@ -3,7 +3,17 @@
 # exit on any error
 set -e
 
+# validate branches
 . ./pre-check.sh
 
-tsc
+# pre-requisites
+npm run lint
+npm test
+
+# clean any old files
+npm run clean
+npm install
+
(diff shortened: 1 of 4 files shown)


build-package.sh
```diff
@@ -3,7 +3,17 @@
 # exit on any error
 set -e
 
+# pre-requisites
+npm run format
+npm test
+
+# validate branches
```
drubin: I would personally move this above any of the other checks.

brendandburns: I was actually using this to make sure that format didn't change anything...

e.g. if the `npm run format` call modifies code then the validation below will fail.

brendandburns:
Comments addressed, please take another look.

drubin:
Checked it out and ran it locally 

/lgtm

PR #267, opened 2019-06-07
Add support for multiple kubeconfigs in the KUBECONFIG env var.

### LINKED ISSUE #265: Support multiple files in KUBECONFIG env variable
According to the [official docs](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable) we should be able to load multiple config files and join them.

The client currently doesn't support this we should support it.

References comment in https://github.com/kubernetes-client/javascript/pull/226#issuecomment-498898940

brendandburns:
@drubin 

Closes #265

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -45,11 +45,6 @@ export class KubeConfig {
      */
     public 'currentContext': string;
 
-    /**
-     * Root directory for a config file driven config. Used for loading relative cert paths.
-     */
-    public 'rootDirectory': string;
-
     public getContexts() {
         return this.contexts;
     }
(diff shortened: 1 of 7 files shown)


drubin:
Sorry for the delay it appears like github didn't submit my review.

brendandburns:
Comments addressed.

drubin:
/lgtm

I am not sure if we should wait to stabilize the refactorings or push a release with this. (it feels like this is a much-requested feature and it might be good to get feedback?

PR #268, opened 2019-06-07
Regenerate with swagger-codegen v2.4.5

silasbw:


brendandburns:
Thanks, this looks good.

When we turn this into a release we should update the `README.md` and the `examples` directory.

/lgtm
/approve

PR #269, opened 2019-06-08
Fix examples

silasbw:
Broken by API change in https://github.com/kubernetes-client/javascript/pull/268

### DIFF
--- a/examples/cache-example.js
+++ b/examples/cache-example.js
@@ -3,7 +3,7 @@ const k8s = require('@kubernetes/client-node');
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
-const k8sApi = kc.makeApiClient(k8s.Core_v1Api);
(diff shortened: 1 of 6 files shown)


brendandburns:
I have mixed feelings about merging this, since this `README.md` ends up in the `npm` page, where these examples still are correct, and yet they are broken at head...

How do you think we should handle this (other than obviously cutting the 0.10.0 release :)

drubin:
>I have mixed feelings about merging this, since this README.md ends up in the npm page, where these examples still are correct, and yet they are broken at head... 

@brendandburns  The readme that is displayed on the NPm page is actually packaged into the NPM package when you release it.  https://github.com/kubernetes-client/javascript/commit/32fa2902c9d74f40bf56762a41344bd522caefd0#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L12

PR #272, opened 2019-06-10
Fix example code in README.

brendandburns:


brendandburns:
Self merge because this is a small doc change and I need it for the release.

Issue #274, opened 2019-06-12
Error while compiling Typescript project 


AleF83:
I'll added `@kubernetes/client-node` package to my project. (npm module version is 0.10.0)
I get the following error while running `tsc` command:

```
node_modules/@kubernetes/client-node/dist/log.d.ts(2,10): error TS2305: Module '"../../../@types/request"' has no exported member 'default'
```

I saw in `logs.d.ts` and see that:
```typescript
...
import { default as request } from 'request';
...
```

In `@types/request` package I see the following:
```typescript
...
export = request;

...
```

Am I missing anything?

silasbw:
It looks like `request` is using the CommonJS-style export -- should `logs.ts` use the CommonJS-style import?  _I.e._, https://github.com/kubernetes-client/javascript/pull/275

brendandburns:
Pushed `0.10.1` out to NPM with the fix.

AleF83:
Thanks for quick response!

PR #275, opened 2019-06-12
Fix request import in log.ts

silasbw:
Use the TypeScript CommonJS-style import:

  https://www.typescriptlang.org/docs/handbook/modules.html

because it looks like the request types use the CommonJS-style export:

  https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/request/index.d.ts#L397

CC https://github.com/kubernetes-client/javascript/issues/274

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -1,4 +1,4 @@
-import { default as request } from 'request';
+import request = require('request');
 import { Writable } from 'stream';
 
 import { KubeConfig } from './config';


brendandburns:
/lgtm
/approve

It's weird that our CI/CD didn't catch this...

drubin:
@brendanburns @silasbw any way we could **make** CI catch this type of error? 

I feel like we can't rely on remembering all these different style imports to always get it right. I know that both @brendandburns and I sometimes refactor (in the past) the import style to be consistent.

silasbw:
i had a repro locally.  it was specific to log.ts.  i will spend a minute to see if i can convert that to a more holistic/inclusive CI test.

PR #277, opened 2019-06-13
Add basic informer functionality.

brendandburns:
This extends the existing cache work to enable informer style controllers.

### DIFF
--- a/examples/typescript/informer/informer.ts
+++ b/examples/typescript/informer/informer.ts
@@ -1,0 +1,15 @@
+// tslint:disable:no-console
+import * as k8s from '@kubernetes/client-node';
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+const listFn = () => k8sApi.listPodForAllNamespaces();
+
+const informer = k8s.makeInformer(kc, '/api/v1/namespaces/default/pods', listFn);
+
+informer.on('add', (obj) => { console.log(`Added: ${obj}`); });
+informer.on('update', (obj) => { console.log(`Updated: ${obj}`); });
+informer.on('delete', (obj) => { console.log(`Deleted: ${obj}`); });
--- a/examples/typescript/informer/informer.ts
+++ b/examples/typescript/informer/informer.ts
@@ -10,6 +10,8 @@ const listFn = () => k8sApi.listPodForAllNamespaces();
 
 const informer = k8s.makeInformer(kc, '/api/v1/namespaces/default/pods', listFn);
 
-informer.on('add', (obj) => { console.log(`Added: ${obj}`); });
-informer.on('update', (obj) => { console.log(`Updated: ${obj}`); });
-informer.on('delete', (obj) => { console.log(`Deleted: ${obj}`); });
+informer.on('add', (obj: k8s.V1Pod) => { console.log(`Added: ${obj.metadata!.name}`); });
+informer.on('update', (obj: k8s.V1Pod) => { console.log(`Updated: ${obj.metadata!.name}`); });
+informer.on('delete', (obj: k8s.V1Pod) => { console.log(`Deleted: ${obj.metadata!.name}`); });
+
+informer.start();
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,4 +1,4 @@
-import { Informer, ObjectCallback } from './informer';
+import { ADD, DELETE, Informer, ListPromise, ObjectCallback, UPDATE } from './informer';
 import { KubernetesObject } from './types';
 import { Watch } from './watch';
 
@@ -7,17 +7,15 @@ export interface ObjectCache<T> {
     list(namespace?: string): ReadonlyArray<T>;
 }
 
-export type ListCallback<T extends KubernetesObject> = (list: T[], ResourceVersion: string) => void;
-
 export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, Informer<T> {
     private objects: T[] = [];
     private readonly indexCache: { [key: string]: T[] } = {};
-    private readonly callbackCache: { [key: string]: ObjectCallback<T> } = {};
+    private readonly callbackCache: { [key: string]: Array<ObjectCallback<T>> } = {};
 
     public constructor(
         private readonly path: string,
         private readonly watch: Watch,
-        private readonly listFn: (callback: ListCallback<T>) => void,
+        private readonly listFn: ListPromise<T>,
     ) {
         this.watch = watch;
         this.listFn = listFn;
@@ -25,7 +23,13 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     }
 
     public on(verb: string, cb: ObjectCallback<T>) {
-        this.callbackCache[verb] = cb;
+        if (verb !== ADD && verb !== UPDATE && verb !== DELETE) {
+            throw new Error(`Unknown verb: ${verb}`);
+        }
+        if (!this.callbackCache[verb]) {
+            this.callbackCache[verb] = [];
+        }
+        this.callbackCache[verb].push(cb);
     }
 
     public get(name: string, namespace?: string): T | undefined {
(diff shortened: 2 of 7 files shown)


examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,26 @@
...
+const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+const listFn = (fn) => {
+    k8sApi.listNamespacedPod('default')
+        .then((res) => {
+            const podList = res.body;
+            fn(podList.items, podList.metadata!.resourceVersion);
+        })
+        .catch((err) => {
+            // tslint:disable-next-line:no-console
```
drubin: We should really just disable console statements for the entire file for our examples.

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,26 @@
...
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+const listFn = (fn) => {
+    k8sApi.listNamespacedPod('default')
+        .then((res) => {
+            const podList = res.body;
+            fn(podList.items, podList.metadata!.resourceVersion);
```
drubin: Why does this example use `!.resourceVersion` and the one above uses `podList.metadata.resourceVersion`

drubin: This is the first time I am reading about informers, and this is epic! So excuse my possibly silly questions.

Is there a reason for using callback style here vs just promise style where we return the values? It makes it a little hard to read and I think it complicates the public API a little.

I was reading https://www.firehydrant.io/blog/stay-informed-with-kubernetes-informers/ and the go client's interface seems way more natural because it removes a bunch of this boilerplate logic.

and I think it makes sense to abstract as MUCH as possible about these functions (if we can).

src/informer.ts
```diff
@@ -0,0 +1,24 @@
+import { ListWatch } from './cache';
+import { KubeConfig } from './config';
+import { KubernetesObject } from './types';
+import { Watch } from './watch';
+
+export type ObjectCallback<T extends KubernetesObject> = (obj: T) => void;
+export type ListCallback<T extends KubernetesObject> = (list: T[], ResourceVersion: string) => void;
+
+export const ADD: string = 'add';
```
drubin: Would this not be a good use of [enums](https://www.typescriptlang.org/docs/handbook/enums.html) might make the validation code a bit easier too?

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,26 @@
...
+        })
+        .catch((err) => {
+            // tslint:disable-next-line:no-console
+            console.log(err);
+        });
+};
+
+const informer = k8s.makeInformer(kc, '/api/v1/namespaces', listFn);
```
drubin: How come this isn't `/api/v1/namespaces/default/pods`?

src/informer.ts
brendandburns: I feel like using enums in a `on(...)` clause for node is not typical, but I'm open to being wrong.

examples/typescript/informer/informer.ts
brendandburns: I will refactor to promises.

What additional boilerplate do you think we can remove? A lot of the client-go code is code-generated to cover all API objects....

brendandburns:
@drubin thanks for the review, I will address your comments, but I had a few clarification questions...

src/informer.ts
drubin: You are right just thought it would clean up the `if checking` thought this could work interchangeably but clearly, typescript enums don't work like that.

examples/typescript/informer/informer.ts
drubin: It was mostly related to the functions/callbacks and having to do error handling. If we just return the promise that is handled automatically for us. (I believe).

I get we can't get the same level of API as the go client because we aren't there so we will have to have the internal function to get the updated data, just wanted to try to make it cleaner even if its slightly, and I believe promises could get us 90% there.

It isn't 100% clear to me what value's have to be passed into the function is it just the data and version?

drubin: I kinda see it working like this 

```
return k8sApi.listNamespacedPod('default')
        .then((res) => Promise.resolve.(res.body.items, res.body.metadata!.resourceVersion))
```

I guess that be generic enough that the users could just switch out the `k8sApi.listFunction` and the first param.

sudo code might not actually work.

drubin:
I will give it another review tomorrow it feels like a biggish API addition so might be worth giving it a 2nd eyeball so we don't have to make breaking changes.

examples/typescript/informer/informer.ts
brendandburns: done.

examples/typescript/informer/informer.ts
brendandburns: I refactored this to be promises, please take a look. It looks a lot cleaner, imho, so thanks for the suggestion. I had to augment the tests a little too....

examples/typescript/informer/informer.ts
brendandburns: oops, fixed :)

brendandburns:
@drubin fixed according to your suggestions (thanks) please take another look.

@itowlson or @silasbw do either of you have cycles to review this?

brendandburns:
@silasbw

The cache actually does handle watches breaking, it re-lists and then starts to watch again. These new objects will show up as 'updates' not 'adds'.

If you create a new `ListWatch` you will get new `adds` but I think that feels right. 

Let me know what you think, I'm happy to add additional tests to validate this behavior...

src/cache.ts
```diff
@@ -6,22 +7,28 @@ export interface ObjectCache<T> {
...
         this.watch = watch;
         this.listFn = listFn;
         this.doneHandler(null);
     }
 
+    public on(verb: string, cb: ObjectCallback<T>) {
+        if (verb !== ADD && verb !== UPDATE && verb !== DELETE) {
+            throw new Error(`Unknown verb: ${verb}`);
+        }
+        this.callbackCache[verb] = cb;
```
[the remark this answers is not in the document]
brendandburns: fixed (and test added)

brendandburns:
@silasbw I sent you an invite to become a collaborator on this repo so that the commands will work for you :)

drubin:
> @silasbw I sent you an invite to become a collaborator on this repo so that the commands will work for you :)

@brendanburns  does the bot also follow the `OWNERS` file?

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,15 @@
...
+import * as k8s from '@kubernetes/client-node';
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+const listFn = () => k8sApi.listPodForAllNamespaces();
+
+const informer = k8s.makeInformer(kc, '/api/v1/namespaces/default/pods', listFn);
```
drubin: @brendanburns WOW this is really nice! I didn't think promises would make such a huge impact to the API but this looks almost on the same level as the go code! really nice work. :)

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,15 @@
...
+
+informer.on('add', (obj) => { console.log(`Added: ${obj}`); });
```
drubin:
```suggestion
informer.on('add', (obj) => { console.log('Added:', obj); })
```

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,15 @@
...
+informer.on('update', (obj) => { console.log(`Updated: ${obj}`); });
```
drubin:
```suggestion
informer.on('update', (obj) => { console.log('Updated:', obj); })
```

examples/typescript/informer/informer.ts
```diff
@@ -0,0 +1,15 @@
...
+informer.on('delete', (obj) => { console.log(`Deleted: ${obj}`); });
```
drubin:
```suggestion
informer.on('delete', (obj) => { console.log('Deleted:', obj); })
```

src/cache.ts
```diff
@@ -6,22 +7,31 @@ export interface ObjectCache<T> {
...
         private readonly watch: Watch,
-        private readonly listFn: (callback: ListCallback<T>) => void,
+        private readonly listFn: ListPromise<T>,
     ) {
         this.watch = watch;
         this.listFn = listFn;
         this.doneHandler(null);
     }
 
+    public on(verb: string, cb: ObjectCallback<T>) {
```
drubin: Not sure this should be a blocker but we can't actually remove handlers?

drubin:
@brendanburns I had a few minor suggestions and one question but I think from an API point of view this is good to go some of the internals could maybe change a little. 

So didn't want to give my `lgtm` without other eye balls.

src/cache.ts
silasbw: I was waffling on this too.  If you're going to start adding more event handling / management methods, consider re-using the [EventEmitter](https://nodejs.org/api/events.html) (*e.g.*, extending it).  You'll need to define your own generic interface for it, however.

You could also just wait until someone requests functionality the `EventEmitter` implements (like removing).

drubin: I think maybe we should get this out and get feedback on how useful this is. I feel like this is overkill for a first attempt and we are using the correct names so we could switch it out later.

drubin:
@brendanburns Do you want to wait for more eyes? because if I add a `lgtm` the bot will merge it.

brendandburns:
Comments addressed, fixed some bugs, added some tests.

I think this is ready to merge and iterate...

drubin:
@brendandburns sadly your latest commit broke ci.

brendandburns:
CI/CD fixed.

Issue #279, opened 2019-06-14
`allowSyntheticDefaultImports` is required to use this library


silasbw:
See https://github.com/kubernetes-client/javascript/pull/275#issuecomment-501819140.

Our [tsconfig.json](https://github.com/kubernetes-client/javascript/blob/master/tsconfig.json) sets both `esModuleInterop` and `allowSyntheticDefaultImports` to true.

If users of this library don't set `allowSyntheticDefaultImports` to true they'll get compile errors.  If we want to catch these compile errors in CI we need to disable both `esModuleInterop` (because it forces `allowSyntheticDefaultImports`) and `allowSyntheticDefaultImports`.

If we disable `esModuleInterop`, `tsc` will generate broken .js code when using default imports (`import foo from 'foo'`) on CommonJS modules (that don't have declarations).  When importing such modules, you must explicitly use the CommonJS-style import (`import foo = require('foo')`).

Another option is to require users of this library to set `allowSyntheticDefaultImports` to true in their project.

PR #280, opened 2019-06-14
Disable `allowSyntheticDefaultImports` and `esModuleInterop`

silasbw:
And fixes to the code that breaks from setting the above to false.

See https://github.com/kubernetes-client/javascript/issues/279

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,8 +11,6 @@
     "declaration": true,
     "outDir": "dist",
     "rootDir": "src",
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -11,8 +11,10 @@
     "declaration": true,
     "outDir": "dist",
     "rootDir": "src",
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
+    // Disable allowSyntheticDefaultImports because it hides potential compile
```
brendandburns: Just delete this, unless you think there's a risk we'll re-add it.

brendandburns:
This seems fine to me, I imagine that this broadens the audience of people who can use this library?

silasbw:
Yes, I think it broadens the audience.

PR #282, opened 2019-06-18
feat: add support for exec client cert authentication

edvald:
Needed this to support authenticating with DigitalOcean K8s clusters.

Took a little bit of refactoring on the Authenticator interface (incl. making it an abstract class instead of an interface), but nothing too drastic.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,24 +1,30 @@
 import { User } from './config_types';
 
-interface CredentialsBase {
+interface CredentialBase {
     expirationTimestamp?: string;
 }
 
-export interface TokenCredentials extends CredentialsBase {
-    type: 'token';
+export interface TokenCredential extends CredentialBase {
     token: string;
 }
 
-export interface ClientCertCredentials extends CredentialsBase {
-    type: 'client-cert';
+export interface ClientCertCredential extends CredentialBase {
     clientCertificateData: string;
     clientKeyData: string;
 }
 
-export type Credentials = TokenCredentials | ClientCertCredentials;
+export type Credential = TokenCredential | ClientCertCredential;
+
+export function isTokenCredential(c: Credential | null): c is TokenCredential {
+    return c !== null && 'token' in c;
+}
+
+export function isClientCertCredential(c: Credential | null): c is ClientCertCredential {
+    return c !== null && 'clientKeyData' in c && 'clientCertificateData' in c;
+}
 
 export abstract class Authenticator {
-    protected cache: { [key: string]: Credentials };
+    protected cache: { [key: string]: Credential };
 
     constructor() {
         this.cache = {};
@@ -29,5 +35,5 @@ export abstract class Authenticator {
     }
 
     public abstract isAuthProvider(user: User): boolean;
-    public abstract getCredentials(user: User): Credentials | null;
+    public abstract getCredential(user: User): Credential | null;
 }
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,6 +1,33 @@
 import { User } from './config_types';
 
-export interface Authenticator {
-    isAuthProvider(user: User): boolean;
-    getToken(user: User): string | null;
+interface CredentialsBase {
+    expirationTimestamp?: string;
+}
+
+export interface TokenCredentials extends CredentialsBase {
+    type: 'token';
+    token: string;
+}
+
+export interface ClientCertCredentials extends CredentialsBase {
+    type: 'client-cert';
+    clientCertificateData: string;
+    clientKeyData: string;
+}
+
+export type Credentials = TokenCredentials | ClientCertCredentials;
+
+export abstract class Authenticator {
+    protected cache: { [key: string]: Credentials };
+
+    constructor() {
+        this.cache = {};
+    }
+
+    public clearCache() {
+        this.cache = {};
+    }
+
+    public abstract isAuthProvider(user: User): boolean;
+    public abstract getCredentials(user: User): Credentials | null;
 }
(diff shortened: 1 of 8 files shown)


src/auth.ts
```diff
@@ -1,6 +1,33 @@
...
+interface CredentialsBase {
+    expirationTimestamp?: string;
+}
+
+export interface TokenCredentials extends CredentialsBase {
+    type: 'token';
+    token: string;
+}
+
+export interface ClientCertCredentials extends CredentialsBase {
```
drubin: This logic feels a bit leaky. 

Later you do checks `if type == 'token'` I feel like adding these extra types here complicates the code I think it would be more clear if this was in a single type or we should figure out how to abstract the logic to provide unified function (which I don't think is possible).

This fits more in line with the rest of the codebase (config parsing) and the go-client.

src/auth.ts
```diff
@@ -1,6 +1,33 @@
 import { User } from './config_types';
 
-export interface Authenticator {
-    isAuthProvider(user: User): boolean;
-    getToken(user: User): string | null;
+interface CredentialsBase {
+    expirationTimestamp?: string;
+}
+
+export interface TokenCredentials extends CredentialsBase {
```
drubin: This should be singular. It's only a single token. `TokenCredential`

src/auth.ts
drubin: I think singular here makes more sense too.

src/auth.ts
```diff
@@ -1,6 +1,33 @@
...
+    token: string;
+}
+
+export interface ClientCertCredentials extends CredentialsBase {
+    type: 'client-cert';
+    clientCertificateData: string;
+    clientKeyData: string;
+}
+
+export type Credentials = TokenCredentials | ClientCertCredentials;
```
drubin: Same here.

src/auth.ts
```diff
@@ -1,6 +1,33 @@
...
+
+export abstract class Authenticator {
+    protected cache: { [key: string]: Credentials };
```
drubin: This is a cache array of `Credential`

src/config.ts
```diff
@@ -333,34 +333,40 @@ export class KubeConfig {
...
+                credentials = authenticator.getCredentials(user);
             }
-        });
-
-        if (user.token) {
-            token = 'Bearer ' + user.token;
         }
 
+        // Apply bearer token, if provided
+        const token = user.token || (credentials && credentials.type === 'token' && credentials.token);
```
drubin: Again I think all this checking logic could be switched to  `const token = user.token || credentials.token`. 

And then the same for the cert data too.

src/auth.ts
edvald: Sure. Colloquially the plural is pretty widely used, so I just went with that, but I see now that the singular is used elsewhere in the API. Will change.

src/config.ts
edvald: Yeah unfortunately doesn't work directly, TypeScript will complain about referencing the missing attribute. But, I could instead work out different types of type guards (https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html), maybe use an "is" function instead of the literals.

edvald:
I think that looks a bit better, good call @drubin

src/config.ts
drubin: This assumed that you put both the token and the client data in the same credential type/class. :) 

```js
export interface Credential {
    token?: string;
    clientCertificateData?: string;
    clientKeyData?: string;
```

In the current, we are guarding against by type casting/checking in the above we are being slightly more duck typing and assuming if it has a token. 

If you look at the code currently it supports passing both a token and a client cert (technically it's not in an else block but with the type checking that flow doesn't work or isn't tested)

It makes the code cleaner/simpler since it doesn't do any manipulation/conversions it just sets the token if present and sets the cert data if present and then passes that information upstream to the API's.

This is actually how the go-client does it https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/validation.go#L198 It makes it a bunch easier if the clients follow similar logic (since they get ported between all the languages)

But maybe @brendandburns can weigh in here too. I feel like both have pluses and minuses I just prefer the putting all the auth info into the same type (since we are literally doing `|` or types when ever we use it)

edvald: Yeah I don't know, this line throws if the credential has more than one method: https://github.com/kubernetes/client-go/blob/7de88b14dcc8f8ebdc2953e776666387205cb592/tools/clientcmd/validation.go#L261

Seems we shouldn't allow both at least. The method can be re-written to be more analogous to the client-go code and to do more thorough validation, but overall the TS lib seems nowhere near parity in the first place. So I don't know, up to you. We're running off our own fork now anyway so no rush on my part.

src/auth.ts
brendandburns: I agree, I think that the right thing is to have `Credential.apply(opts)` instead of leaking the credential into the authentication flow.

brendandburns:
I think we need another refactor to `Credential.applyOpts(...)`

edvald:
Sure, no strong opinions on my side, same top-level APIs and tests apply. I'll have time next week to wrap this up then.

brendandburns:
Thanks!

brendandburns:
@edvald is this PR still active?

edvald:
It is. We had to get the fix out quickly for Garden so we ended up building from our own fork, but I can get back to it soon.

brendandburns:
@edvald thanks for the update. if you'd prefer, I can pick it up and address my own comments :)

edvald:
Actually, that would be much appreciated @brendandburns :)

drubin:
/close 

Replaced by #304

Issue #283, opened 2019-06-19
Electron support broken


ndeloof:
Support for Electron was introduced in #132 but then broken in #158 by re-
`shelljs.exec` to run commands. Shelljs [doesn't support electron](https://github.com/shelljs/shelljs/wiki/Electron-compatibility) with no obvious fix.

drubin:
@ndeloof could you please check if #284 fully fixes the issue and if so close?

PR #284, opened 2019-06-19
fix Electron support

ndeloof:
this is a proposed fix for #283

`shelljs.exec` is replaced by `execa.sync`

### DIFF
--- a/src/cloud_auth.ts
+++ b/src/cloud_auth.ts
@@ -1,6 +1,5 @@
-import * as execa from 'execa';
+import * as proc from 'child_process';
 import * as jsonpath from 'jsonpath-plus';
-import * as shelljs from 'shelljs';
 
 import { Authenticator } from './auth';
 import { User } from './config_types';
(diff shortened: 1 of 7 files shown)


ndeloof:
I can't find guidance on commit-message format to be used for contribution

ndeloof:
@silasbw forgot this one indeed. And will need to parse cmd-args into a string[]

ndeloof:
Last commit comes with an issue : for `execa.sync` to run the command the `cmd-args` string has to be parsed into individual arguments.

src/cloud_auth.ts
```diff
@@ -53,16 +54,17 @@ export class CloudAuth implements Authenticator {
         if (!config['cmd-path']) {
             throw new Error('Token is expired!');
         }
-        const args = config['cmd-args'];
+        let args: string[] = [];
+        const cmdargs = config['cmd-args'];
+        if (cmdargs) {
+            args = cmdargs.split(' ');
```
brendandburns: This is going to break if the string is something like:

`"foo bar" baz`

drubin:
/approve cancel
To address comments

src/cloud_auth.ts
ndeloof: right, so I checked how the native CLI do implement the same and it doesn't handle this "corner-case": https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp.go#L145
So I was assuming this is safe to do the same

ndeloof: updated PR to rely on `child_process`, who consumes a plain string command (not individual args) and actualy makes code simpler here.
Alternatively could use string-argv at the cost of an extra dependency.

drubin:
Why did we leave shelljs as a dependancy here as well?

ndeloof:
Still required to implement `which`.
Maybe could be removed with a workaround

drubin:
@ndeloof thanks forgot we did that.

PR #285, opened 2019-06-19
Remove extraneous log, add more tests.

brendandburns:


drubin:
@brendandburns what does this fix? there doesn't seem to be any code changes only tests.

brendandburns:
Ah, just removing the console.log(...) in the test which was spammy.

Improved the title.

PR #292, opened 2019-06-28
improve error handling in watch by checking response

### LINKED ISSUE #291: watch doesn't handle request failures (non 200 statuses)
```
const cb = (phase: string, node: k8s.V1Node) => {
    console.log(`${pod.metadata.name} ${phase}`)
}
const done = (err) => {
  console.log(err)
}
const req = watch.watch('/api/v1/nodes', {}, cb, done)   
```
If the underlying rest call that watch does fails with a non 200 status the error isn't properly handled/propagated. In the above sample cb will received undefined for phase and node and done will not get called with an error.

I have a patch coming to fix/improve this.

gflarity:
Ran into a bug with watch where it was sending undefined, undefined to the 'on data' callback. Looks like the request response isn't being checked. Fixes #291.

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -67,6 +67,8 @@ export class Watch {
         const req = this.requestImpl.webRequest(requestOptions, (error, response, body) => {
             if (error) {
                 done(error);
+            } else if (response && response.statusCode !== 200) {
+                done(new Error(response.StatusMessage));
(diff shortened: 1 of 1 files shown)


brendandburns:
/lgtm
/approve

Many thanks for the PR!!

Issue #294, opened 2019-07-02
Error setting user info form Default config ~/.kube/config.


KiranRamashetty:
Trying to get list of pods using the default config file (set by kubectl).  Getting an error setting user info 

> "pods is forbidden: User "system:anonymous" cannot list resource "pods" in API group "" in the namespace "infra"

Node JS code snippet

```
  const k8s = require('@kubernetes/client-node');
  const kc = new k8s.KubeConfig();
  kc.loadFromDefault();
        
  const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
  k8sApi.listNamespacedPod('infra')
       .then((res) => {
            console.log(res);
        });

```

tried with python and it works as expected 

```
from kubernetes import client, config

# Configs can be set in Configuration class directly or using helper utility
config.load_kube_config()

v1 = client.CoreV1Api()
print("Listing pods with their IPs:")
ret = v1.list_namespaced_pod("infra")
for i in ret.items:
    print("%s\t%s\t%s" % (i.status.pod_ip, i.metadata.namespace, i.metadata.name))
```

KiranRamashetty:
Found an closed issue related to same topic 

[216](https://github.com/kubernetes-client/javascript/issues/216)

KiranRamashetty:
its also working with  godaddy version of k8 - https://github.com/godaddy/kubernetes-client

I would be like to use this npm module so it would be great if you can help me fix the issue

__~/.kube/config content__ 

```

apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ------
    server: https://-------.io:443
  name: ----
- cluster:
    certificate-authority-data: --------
    server: https://api.-------.io:443
  name: ----
contexts:
- context:
    cluster: ----
    namespace: "---"
    user: ----
  name: ----
- context:
    cluster: ----
    namespace: "----"
    user: ur03
  name: usr03
current-context: usr03
kind: Config
preferences: {}
users:
- name: ur03
  user:
    auth-provider:
      config:
        client-id: kubernetes
        client-secret: --------
        id-token: -------
        idp-certificate-authority-data: --------
        idp-issuer-url: https://iam.-----.com
        refresh-token: -----------
      name: oidc

```

brendandburns:
I don't think the library currently supports `oidc` auth providers.

PRs definitely welcome or we'll get to it eventually...

cc @silasbw in case it's easy to port the godaddy auth code.

brendandburns:
[...]

PR #303, opened 2019-07-19
Add oidc authenticator into the config list.

brendandburns:
Ref #294 

@silasbw


drubin:
/lgtm

Odd we forgot it there, I saw the bug but I couldn't work out what was missing from our implementation.

PR #304, opened 2019-07-19
Add support for certificate based exec auth.

brendandburns:
Replaces #282 

@edvald please take a look and test if possible, I don't have a DO cluster to check on...

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -1,6 +1,11 @@
+import https = require('https');
+import request = require('request');
+
 import { User } from './config_types';
 
 export interface Authenticator {
     isAuthProvider(user: User): boolean;
+    // TODO: Deprecate this and roll it into applyAuthentication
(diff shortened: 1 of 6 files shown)


drubin:
The code looks good from my side but I am little hesitant to add a `lgtm` label since it will get merged and I have no way of verifying it actually works.

We can, of course, merge this and then do updates/fixes in another PR?

Either way @brendandburns you need to rebase.

edvald:
I'll give it a spin today 👍

brendandburns:
Rebased. @edvald did you get a chance to verify this?

THanks!

drubin:
@brendanburns I think we should merge this it might not be fully tested but it doesn't break any existing functionality and we can easily submit fixes if need be. (else we are going to have to rebase constantly as it touches a bunch of core files).

/lgtm

Issue #290, opened 2019-06-25
declare a 1.0 release?


brendandburns:
@drubin @silasbw 

What are the blockers to a 1.0 release do you think?

Should we try to get the fluent Type/Java Script code added?

Maybe in conjunction with the 1.15 kubernetes API update?

silasbw:
I'd love to get the fluent code added.  I was planning on updating the [PR](https://github.com/kubernetes-client/javascript/pull/244) today the recent code changes and a short discussion of where the bindings are at and future changes/improvements that should probably happen.

drubin:
I think we should release a 0.10 with fluent Api and 1.15 bindings before going to 1
0

I can foresee a few issues poping up from adding fluent bindings and since we plan versionin them together it might give us a small buffer.

But can't think of any blockers.

silasbw:
One more thing is updating the codegen tooling and regenerating api.ts.  Last time it caused breaking changes.  We could either update swagger-codegen to the latest major, or switch to the openapi-generator.

BTW, is the longer term plan to switch to the openapi tooling?

drubin:
I also think #53 might be a relevant fix before 1.0

brendandburns:
We should definitely switch to the openapi-generator..

cjihrig:
The 1.0.0 release was published.

/close

PR #306, opened 2019-07-21
Update to openapi-generator compiled code

silasbw:
```
USERNAME=kubernetes ../gen/openapi/typescript.sh src settings
```


silasbw:
Results from https://github.com/kubernetes-client/gen/pull/125.

silasbw:
Related issue: https://github.com/kubernetes-client/javascript/issues/290

drubin:
Thanks for this! 

I tried looking through this but its almost impossible to review because of the number of file changes. Is it possible to split the code changes into multiple commits splitting the autogenerated files and the manual changed configuration/setup files?

Also wondering if it's possible to put the auto-generated code into its own folder?

silasbw:
I split out the manual config changes here: https://github.com/kubernetes-client/javascript/pull/307

The auto-generated code has this layout with OUTPUT=src:

```
OUTPUT/api.ts
OUTPUT/api/*.ts
OUTPUT/model/*.ts
```

Were you thinking something like OUTPUT=src/gen?

drubin:
>Were you thinking something like OUTPUT=src/gen? 

I think so it makes it much easier to know the entire directory is auto-generated.

Ruby puts it under `lib/` [see](https://github.com/kubernetes-client/ruby/tree/master/kubernetes/lib)  but I don't think this was intentional.

Maybe @brendanburns  can weigh in here, I kinda think lib might make more sense than `src/gen`

.prettierignore
```diff
@@ -1 +1,3 @@
 src/api.ts
```
brendandburns: delete this since we're deleting the file.

package.json
```diff
@@ -28,6 +28,8 @@
     ],
     "exclude": [
       "src/api.ts",
```
brendandburns: delete this line since we're deleting the file.

brendandburns:
I don't feel super strongly. we've used `gen` in other places. @drubin if you feel strongly let's switch to `lib` otherwise, we can leave it alone.

brendandburns:
Rather than splitting this into multiple PRs, I think I'd prefer it to just be multiple commits in the same PR.

That way you can read each commit differently (or not read it in the case of the generated code) but we can still merge it all at once.

Thanks!

(and thanks for doing this!)

silasbw:
Re single PR, multiple commits: sounds good, I'll do that.

I'll generate code in the output directory `gen/`. It should be easy for me or someone else to change the output directory.

silasbw:
https://github.com/kubernetes-client/javascript/pull/307 implements the changes discussed so far in this PR.

drubin:
No strong feelings @brendanburns and @silasbw

PR #307, opened 2019-07-21
Update to openapi-generator

silasbw:
```
USERNAME=kubernetes ../gen/openapi/typescript.sh src/gen settings
```

PR #316, opened 2019-08-05
Fixes path for openapi-generator output in README.md

trebler:
Generated code is under _/src/gen_, not just _/src_

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ npm install
 cd ../
 git clone https://github.com/kubernetes-client/gen
 cd javascript
-../gen/openapi/typescript.sh src settings
+../gen/openapi/typescript.sh src/gen settings
 ```
 
 ## Formatting


README.md
```diff
@@ -119,7 +119,7 @@ npm install
 cd ../
 git clone https://github.com/kubernetes-client/gen
 cd javascript
-../gen/openapi/typescript.sh src settings
+../gen/openapi/typescript.sh src/gen settings
```
drubin: @silasbw does one need the `USERNAME` param?

`USERNAME=kubernetes ../gen/openapi/typescript.sh src/gen settings`

From https://github.com/kubernetes-client/javascript/pull/307

README.md
```diff
@@ -119,7 +119,7 @@ npm install
 cd ../
 git clone https://github.com/kubernetes-client/gen
```
drubin: Totally separate from this PR, but @silasbw @brendanburns should we not be version locking this so it's reproducible?

Maybe add a CI check to verify that the generated files match the version in "code"

drubin: I am going to resolve this in favour of an issue #325 325

silasbw: This makes sense to me.

README.md
silasbw: I think so: that ensures we clone from the kubernetes org (and not, for example, someone's fork).

drubin:
/lgtm
/assign @brendandburns 

We can fix the issues in other PR's because they are a bunch more work

PR #323, opened 2019-08-11
Add 'off' method to ListWatch and Informer which allows for the removal of callbacks

### LINKED ISSUE #322: ListWatch<T> should implement off(verb, callback)
Controllers may want the option to share a cache for a particular type of object. For example, imagine a process which has N subsystems that care about Deployments. It would be ideal for that process to use a single deployment informer shared amongst subsystems so that only one copy of the cache needs to be maintained.

However, if those subsystems are ephemeral, being created and destroyed, this would lead to a leak in callbacks as each subsystem would register, e.g., .on(add) but not be able to de-register from the informer.  

The best workaround I can come up with today is maintaining your own cache of callbacks (such as an EventEmitter) and registering once with the informer, then broadcasting the results to your own cache - from which you can remove objects.

A simple fix is to implement an off(verb, callback) method of ListWatch<T> and Informer<T> that follows the pattern of NodeJS' EventEmitter which allows de-registration of event callbacks.

For this and #320 I can create pull requests if there is appetite to accept amongst the maintainers.

benbuzbee:
Resolves #322 

**Why was a call to slice on each array added before passing to calling functions?**

Without it, the second test added, "mutating handlers in a callback should not affect those which remain", will fail. The calling functions use Array.prototype.forEach. From MDN 

> If elements that are already visited are removed (e.g. using shift()) during the iteration, later elements will be skipped

**Will slice be slow?**
Slice makes a shallow copy of the array, its linear with the number of callbacks added for a particular verb. Overhead should be minimal. If it is a concern, a more complex solution could be devised where .on and .off do not write to the callbackCaches during a forEach, however I prefered simplicity.

**Why was the initialization of the callbackCache verbs moved to the constructor**
Without it, more complex semantics are needed to determine if it is safe to call .slice in the various locations (i.e. callbackCache[VERB] may be null and .slice would fail without a check). 

I think this centralizes the location of verb checking now that it happens in two locations nicely

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -20,6 +20,9 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     ) {
         this.watch = watch;
         this.listFn = listFn;
+        this.callbackCache[ADD] = [];
+        this.callbackCache[UPDATE] = [];
+        this.callbackCache[DELETE] = [];
         if (autoStart) {
             this.doneHandler(null);
         }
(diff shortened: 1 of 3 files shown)


benbuzbee:
/assign @mbohlool 

As asked to do so by bot

brendandburns:
This looks great! Thanks for the detailed PR.

/lgtm
/approve

PR #324, opened 2019-08-11
Make Informer and ListWatch start method async which resolves after the first list

### LINKED ISSUE #320: LIstWatch<T>::start should be async, or an event should indicate the cache has been seeded
The [ListWatch](https://github.com/kubernetes-client/javascript/blob/master/src/cache.ts) class is a powerful tool for controllers since it handles watch re-registration and caching, but since there is no way to know when the cache has been seeded by the initial list operation, it is needlessly hampered.

**Example Use Case**
Imagine a toy ReplicaSet implementation. As part of your reconciliation loop you want to see if count(pods you own) === spec.replicas. If the number is less, you want to create that many pods. How do you count the number of pods in the cluster?

Ideally, you check your informer list, podInformer.list(), and count the ones you own. This is fast because the informer is using efficient caching logic courtesy of the kubernetes javascript client maintainers, and we know the list is up-to-date within some watch-update threshold. However, it is currently impossible to know whether the cache's first list operation has occurred, so you must listNamespacedPods yourself, negating the efficiencies of the cache.
Worse, it is never known-safe to check the informer list - thus its impossible to trust the list unless you keep your own copy. 

**Fix proposals***
The simplest thing to do here would be to mark ListWatch<T>::start (and Informer<T>::start) as async and await this.doneHandler(null).  doneHandler is already an async function that completes only after the list operation
https://github.com/kubernetes-client/javascript/blob/f119da07e8286caf4210f46e497d803e9610ae9a/src/cache.ts#L28

This is largely backwards compatible too, since start has no return value now and would be treated the same, except that it may break some linters or processors that note un-awaited promises.

Now your reconciliation loop can simply await on the result of that promise before calling podInformer.list()

Another possibility would be to add a new event that indicates the cache has been seeded.

I think these changes are small and come with a lot of utility.

benbuzbee:
Resolve #320 
See issue for discussion on use case and alternative implementations

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -28,8 +28,8 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
         }
     }
 
-    public start() {
-        this.doneHandler(null);
+    public async start(): Promise<void> {
+        await this.doneHandler(null);
     }
 
     public on(verb: string, cb: ObjectCallback<T>) {
--- a/src/cache_test.ts
+++ b/src/cache_test.ts
@@ -656,6 +656,44 @@ describe('ListWatchCache', () => {
 
         expect(addedList.length).to.equal(1);
     });
+     it('should resolve start promise after seeding the cache', async () => {
+        const fakeWatch = mock.mock(Watch);
+        const list: V1Namespace[] = [
+            {
+                metadata: {
+                    name: 'name1',
+                } as V1ObjectMeta,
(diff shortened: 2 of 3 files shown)


brendandburns:
/lgtm
/approve

(but has merge conflicts that need to be resolved)

Thanks!

benbuzbee:
@brendandburns conflicts with my previous change resolved and re-pushed.

Thank you

PR #326, opened 2019-08-15
Update compiled code with openapi-generator v4.1.0

silasbw:
v4.1.0 generates code the allows asynchronous authentication methods.

Using kubernetes-client/gen@795993978f8563a9c09089993be4d2b97d492eec

```
USERNAME=kubernetes ../gen/openapi/typescript.sh src/gen settings
```

cc https://github.com/kubernetes-client/javascript/issues/294


settings
```diff
@@ -23,3 +23,5 @@ export CLIENT_VERSION="0.8-SNAPSHOT"
 
 # Name of the release package
 export PACKAGE_NAME="@kubernetes/node-client"
+
+export OPENAPI_GENERATOR_COMMIT=v4.1.0
```
drubin: Should we be using `export KUBERNETES_BRANCH="release-1.13"` now that 1.15 is out

silasbw: Ah, yeah, I think so.  I'll update and regenerate.

PR #333, opened 2019-08-22
Add a refreshing auth token for in-cluster configurations.

### LINKED ISSUE #317: client should periodically reload InClusterConfig token
See https://github.com/kubernetes/kubernetes/issues/70679

brendandburns:
Closes #317

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -12,6 +12,7 @@ import { Authenticator } from './auth';
 import { CloudAuth } from './cloud_auth';
 import { Cluster, Context, newClusters, newContexts, newUsers, User } from './config_types';
 import { ExecAuth } from './exec_auth';
+import { FileAuth } from './file_auth';
 import { OpenIDConnectAuth } from './oidc_auth';
 
 // fs.existsSync was removed in node 10
@@ -28,6 +29,7 @@ export class KubeConfig {
     private static authenticators: Authenticator[] = [
         new CloudAuth(),
         new ExecAuth(),
+        new FileAuth(),
         new OpenIDConnectAuth(),
     ];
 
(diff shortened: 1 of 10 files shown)


src/file_auth.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export class FileAuth implements Authenticator {
+    private token: string | null = null;
+    private lastRead: Date | null = null;
+
+    public isAuthProvider(user: User): boolean {
+        return user.authProvider && user.authProvider.config && user.authProvider.config.tokenFile;
+    }
+
+    public getToken(user: User): string | null {
```
silasbw: I noticed https://github.com/kubernetes-client/javascript/blob/master/src/auth.ts#L8.  Is it worth rolling this `getToken` implementation into `applyAuthentication` now?

src/file_auth.ts
```diff
@@ -0,0 +1,48 @@
...
+            this.readToken(user.authProvider.config.tokenFile);
+        }
+        return this.token;
+    }
+
+    public async applyAuthentication(user: User, opts: request.Options | https.RequestOptions) {
+        // nothing to do.
+    }
+
+    private readToken(filePath: string) {
```
silasbw: renaming nit: `readToken` to `updateToken`

I'd expect `readToken` to return something read.  `updateToken` implies updating some state.

drubin: `refreshToken` feels like a better name for this.

src/config.ts
```diff
@@ -190,7 +192,12 @@ export class KubeConfig {
         this.users = [
             {
                 name: userName,
-                token: fs.readFileSync(`${pathPrefix}${Config.SERVICEACCOUNT_TOKEN_PATH}`).toString(),
```
drubin: @brendanburns What happens in the case that still is still a valid form of authentication?

Does it feel like we are removing this functionality of being able to raw pass just a token we should be updating more tests/code to clean up the usage of `user.token`

PR #334, opened 2019-08-22
add typedoc generation for corev1Api

kgrubb:
I just started playing with this library and I noticed most of the documentation is jsdoc in the codebase. I thought it may be easier for people to read and search via some typedocs.

Here's some examples of the output from running `npm run docs`:

_main page:_
Screen Shot 2019-08-22 at 5 02 47 PM

_search example:_
Screen Shot 2019-08-22 at 5 03 09 PM

_example docs for listnamespacedPod:_
Screen Shot 2019-08-22 at 5 03 20 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "generate": "./generate-client.sh",
     "watch": "tsc --watch",
     "test": "nyc mocha",
-    "prepare": "npm run build"
+    "prepare": "npm run build",
+    "docs": "typedoc src/gen/api/coreV1Api.ts"
   },
   "nyc": {
     "include": [
@@ -85,6 +86,7 @@
     "ts-mockito": "^2.3.1",
     "ts-node": "^8.2.0",
     "tslint": "^5.17.0",
+    "typedoc": "^0.15.0",
     "typescript": "^3.1.3"
   },
   "bugs": {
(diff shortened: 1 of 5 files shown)


kgrubb:
i signed it

kgrubb:
I signed it

kgrubb:
I signed it

kgrubb:
I'm not super sure what needs to be done to get the CLA check working. I've signed it already. Is there an alternative way to show that I've signed?

/assign @mbohlool

package.json
```diff
@@ -21,7 +21,8 @@
     "generate": "./generate-client.sh",
     "watch": "tsc --watch",
     "test": "nyc mocha",
-    "prepare": "npm run build"
+    "prepare": "npm run build",
+    "docs": "typedoc src/gen/api/coreV1Api.ts"
```
brendandburns: Can you extend this to add docs for other/all parts of the library? Does typedoc support a wild-card?

brendandburns:
Thanks for the PR! If you're open to it, I'd love to see docs generated for all of the APIs not just v1Core...

Thanks!

package.json
kgrubb: I've swapped it to `src/gen/api`. Are there any other directories that would be useful to have typedocs generated?

PR #346, opened 2019-09-25
Fix packaging.

brendandburns:
We changed where we generate the code, but we didn't update package.json to include it in the npm package...

cc @drubin @silasbw

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,10 @@
   "files": [
     "dist/*.ts",
     "dist/*.js",
+    "dist/gen/*.ts",
+    "dist/gen/*.js",
+    "dist/gen/**/*.ts",
+    "dist/gen/**/*.js",
     "README.md"
   ],
   "main": "dist/index.js",


package.json
```diff
@@ -9,6 +9,12 @@
   "files": [
     "dist/*.ts",
     "dist/*.js",
+    "dist/gen/*.ts",
```
drubin: https://docs.npmjs.com/files/package.json#files

Do you think we could do 

`"dist/**/*.js"`
`"dist/**/*.ts"`

Wondering if this would be better then we don't need to remember as the folders change?

silasbw: Would that pull in the examples?  e.g., `dist/exec/`

Use the below instead?
```
"dist/gen/**.js",
"dist/gen/**.ts"
```

drubin:
Does that also do directory globbing as far as I know it only does current directory.

silasbw:
oops, you're right.  I meant:

```
"dist/gen/**/*.js",
"dist/gen/**/*.ts"
```

package.json
drubin: @silasbw Why would you want to explicitly only include gen?

Is there something different between what I suggested above here? 

```
"dist/**/*.js"
"dist/**/*.ts"
```

brendandburns:
Added the glob, please take another look.

Thanks!

drubin:
/lgtm

I still think it would be nicer to do

"dist/**/*.js"
"dist/**/*.ts"

But don't think it's a blocker and this fixes the actuall issue.

PR #351, opened 2019-10-04
Use newer openAPI generator

### LINKED ISSUE #336: Reject with error instead of object
It would be nice if the promises rejected with an error and not with an object. Currently when I try to retrieve an object that is not present a 404 is thrown as an object which doesn't implement `Error`

https://github.com/kubernetes-client/javascript/blob/master/src/gen/api/customObjectsApi.ts#L151

aeb-sia:
Fixes #336

### DIFF
--- a/settings
+++ b/settings
@@ -30,4 +30,4 @@ export CLIENT_VERSION="0.8-SNAPSHOT"
 # Name of the release package
 export PACKAGE_NAME="@kubernetes/node-client"
 
-export OPENAPI_GENERATOR_COMMIT=v4.1.0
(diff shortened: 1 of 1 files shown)


aeb-sia:
hmm.. Do I really need to sign up just for this version change?

brendandburns:
@aeb-sia Unfortunately yes, you need to sign the CLA for all changes. It's not really a "sign-up" it is accepting the code agreement so we know you are donating the code.

If you'd rather not sign it, I can replicate this change and send a PR instead. Let me know.

aeb-sia:
To comply I'd have to register my employer first, so I guess it's easier if you go ahead.

PR #238, opened 2019-04-25
fixed bug on windows running cmd

mentaman:
Hey I got this error on windows:

`Error: Failed to refresh token: '%localappdata%\Google\Cloud' is not recognized as an internal or external command`

Where the real path is actually:
"%localappdata\Google\Cloud SDK\cloud_env.bat"

but since it had space in the path, it didn't parse it correctly.
So I wrapped the path in quotes and it fixed it.

I checked it only on windows and it works great now.
I recommend checking it still works the same on other platforms.

PR #352, opened 2019-10-10
Use command as quoted string in CloudAuth

jnummelin:
We faced a case where `gcloud` for some reason was installed in path that contains whitespaces. This results in slightly weird (but still perfectly working[1]) kubeconfig:
```
cmd-path: /Users/auser/exec -l /bin/bash/google-cloud-sdk/bin/gcloud
```

Now as the `CloudAuth` uses plain `child_process` for exec it naturally breaks in this case.

It used to work with 0.10.2 version, but seems https://github.com/kubernetes-client/javascript/commit/0055a50e954aef7013a30f2cc0617c6fab609204 broke it when `CloudAuth` was changed to use plain `child_process`.

I'd appreciate some pointers and ideas how to make the test case actually portable and without any need to manually first create symlinks. Would it make sense to dynamically create a symlink to `echo` in some temp dir with whitespaces in the name?

[1] `kubectl` works perfectly still, so did previous version of JS client as it used `execa`

### DIFF
--- a/src/cloud_auth.ts
+++ b/src/cloud_auth.ts
@@ -63,6 +63,8 @@ export class CloudAuth implements Authenticator {
         if (!cmd) {
             throw new Error('Token is expired!');
         }
+        // Wrap cmd in quotes to make it cope with spaces in path
+        cmd = `"${cmd}"`;
         const args = config['cmd-args'];
         if (args) {
             cmd = cmd + ' ' + args;
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -792,14 +792,6 @@ describe('KubeConfig', () => {
             }
         });
         it('should exec succesfully with spaces in cmd', async () => {
-            /**
-             *
-             * to test this, symlink echo to dir that has spaces:
-             * mkdir -p /tmp/foo\ bar/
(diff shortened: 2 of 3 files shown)


drubin:
Thanks for this! 

I remember dealing with something similar with windows pathing previously too https://github.com/kubernetes-client/javascript/pull/238/ it seems like this is a constant problem of back and forth. So your idea of adding a test case for this I believe would solve it hopefully once and for all.

>I'd appreciate some pointers and ideas how to make the test case actually portable 

We support windows so doing a `symlink` during the tests might not be the best way to handle this. 

But what about potentially trying one of these 2 options(just ideas). 

1. Create a nodejs file with a space in it that echo's the arguments back to the user and trying to execute that as part of the test? for example `test/echo space.js`  and trying to use that?
1. Add a test case that validates that the cmd is always quoted before sending it to the function that executes the command.

How does that sound?

brendandburns:
We use `mockfs` in a number of places. I'm not sure if spaces work in `mockfs` correctly though.

jnummelin:
@drubin Thanks for the ideas. I implemented the test case with option 1, with a small js script named `echo space.js`. :)

brendandburns:
/lgtm
/approve

Thanks for the PR!

Issue #287, opened 2019-06-21
patchNode doesn't use the correct headers by default


aackerman:
By default patchNode should supply the `application/merge-patch+json` header.

I had to call the API as follows to get a patch node call to work

```js
let patchBody = {
  apiVersion: 'v1',
  kind: 'Node',
  metadata: { name: nodeName },
  spec: { unschedulable: true }
}
return api.patchNode(nodeName, patchBody, undefined, undefined, {
  headers: {
    'Content-Type': 'application/merge-patch+json'
  }
})
```

brendandburns:
It appears that the generated typescript code never tries to set a custom content type, so as far as the generated code is concerned this is the right way to do this.

One could argue that it should do something different (though which header is it supposed to chose, since there are three potential content types for patches?)

But either way, this is a bug in the code generator (if it is a bug at all) not this library, so closing.

OJezu:
If this is a bug in code generator not in the library, was this taken to the code generator bug tracker?

There is also hardly any documentation how to use the patch functions.

woodcockjosh:
Thanks @aackerman for posting the workaround. Anyway if its a code generator problem can someone please link me the issue to the bug in the code generator?

brendandburns:
I don't think it's actually a code generator bug. There are three different possible content-type headers to chose from, how is the code generator supposed to know which one to send. Having the user explicitly select the header seems like the right approach.

woodcockjosh:
@brendandburns this defeats the purpose of the sdk. While it may be difficult for the code generator to know with the current architecture. The end user shouldn't have to know this. If I use kubectl I don't have to know the content type of the request.

woodcockjosh:
I could be mistaken but it appears all PATCH requests have the same request header content type in v1.16. Can you not make the generator look at the http method to set the correct content type?

PR #356, opened 2019-10-14
fix patchNamespacedDeployment

filipecaixeta:
patchNamespacedDeployment fails with '415: Unsupported Media Type' when Content-Type is 'application/json'.
Updated the default Content-Type to application/strategic-merge-patch+json

### DIFF
--- a/src/gen/api/appsV1Api.ts
+++ b/src/gen/api/appsV1Api.ts
@@ -2878,7 +2878,7 @@ export class AppsV1Api {
             .replace('{' + 'name' + '}', encodeURIComponent(String(name)))
             .replace('{' + 'namespace' + '}', encodeURIComponent(String(namespace)));
         let localVarQueryParameters: any = {};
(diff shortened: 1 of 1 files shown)


filipecaixeta:
I signed it

brendandburns:
@filipecaixeta unfortunately, this code is generated, so we can't patch things this way, since it will get overwritten the next time we regenerate.

Also, there are actually multiple different potential headers you could send.

The best workaround for this is:

```javascript
let patchBody = {
  apiVersion: 'v1',
  kind: 'Node',
  metadata: { name: nodeName },
  spec: { unschedulable: true }
}
return api.patchNode(nodeName, patchBody, undefined, undefined, {
  headers: {
    'Content-Type': 'application/merge-patch+json'
  }
})
```

Please see https://github.com/kubernetes-client/javascript/issues/287 for more details.

brendandburns:
I'm going to close this since there's nothing else to do here, please use `/reopen` if you want to reopen the issue for any reason.

PR #357, opened 2019-10-16
Fix OIDC refresh token storage with user auth config

jnummelin:
Previously the `refresh-token` was not persisted on the auth config at all. This rendered the whole OIDC auth pretty much broken, it could only work for a single refresh, after that the refresh token would have been used and on the next try the client would try to use already used token --> errors.

This PR makes the refresh token to be properly stored on the oidc auth config along with the real token on token refresh.


brendandburns:
/approve

Thanks for the PR!

PR #368, opened 2019-11-15
Fix bugs in programmatic cluster config. Add unit test.

### LINKED ISSUE #365: Calling addCluster/User/Context fails if nothing was previously loaded
**Failing Test Code**
```typescript
kc.addCluster({
  name: 'test',
  server: `https://localhost:9889`,
  skipTLSVerify: true,
})
///TypeError: Cannot read property 'forEach' of undefined
///        at KubeConfig.addCluster (/Users/test/project/node_modules/@kubernetes/client-node/dist/config.js:176:23)
kc.addUser({
  token: 'test',
  username: 'test',
  name: 'test',
})
kc.addContext({
  cluster: 'test',
  name: 'test',
  user: 'test',
})
kc.setCurrentContext('test')

```
**Code Referenced By StackTrace**
```javascript
// ...
    addCluster(cluster) {
        this.clusters.forEach((c, ix) => {
            if (c.name === cluster.name) {
                throw new Error(`Duplicate cluster: ${c.name}`);
            }
        });
        this.clusters.push(cluster);
    }
    addUser(user) {
        this.users.forEach((c, ix) => {
            if (c.name === user.name) {
                throw new Error(`Duplicate user: ${c.name}`);
            }
        });
        this.users.push(user);
    }
    addContext(ctx) {
        this.contexts.forEach((c, ix) => {
            if (c.name === ctx.name) {
                throw new Error(`Duplicate context: ${c.name}`);
            }
        });
        this.contexts.push(ctx);
    }
// ...
```

It looks like it's as simple as `this.clusters`, `this.users`, `this.contexts` not being defined when the `KubeConfig` is created. The fix would be to just define them when the object is created. 

I would create a PR, but I've been fighting the CLA for a while now, and never got it figured out.

brendandburns:
Fixes #365

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -53,6 +53,12 @@ export class KubeConfig {
      */
     public 'currentContext': string;
 
+    constructor() {
+        this.contexts = [];
+        this.clusters = [];
+        this.users = [];
+    }
+
     public getContexts() {
         return this.contexts;
     }
@@ -224,6 +230,9 @@ export class KubeConfig {
     }
 
     public addCluster(cluster: Cluster) {
+        if (!this.clusters) {
+            this.clusters = [];
+        }
         this.clusters.forEach((c: Cluster, ix: number) => {
             if (c.name === cluster.name) {
                 throw new Error(`Duplicate cluster: ${c.name}`);
@@ -233,6 +242,9 @@ export class KubeConfig {
     }
 
     public addUser(user: User) {
+        if (!this.users) {
+            this.users = [];
+        }
         this.users.forEach((c: User, ix: number) => {
             if (c.name === user.name) {
                 throw new Error(`Duplicate user: ${c.name}`);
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -233,6 +236,9 @@ export class KubeConfig {
     }
 
     public addUser(user: User) {
+        if (!this.users) {
```
drubin: Would it not be better to put this in the constructor? I think it would be more logical and easier to follow too.

drubin: Not a blocker just curious before I add my lgtm (and it merges)

brendandburns: The fields involved are public, so it's always possible that someone will overwrite them with something undefined or null, so the check here is safer.

brendandburns:
Comment addressed, please re-check.

Thanks!

drubin:
/lgtm
Thanks for the clarification

PR #369, opened 2019-11-15
Add some missing error handling.

### LINKED ISSUE #364: makeInformer watch stops receiving updates for "/api/v1/configmaps"
```
  kc.loadFromDefault();

  const listFn = () => k8sApi.listConfigMapForAllNamespaces();
  const informer = k8s.makeInformer(kc, "/api/v1/configmaps", listFn);

  informer.on("add", (obj: any) => {
    console.log(`Added: ${obj.metadata!.name}`);
  });
  informer.on("update", (obj: any) => {
    console.log(`Updated: ${obj.metadata!.name}`);
  });
  informer.on("delete", (obj: any) => {
    console.log(`Deleted: ${obj.metadata!.name}`);
  });

  await informer.start();
```
Code above works as expected for about the first 10-15 mins, but stops receiving updates for config maps after that.

brendandburns:
May help fix #364 

cc @mag415

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -18,6 +18,7 @@ export class DefaultRequest implements RequestInterface {
 }
 
 export class Watch {
+    public static SERVER_SIDE_CLOSE = { error: 'Connection closed on server' };
     public config: KubeConfig;
     private readonly requestImpl: RequestInterface;
 
@@ -64,6 +65,10 @@ export class Watch {
                 // ignore parse errors
             }
         });
+        stream.on('error', (err) => {
+            done(err);
+        });
+        stream.on('close', () => done(null));
         const req = this.requestImpl.webRequest(requestOptions, (error, response, body) => {
             if (error) {
                 done(error);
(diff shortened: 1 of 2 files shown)


src/watch.ts
```diff
@@ -64,6 +64,12 @@ export class Watch {
                 // ignore parse errors
             }
         });
+        stream.on('error', (err) => {
+            done(err);
+        });
+        stream.on('close', () => {
+            done(null);
```
drubin: is this related to it closing on the serverside? if so I think we should be passing a custom error object to the `done()` function

brendandburns: This would be a server-side close, but I think that's an expected case? I don't think we want to return an error there.

brendandburns:
Unit tests added (and they fail w/o the code changes) so this is definitely a bug.

@drubin please take another look. Thanks!

drubin:
/lgtm

Thanks for adding the tests!

Issue #340, opened 2019-09-13
Anything similar to Golang clientcmd ModifyConfig?


ZedYeung:
It looks like currently, we could only read the config file, is there any method like Golang clientcmd ModifyConfig so that we could modify the config file?

brendandburns:
Not currently. Can you describe your use case in more detail? Do you want to modify and then save to file a kubeconfig?

ZedYeung:
yes, ideally, i hope it could be same as  Golang clientcmd ModifyConfig.

brendandburns:
For future reference, code is here:
https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/config.go#L158

Hrm. Why not just modify the config object in Javascript and then export to YAML?

That function seems super twisty and it's not exactly clear to me why its useful.

Can you describe your use case in more detail?

Thanks

ZedYeung:
That is exactly what it did. It is very true that function is super twisty with some rules. I  just want to keep consistent with that official rule to modify the config file, in case some specific rule would be broken. But looks like that is not easy. What I do now is just modifying only one file in the order `env KUBECONFIG and then ~/.kube/config`. Hope that is good.

brendandburns:
That seems reasonable for now. I'll leave this issue open, but I don't expect we'll implement this in the near term.

machine424:
If I really understand how tokens refreshing is implemented, new fetched tokens (at least for oidc) are not stored in `~/.kube/config`, that means, if a refresh occurs, `~/.kube/config` becomes obsolete. Also if other clients are to be run after, they will get old and unusable tokens from `~/.kube/config`. The current refreshing process is only useful for a single long-running client ? I see that you've started working on configuration exporting (https://github.com/kubernetes-client/javascript/pull/372) Will tokens persistence be implemented soon ? Should it be done by the app using the client ?

brendandburns:
Token persistence is tricky, because it is potentially racy between multiple different clients running on the same machine.

I'd prefer that people handle it themselves in their application code rather than this library accidentally corrupting someone's kubeconfig.

PR #372, opened 2019-11-20
Add an export function for KubeConfigs. Add unit tests too.

brendandburns:
Partial fix for #340

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -10,7 +10,17 @@ import shelljs = require('shelljs');
 import * as api from './api';
 import { Authenticator } from './auth';
 import { CloudAuth } from './cloud_auth';
-import { Cluster, Context, newClusters, newContexts, newUsers, User } from './config_types';
+import {
+    Cluster,
+    Context,
+    exportCluster,
+    exportContext,
+    exportUser,
+    newClusters,
+    newContexts,
+    newUsers,
+    User,
+} from './config_types';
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -320,6 +330,20 @@ export class KubeConfig {
...
+            apiVersion: 'v1',
+            kind: 'Config',
+            clusters: this.clusters.map(exportCluster),
+            users: this.users.map(exportUser),
+            contexts: this.contexts.map(exportContext),
+            preferences: {},
+            'current-context': this.getCurrentContext(),
+        };
+
+        return JSON.stringify(configObj);
```
drubin: Is there a reason we are exporting the data as json and not an object? Just wondering if people would want the json or maybe rather use yaml?

I didn't look at the go implementation though

brendandburns: JSON is valid YAML :)

The purpose of this is to serialize it to a string. I can change the name to 'SerializeConfig' if that's clearer.

drubin: Ah yes, I totally forgot. 

I was just thinking most people would use this to write their kubeconfig which is currently in `yaml` and if their intention was to convert it to `yaml` to be human-readable they would need to parse this and then convert it anyway.

drubin:
/lgtm

We can address the other issues or name in another PR?

PR #376, opened 2019-12-07
Update formatting.

brendandburns:


drubin:
These are great but don't we have ci for prettier formatting? (just curious if these are going to get overridden by some script)

brendandburns:
Something is weird here. This was the output from `npm run format` on my machine, but clearly it is making the CI unhappy.  I think that I was using a different version of prettier...

I'll see if updating package.json fixes things.

drubin:
Let me know if I can help

PR #381, opened 2019-12-21
Add ipv6 support.

### LINKED ISSUE #380: in-cluster config builder does not work properly in an ipv6 cluster
https://github.com/kubernetes-client/javascript/blob/d262aff9acc43efda602142a587baa0183c4970a/src/config.ts#L204

assumes a DNS name or ipv4 address... it does not work correctly with an ipv6 address (the ipv6 host must be wrapped in brackets)

brendandburns:
Fixes #380 

cc @liggett @thockin

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,7 @@
 import execa = require('execa');
 import fs = require('fs');
 import https = require('https');
+import net = require('net');
 import path = require('path');
 
 import yaml = require('js-yaml');
@@ -197,11 +198,17 @@ export class KubeConfig {
             scheme = 'http';
         }
 
+        // Wrap raw IPv6 addresses in brackets.
+        let serverHost = host;
+        if (host && net.isIPv6(host)) {
+            serverHost = `[${host}]`;
+        }
+
         this.clusters = [
             {
                 name: clusterName,
                 caFile: `${pathPrefix}${Config.SERVICEACCOUNT_CA_PATH}`,
-                server: `${scheme}://${host}:${port}`,
+                server: `${scheme}://${serverHost}:${port}`,
                 skipTLSVerify: false,
             },
         ];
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -197,11 +197,17 @@ export class KubeConfig {
             scheme = 'http';
         }
 
+        // Wrap raw IPv6 addresses in brackets.
+        let serverHost = host;
+        if (host && host.indexOf(':') != -1) {
```
drubin: There are methods to check if its an ipv6 Adress.

src/config.ts
```diff
@@ -197,11 +197,17 @@ export class KubeConfig {
...
+            serverHost = `[${host}]`;
+        }
+
         this.clusters = [
             {
                 name: clusterName,
                 caFile: `${pathPrefix}${Config.SERVICEACCOUNT_CA_PATH}`,
-                server: `${scheme}://${host}:${port}`,
+                server: `${scheme}://${serverHost}:${port}`,
```
drubin: Maybe we should also be building up the address with Uri helpers? Maybe they can also handle this natively?

src/config_test.ts
```diff
@@ -1073,6 +1073,32 @@ describe('KubeConfig', () => {
...
+            mockfs.restore();
+            delete process.env.KUBERNETES_SERVICE_HOST;
+            delete process.env.KUBERNETES_SERVICE_PORT;
+
+            const cluster = kc.getCurrentCluster();
+            expect(cluster).to.not.be.null;
+            if (!cluster) {
+                return;
+            }
+            expect(cluster.server).to.equal('http://[::1234:5678]:80');
```
drubin: Thanks for adding a test!

src/config.ts
brendandburns: done.

src/config.ts
brendandburns: I couldn't tell from docs if they handled this or not, so I left it as is.

brendandburns:
Comments addressed, style fixed, please take another look.

Thanks!

PR #388, opened 2020-01-08
Improve the OIDC auth to pull the expiration date from the token.

### LINKED ISSUE #386: 401 Unauthorized with oidc user.
Hi,

I'm getting a consistent 403 forbidden. This doesn't happen with the Python library or kubectl.  The user is an oidc user, backed by Auth0.  I have ensured that the token is valid. 

I can't see any particular place where I need to pass in or set a flag. The cluster is a kops-generated cluster with its own generated certificate.

Looking at https://github.com/kubernetes-client/javascript/blob/master/src/oidc_auth.ts#L43 , I see that I have a flag in my config file `extra-scopes: openid profile`, which doesn't appear to be an option in the dict there.  But this doesn't seem to matter in the Python library - https://github.com/kubernetes-client/python-base/blob/6b6546131217a2a9fdcf431a286c346619d2923a/config/kube_config.py#L289 

First, the package.json:
```
    "dependencies": {
        "@kubernetes/client-node": "^0.11.0",
[snip]
        "kubernetes-client": "^0.11.0",
```

And for an example function:

```
function Example() {
    console.log(kubeconfig.currentContext); // this is correctly returned
    console.log("User", kubeconfig.getCurrentUser());   // this is the correct user object.
    console.log("Context", kubeconfig.getContextObject('Redacted'));  // This corresponds to the ~/.kube/config file.

    k8sApi.listNamespacedPod('default')
        .then((res:any) => {
            console.log(res.body.response);
        })
        .catch((error: any) => {
            // Yet, here we have 401.
            console.log("Error: ", error.response.body);
        });
}
```

I'm somewhat stumped here, I'm afraid.

brendandburns:
Fixes #386 (I hope)

See the relevant info from the golang OIDC implementation here:

https://github.com/kubernetes-client/javascript/issues/386#issuecomment-571883404

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1167,7 +1167,7 @@
     "handlebars": {
       "version": "4.5.3",
       "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.5.3.tgz",
-      "integrity": "sha512-3yPecJoJHK/4c6aZhSvxOyG4vJKDshV36VHp0iVCDVh7o9w2vwi3NSnL2MMPj3YdduqaBcu7cGbggJQM0br9xA==", 
+      "integrity": "sha512-3yPecJoJHK/4c6aZhSvxOyG4vJKDshV36VHp0iVCDVh7o9w2vwi3NSnL2MMPj3YdduqaBcu7cGbggJQM0br9xA==",
       "dev": true,
       "requires": {
         "neo-async": "^2.6.0",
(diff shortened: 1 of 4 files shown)


brendandburns:
Ok, this fix was validated (and I improved the testing too...)

This is ready to merge, I think.

@drubin any chance you can take a look?

PR #391, opened 2020-01-11
Remove some unnecessary validation.

### LINKED ISSUE #390: Don't fail on missing kubeconfig version / kind
These fields are "legacy", optional, and to be removed from downstream per the source types in client-go, as far as i know they can only ever be "Config" and "v1". Creating a client should not fail without these omitempty fields being present.

https://github.com/kubernetes/client-go/blob/940f07561929308fc33fc58bfa185f0544b84846/tools/clientcmd/api/types.go#L32-L41

originally reported in https://github.com/kubernetes-sigs/kind/issues/1242

brendandburns:
Fixes #390 

cc @bentheelder

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -157,9 +157,6 @@ export class KubeConfig {
 
     public loadFromString(config: string) {
         const obj = yaml.safeLoad(config) as any;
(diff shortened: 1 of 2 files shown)


BenTheElder:
Thanks!

PR #393, opened 2020-01-15
Add error handling to informer

kellycampbell:
Adds a new callback type of `error` for informers so that errors received by the cache.doneHandler, e.g. `Unauthorized` or `Not Found` are surfaced and not lost.


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #396, opened 2020-01-23
Fix watch to emit an error twice on error. Add tests.

### LINKED ISSUE #392: watch with resourceVersion too old triggers done() callback with null err
It appears that when the resourceVersion is no longer available in the cluster we probably are receiving a `410 GONE` response. However I'm observing that the `done()` callback has a `null` value for the error.

Any way we can more sanely handle this scenario?

Thanks!

brendandburns:
Fixes #392

### DIFF
--- a/src/watch_test.ts
+++ b/src/watch_test.ts
@@ -130,6 +130,7 @@ describe('Watch', () => {
             pipe: (stream) => {
                 stream.write(JSON.stringify(obj1) + '\n');
                 stream.emit('error', errIn);
+                stream.emit('close');
             },
         };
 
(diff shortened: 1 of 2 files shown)


src/watch_test.ts
```diff
@@ -168,7 +169,11 @@ describe('Watch', () => {
         expect(receivedObjects).to.deep.equal([obj1.object]);
 
         expect(doneCalled).to.equal(true);
-        expect(doneErr).to.deep.equal(errIn);
+        expect(doneErr.length).to.equal(2);
+        if (doneErr.length == 2) {
```
drubin: This `if check` is redundant as the first check already guards against it.

drubin:
@brendandburns the linting is off but it looks good.

src/watch_test.ts
brendandburns: done.

brendandburns:
lint fixed, comment addressed. Please re-check.

Thanks!

PR #400, opened 2020-01-27
Add token refresh info and config file rewriting.

### LINKED ISSUE #399: Kubernetes config not updated when token refreshed
I'm using OIDC authentication and running in the the following issue:

```
OpenIdConnectError: invalid_request (Refresh token is invalid or has already been claimed by another client.)
    at Client.requestErrorHandler (/Users/inikolaev/Projects/kube/node_modules/openid-client/lib/helpers/error_handler.js:16:11)
    at processTicksAndRejections (internal/process/task_queues.js:85:5) {
  error: 'invalid_request',
  error_description: 'Refresh token is invalid or has already been claimed by another client.'
}
``` 

After debugging it looks like token is being refreshed correctly the first time, but Kubernetes configuration is not being automatically updated and next time I start application I load old configuration and try to refresh my token using invalid refresh token.

I'm loading configuration from `~/.kube/config` and I was expecting it to be updated automatically. I'm not sure is this is how it's supposed to work and if so do you have any pointers on how to update that configuration properly?

brendandburns:
Should fix #399

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -64,10 +64,13 @@ export class KubeConfig {
      */
     public 'currentContext': string;
 
+    private configFile: string;
+
     constructor() {
         this.contexts = [];
         this.clusters = [];
         this.users = [];
+        this.configFile = '';
     }
 
     public getContexts() {
@@ -125,6 +128,7 @@ export class KubeConfig {
         const rootDirectory = path.dirname(file);
         this.loadFromString(fs.readFileSync(file, 'utf8'));
         this.makePathsAbsolute(rootDirectory);
+        this.configFile = file;
     }
 
     public async applytoHTTPSOptions(opts: https.RequestOptions) {
@@ -404,7 +408,11 @@ export class KubeConfig {
             opts.headers = [];
         }
         if (authenticator) {
-            await authenticator.applyAuthentication(user, opts);
+            const refresh = await authenticator.applyAuthentication(user, opts);
+            if (refresh && this.configFile.length > 0) {
+                const contents = this.exportConfig();
+                await fs.writeFile(this.configFile, contents, () => {});
+            }
         }
 
         if (user.token) {
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -3,7 +3,12 @@ import request = require('request');
 
 import { User } from './config_types';
 
+export interface Token {
+    token: string | null;
+    refreshed: boolean;
+}
+
 export interface Authenticator {
     isAuthProvider(user: User): boolean;
-    applyAuthentication(user: User, opts: request.Options | https.RequestOptions): Promise<void>;
+    applyAuthentication(user: User, opts: request.Options | https.RequestOptions): Promise<boolean>;
 }
--- a/src/cloud_auth.ts
+++ b/src/cloud_auth.ts
@@ -3,7 +3,7 @@ import https = require('https');
 import * as jsonpath from 'jsonpath-plus';
 import request = require('request');
 
-import { Authenticator } from './auth';
+import { Authenticator, Token } from './auth';
 import { User } from './config_types';
 
 /* FIXME: maybe we can extend the User and User.authProvider type to have a proper type.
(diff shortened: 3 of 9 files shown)


src/config.ts
```diff
@@ -404,7 +408,11 @@ export class KubeConfig {
             opts.headers = [];
         }
         if (authenticator) {
-            await authenticator.applyAuthentication(user, opts);
+            const refresh = await authenticator.applyAuthentication(user, opts);
+            if (refresh && this.configFile.length > 0) {
+                const contents = this.exportConfig();
+                await fs.writeFile(this.configFile, contents, () => {});
```
inikolaev: I have one concern about this - if config file was changed on disk, e.g. by using `kubectl` in another context, then this will discard any changes made there and can lead to the similar problem with stale refresh token. Since we know what has been changed, you could patch config file on disk, instead of exporting current config.

brendandburns: This race is that this is why there is a `tokenFile` directive in the kubeconfig file. (I'm realizing actually that this PR isn't correct because it doesn't properly handle `tokenFile`

Perhaps I will change this so that it only does the refresh when `tokenFile` is present.

(concerns like this are why this library doesn't have refresh currently :)

inikolaev: But in my case for example, I don't have `tokenFile`, but I still need token to be refreshed. 
Ok, maybe for now I will try to craft some custom solution which will track when config is changed by the client and merge those changes back into the kubeconfig file and see how that works.

brendandburns: The trouble is that the file always could have changed underneath you between when you check and when you write. There's no way to guarantee that it won't occur.

inikolaev: That is true. 

My idea was that I could maintain two copies of kubeconfig: the original one, which I loaded when application started, and a copy which is being updated by the library. Then I could periodically load kubeconfig from the filesystem, somehow make a diff with my original copy to see what has changed, do the same for the changes made by library. Then order changes by date, resolve conflicts and write back. 

The file needs to be locked during all this time to prevent any modification in-between, at least I think this is how it's done in Go client.

But yeah, that is not a trivial task :)

drubin:
@brendandburns any update on this?

src/config.ts
inikolaev: What if instead of writing to a file it would fire some event instead, which could be handled by some other component? I've been working on this kind of workaround for my own purposes:

```js
import * as fs from 'fs';
import { KubeConfig, dumpYaml, loadYaml, findHomeDir } from '@kubernetes/client-node';

export class KubeConfigSynchronizer {
    private config: KubeConfig;
    private path: string = `${findHomeDir()}/.kube/config`;

    constructor(config: KubeConfig) {
        this.patch(); // patch KubeConfig to install update listeners
        this.config = config;

        fs.watch(this.path, (event, filename) => {
            config.loadFromDefault();
            console.log('Kube config reloaded');
        });
    }

    update() {
        const data = dumpYaml(loadYaml(this.config.exportConfig()));
        fs.writeFile(this.path, data, {}, () => {
            console.log("Kube config updated");
        });        
    }

    private patch() {
        const MyKubeConfig = KubeConfig as any
        MyKubeConfig.authenticators.forEach((authenticator: any) => {
            const applyAuthentication = authenticator.applyAuthentication;
            const newApplyAuthentication = (user: any, opts: any) => {
                return applyAuthentication.call(authenticator, user, opts).then((result: any) => {
                    this.update();
                    return result;
                }).catch((e: any) => {
                    throw e;
                });
            };
            authenticator.applyAuthentication = newApplyAuthentication.bind(authenticator);
        });
    }
}
```

It's not ideal right now, because sometimes `loadFromDefault()` throws an exception which I haven't looked into yet, but it would make the whole thing a bit easier if I could subscribe for kubeconfig update events, instead of patching authenticators.

brendandburns:
@drubin this stalled out b/c of the race conditions discussed above. I'll close this for now and re-open when I have a good plan.

PR #408, opened 2020-02-02
Apply authentication when streaming logs

### LINKED ISSUE #405: Log class seems to be outdated
I'm trying to stream logs from the pod using `Log` class and I'm getting an unauthenticated error:

```json
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}
```

At the same time both `kubectl logs` and `readNamespacedPodLog` works.

I guess the problem is that `Log` uses

```js
this.config.applyToRequest(requestOptions);
```

while `readNamespacedPodLog` uses

```js
let authenticationPromise = Promise.resolve();
authenticationPromise = authenticationPromise.then(() => this.authentications.BearerToken.applyToRequest(localVarRequestOptions));
authenticationPromise = authenticationPromise.then(() => this.authentications.default.applyToRequest(localVarRequestOptions));
```

in order to apply proper credentials.

That makes me wonder if `Log` class still a thing and I can rely on it, or should I fallback on `readNamespacedPodLog`, though I'm not clear whether I can use `follow` with it and how.

EDIT: I guess I can `pipe` the response in a similar way it's done in the `Log` class:

```js
api.readNamespacedPodLog(pod, namespace, container, )
  .then((response) => {
    response.response.pipe(writable);
  });

inikolaev:
This fixes #405 which is caused by `Config.applyToRequest` being an asynchronous function and thus should be awaited on. Method `Log.log`, which calls this function, doesn't await for the authentication to be properly applied and causes API call sent without any credentials, which resulted in 401 response from the server.

This change converts `Log.log` method into `async` which returns `Promise<request.Request>` and awaits for `Config.applyToRequest` to complete before proceeding.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -401,7 +401,7 @@ export class KubeConfig {
         });
 
         if (!opts.headers) {
-            opts.headers = [];
+            opts.headers = {};
         }
         if (authenticator) {
             await authenticator.applyAuthentication(user, opts);
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -212,7 +212,7 @@ describe('KubeConfig', () => {
             kc.applytoHTTPSOptions(opts);
 
             expect(opts).to.deep.equal({
-                headers: [],
+                headers: {},
                 ca: new Buffer('CADATA2', 'utf-8'),
                 cert: new Buffer('USER2_CADATA', 'utf-8'),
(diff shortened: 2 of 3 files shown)


inikolaev:
I have signed CLA

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #410, opened 2020-02-05
Fixed missing variable initialization

runi95:
Since ws was never initialized on line 86 it defaulted to undefined which means that if(ws !== null) on line 100 becomes true


brendandburns:
This looks good to me but we need you to sign the CLA.

Thanks!

runi95:
The CLA has been signed now.

PR #415, opened 2020-02-16
Initial implementation of "top" command for nodes

brendandburns:
Not quite the same as in `kubectl` since we don't support the metrics API (yet)


brendandburns:
Friendly ping to @drubin if you have time to review. If not, no big deal.

Thanks!

drubin:
/lgtm

I thought I posted this when I looked at it a few days ago sorry...

The calculation logic seems a bit overly complex to be doing in the client but as you said we don't support the metrics Api yet.

PR #418, opened 2020-02-27
Update openapi generator v4.2.3

sheldonkwok:
I was interested in adding a keepalive agent to the kubernetes clients requests. 
This [PR](https://github.com/OpenAPITools/openapi-generator/pull/4633) adds an "interceptor" model to the request options and it's in openapi generator v4.2.3. 

In this PR, I have updated settings for this new openapi generator. I also changed the kubernetes-client/gen commit to one that uses node v10 since v8 is deprecated.

I ran the gen locally and it completed successfully but didn't include that since I wasn't sure if there were any other steps required.


brendandburns:
Please do run the gen locally and include that as a separate commit in this PR.

Thanks!

sheldonkwok:
Okay, pushed. I noticed the [kubernetes/api/openapi-spec/swagger.json ](https://github.com/kubernetes/kubernetes/tree/master/api/openapi-spec)is now `unversioned` so everything is tagged with the branch name instead of the actual version. Is that okay now?

brendandburns:
Do you have any idea why the tag changed? is it because of a change in the openapi generator?

I'd like to understand why that changed, and then we can merge this.

Thanks!

sheldonkwok:
Yeah in this commit the version was dropped from the openapi spec swagger.json
https://github.com/kubernetes/kubernetes/commit/146d503cfefe2f06ae4cb0b28fc6264c965f65b6#diff-9ce7ea8441086bf1902b4f936f4601d0

This seems to have been applied across multiple release branches. We were using branch `release-1.15`. In https://github.com/kubernetes-client/gen/blob/master/openapi/preprocess_spec.py#L379, unversioned specs end up taking on the branch/tag name so it became `release-1.15`. I just regenerated using the latest 1.15 tag (`1.15.10`). So now the file headers look more consistent. It seems like we should use the tag over the branch moving forwards.

Issue #349, opened 2019-09-27
V1ServicePort.targetPort type changed from any to object in 0.10.3


zen0wu:
An issue I noticed while upgrading.

It used to be `any`, which works fine. Now it's `object`, where giving a number would cause a type error.

https://github.com/kubernetes-client/javascript/blob/master/src/gen/model/v1ServicePort.ts#L37

brendandburns:
Thanks for pointing this out. I'm not sure where that came from... I will dig in and see what we need to do to fix it.

Note that you can always pass the number as a string (e.g. `"123"`) the code will do the right thing.

zenozen:
Thanks for responding. Passing as string doesn't pass the type check too, string doesn't conform to object. Probably need to with `any`.

xogeny:
I found a similar issue with `ExtensionsV1beta1IngressBackend`.  In that type definition, `servicePort` is listed as `object` but it should really be `number | string`.  You can see where the confusion may be coming from if you look at the `swagger.json` code found [here](https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json):

```json
...
    "io.k8s.api.networking.v1beta1.IngressBackend": {
      "description": "IngressBackend describes all endpoints for a given service and port.",
      "properties": {
        "serviceName": {
          "description": "Specifies the name of the referenced service.",
          "type": "string"
        },
        "servicePort": {
          "$ref": "#/definitions/io.k8s.apimachinery.pkg.util.intstr.IntOrString",
          "description": "Specifies the port of the referenced service."
        }
      },
      "required": [
        "serviceName",
        "servicePort"
      ],
      "type": "object"
    },
...
```

Specifically, note the ` "$ref": "#/definitions/io.k8s.apimachinery.pkg.util.intstr.IntOrString",`.  I suspect that what ever automatic process is used to generate the TypeScript definitions, it gets confused by the `$ref` here and doesn't translate it into the correct type.

cwardcode:
This is preventing me from being able to upgrade to 0.11.0 as well. Are there any updates as to when this will be resolved?

nicholascapo:
I am seeing the same problem with `V1RollingUpdateDaemonSet.maxUnavailable` [0]
Neither `string` nor `number` are assignable to `object`.

[...]

PR #425, opened 2020-04-06
fix: wrong targetPort type preventing service api calls

jhkimac:


drubin:
Hi, @jhkimac Thanks so much for this contribution! 

Sadly we can't accept this as is, as this code is automatically generated from the openApi spec. So we need to fix it upstream in the generator or in the openApi spec. 

See  https://github.com/kubernetes-client/javascript#re-generating-code which calls https://github.com/kubernetes-client/javascript/blob/master/generate-client.sh

How does this fix it compared to the `object` value?

/assign @drubin

jhkimac:
Hi @drubin, I will try it out after a take a bit of rest. In the meantime, I have also found similar behavior in V1Container's `V1Container.livenessProbe.httpGet.port`, type enforcing object instead of number. I cannot find this problem in the openApi spec (or I am just too sleepy) at the moment. Will update after tomorrow.

brendandburns:
This is discussed in #349 there's something weird going on with the Swagger and the code generator. It has to do with the way that IntOrString is represented in Swagger, and also what looks like it might be a bug in the code generator.

I've take a few attempts at fixing this, but it hasn't gone anywhere.

As @drubin says, we can't manually edit the code generated files.

brendandburns:
Closing this PR since we can't merge it. Please continue discussion as needed in #349

PR #433, opened 2020-04-20
add github action.

brendandburns:
cc @drubin


.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [ '10', '8' ]
```
drubin, line 14: Any chance we can add ['12]

```suggestion
        node: [ '12', '10', '8' ]
```

As this is the latest LTS versions.

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Node ${{ matrix.node }} validation
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      - run: npm install
```
drubin, line 21: Maybe we can add back 
https://github.com/kubernetes-client/javascript/blob/master/.travis.yml#L6-L9

```suggestion
      # Pre check to check that version matches package-lock.json
      # needs to happen before npm install
      - run: node version-check.js
      - run: npm install
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+      - run: npm test
+      - run: npm lint
```
drubin, line 23: Missing new line.

drubin:
Can these be activated before merging? or I guess they can't

.github/workflows/test.yml
brendandburns: done (removed '8')

.github/workflows/test.yml
brendandburns: done.

.github/workflows/test.yml
brendandburns: done.

brendandburns:
Comments addressed. There is some tool that looks like it can run it locally (https://github.com/nektos/act) if I run into trouble validating this or need to make a bunch of changes, I will try that.

brendandburns:
@drubin I think the `/lgtm` needs to be on it's own line :)

drubin:
/lgtm

Sorry :)

PR #434, opened 2020-04-20
Fix lint action, fix lint error.

brendandburns:
cc @drubin this is what I get for not testing this :P


drubin:
I mean to be fair I didn't notice it either

/lgtm

Wow github actions are quick!

PR #437, opened 2020-04-25
[DOC] Show an example of creating an ingress

rayfoss:
Something nearly everyone actually does, an is actually quite painful if you're not used to typescript or how kubeConfig works.


brendandburns:
/lgtm
/approve

Thanks!

Issue #375, opened 2019-12-03
Interest in generic object API class


ddgenome:
I've created a generic Kubernetes object API class that allows you to pass any valid Kubernetes spec object to the action, e.g., create/delete/list/patch, that you want to perform.

https://github.com/atomist/sdm-pack-k8s/blob/43f41c251527f42fd9859453c582ee878af1bb96/lib/kubernetes/api.ts

This allows you to more easily perform tasks equivalent to `kubectl create|apply|delete` using the SDK.

You can see how it can be used to mimic the functionality of `kubectl apply` here:

https://github.com/atomist/sdm-pack-k8s/blob/43f41c251527f42fd9859453c582ee878af1bb96/lib/kubernetes/apply.ts

Is there any interest in incorporating this functionality into this repo's official Kubernetes JavaScript/Node.js client?

brendandburns:
Sure, this looks very useful. We'd welcome a PR.

Can you include an example of how to use it for the examples directory as well?

Thanks!

ddgenome:
Sounds good. I'll start work on it.

travisghansen:
This. Also generic watches (that reconnect and detect dead peers) by simply passing the endpoint without all the autogenerated garbage.

binura-g:
@ddgenome This is extremely useful, thanks! 💯

ddgenome:
I did not include a generic list or watch for the initial PR. List didn't seem to make sense, since if you have a spec, you'd read not list. As this was already delayed several months due to availability, I did not want to further delay it figuring out how to do a generic watch.

PR #446, opened 2020-05-11
Add KubernetesObjectApi

ddgenome:
Add generic object API that dynamically determines endpoint based on
the provided spec.  Add KubernetesObjectApi to index exports.  Provide
tests and example usage.

Closes kubernetes-client/javascript#375

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -51,31 +51,15 @@ export class KubernetesObjectApi extends ApisApi {
      */
     public static makeApiClient(kc: KubeConfig): KubernetesObjectApi {
         const client = kc.makeApiClient(KubernetesObjectApi);
-        if (kc.currentContext) {
-            const currentContext = kc.getContextObject(kc.currentContext);
-            if (currentContext && currentContext.namespace) {
-                client.defaultNamespace = currentContext.namespace;
-            }
-        }
+        client.setDefaultNamespace(kc);
         return client;
     }
 
-    /**
-     * Return whether the name of the resource should be appended to the API URI path.  When creating resources, it is
-     * not appended.
-     *
-     * @param action API action, see [[K8sApiAction]]
-     * @return true if name should be appended to URI
-     */
-    protected static appendName(action: KubernetesApiAction): boolean {
-        return action !== 'create';
-    }
-
-    /** Default default namespace, so to speak. */
-    private static readonly defaultDefaultNamespace = 'default';
-
     /** Initialize the default namespace.  May be overwritten by context. */
-    protected defaultNamespace: string = KubernetesObjectApi.defaultDefaultNamespace;
+    protected defaultNamespace: string = 'default';
+
+    /** Cache resource API response. */
+    protected apiVersionResourceCache: Record<string, V1APIResourceList> = {};
 
     /**
      * Create any Kubernetes resource.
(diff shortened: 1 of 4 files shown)


ddgenome:
One thing to note here is that I changed the `makeApiClient` semantics to allow for the generic API to more closely mimic the behavior of `kubectl`. Specifically, I needed to determine if namespace is set in the current context and default to that value. If it is more desirable to maintain the `makeApiClient` semantics, that functionality can be removed and this can just default to the "default" namespace.

src/object.ts
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    /**
+     * Return whether the name of the resource should be appended to the API URI path.  When creating resources, it is
+     * not appended.
+     *
+     * @param action API action, see [[K8sApiAction]]
+     * @return true if name should be appended to URI
+     */
+    protected static appendName(action: KubernetesApiAction): boolean {
```
brendandburns, line 70: This method feels unnecessary since it's just a single boolean compare. Just use the comparison everywhere.

src/object.ts
```diff
@@ -0,0 +1,538 @@
...
+        return action !== 'create';
+    }
+
+    /** Default default namespace, so to speak. */
+    private static readonly defaultDefaultNamespace = 'default';
```
brendandburns, line 75: this doesn't seem necessary, just use `'default'` below.

src/object.ts
```diff
@@ -0,0 +1,538 @@
...
+            method: 'GET',
+            qs: localVarQueryParameters,
+            headers: localVarHeaderParams,
+            uri: localVarPath,
+            useQuerystring: this._useQuerystring,
+            json: true,
+        };
+
+        try {
+            const getApiResponse = await this.requestPromise<V1APIResourceList>(
```
brendandburns, line 488: Probably want to cache this lookup in memory in the client for the sake of performance.

Not required to merge this PR, but at least add a TODO and file an issue.

(or just add it to this PR)

brendandburns:
Thanks for the PR. I had a few comments. The only substantial one is wrt to caching the API info lookup.

src/object.ts
ddgenome: How long should it be cached? The response will change as CRDs come and go. Does the client already have a caching solution that can be used here?

src/object.ts
ddgenome: Will do.

src/object.ts
ddgenome: I'd prefer to keep this as the name of the method serves as documentation of the intent. The logic expressed here:

```typescript
        if (KubernetesObjectApi.appendName(action)) {
            if (!spec.metadata.name) {
                throw new Error('Required spec property name is not set');
            }
            parts.push(encodeURIComponent(String(spec.metadata.name)));
        }
```

is more clear than this:

```
        if (action !== 'create') {
            if (!spec.metadata.name) {
                throw new Error('Required spec property name is not set');
            }
            parts.push(encodeURIComponent(String(spec.metadata.name)));
        }
```

Although admittedly, the utility of the method was higher when `list` was supported and the body of the method looked `!(action === 'create' || action === 'list')`.

ddgenome: Although perhaps the error being thrown communicates enough.

src/object.ts
brendandburns: I would cache until you fail to find something, and then I would try reloading.

brendandburns: `kubectl` caches in `~/.kube/schema` fwiw.

src/object.ts
brendandburns: I actually feel like the one with `'create'` is clearer. Otherwise I have to jump into the method to understand why/when you need to append the name.

ddgenome: I did change it as you requested.

src/object.ts
ddgenome: Cache added.

ddgenome:
All review comments provided thus far have been addressed.

PR #474, opened 2020-07-10
Update apply-example.ts

Leo7654:
https://github.com/kubernetes/kubectl/blob/07dc513798576e6263f798902ef156e500ae09db/pkg/cmd/patch/patch.go#L76
https://github.com/kubernetes/kubectl/blob/1c9bdd2137b2255e4f40424e8fd55990208cf3bc/pkg/cmd/apply/apply_test.go#L1036

### DIFF
--- a/examples/typescript/apply/apply-example.ts
+++ b/examples/typescript/apply/apply-example.ts
@@ -31,7 +31,9 @@ export async function apply(specPath: string): Promise<k8s.KubernetesObject[]> {
             // block.
             await client.read(spec);
             // we got the resource, so it exists, so patch it
-            const response = await client.patch(spec);
+            const options = { headers: { 'content-type': 'application/merge-patch+json' } }
+            const response =
+                await objectApi.patch(spec, undefined, undefined, undefined, undefined, options)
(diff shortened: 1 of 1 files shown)


brendandburns:
This looks fine to me, but it is failing linting.

Run `npm run lint` in your local client and fix any errors.

examples/typescript/apply/apply-example.ts
```diff
@@ -31,7 +31,9 @@ export async function apply(specPath: string): Promise<k8s.KubernetesObject[]> {
             // block.
             await client.read(spec);
             // we got the resource, so it exists, so patch it
-            const response = await client.patch(spec);
+            const options = { headers: { 'content-type': 'application/merge-patch+json' } }
```
ddgenome, line 34: Like `kubectl apply`, the object API defaults to strategic merge patching, https://github.com/kubernetes-client/javascript/blob/b218eb9e7f6080239b7061d42a088749b1d4a749/src/object.ts#L444-L446 . Without an explanation or comment, it seems to me changing from the default may lead to confusion.

PR #476, opened 2020-07-14
Create examples/scale-deployment.js

phantomlsh:
A simple example of scaling a particular deployment by namespace and name

### DIFF
--- a/examples/scale-deployment.js
+++ b/examples/scale-deployment.js
@@ -1,7 +1,7 @@
 const k8s = require('@kubernetes/client-node');
 
 const kc = new k8s.KubeConfig();
-kc.loadFromFile('./config');
+kc.loadFromDefault();
 
 const k8sApi = kc.makeApiClient(k8s.AppsV1Api);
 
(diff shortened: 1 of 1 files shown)


phantomlsh:
I signed it

examples/scale-deployment.js
```diff
@@ -0,0 +1,29 @@
+const k8s = require('@kubernetes/client-node');
+
+const kc = new k8s.KubeConfig();
+kc.loadFromFile('./config');
```
brendandburns, line 4: Can you switch this to: `kc.loadFromDefault()` since that's the happy path for most users.

examples/scale-deployment.js
```diff
@@ -0,0 +1,29 @@
...
+
+const k8sApi = kc.makeApiClient(k8s.AppsV1Api);
+
+const targetDeploymentName = 'docker-test-deployment';
+
+async function scale(namespace, name, replicas) {
+  // find the particular deployment
+  const res = await k8sApi.listNamespacedDeployment(namespace);
```
brendandburns, line 12: This should probably be `k8sApi.readNamespacedDeployent(namespace, name);` Then you don't need the `for` loop.

brendandburns:
Thanks for sending a PR for this! I added a couple small comments then this is ready to merge.

PR #477, opened 2020-07-20
Remove unnecessary assignment in constructor

jvassev:


drubin:
Hi, @jvassev firstly welcome, and thanks for your first contribution. :)  

These variables are used in other functions later down in the file, could you maybe explain a bit why this is needed?

```js
await this.watch.watch(
            this.path,
            { resourceVersion: list.metadata!.resourceVersion },
            this.watchHandler.bind(this),
            this.doneHandler.bind(this),
        );
```

Surprisingly the tests don't fail.

In order for us to merge any contribution you will need to sign the CLA, instructions should be in the comment above, let us know if you have any issues.

jvassev:
Hi @drubin,
The properties get assigned to because the constructor uses the `public constructor(private readonly path: string)` syntax which causes the assignment.

I found this while inspecting the emitted JS and stumbled on (notice the repeated assignments):
```
class ListWatch {
    constructor(path, watch, listFn, autoStart = true) {
        this.path = path;
        this.watch = watch;
        this.listFn = listFn;
        this.objects = [];
        this.indexCache = {};
        this.callbackCache = {};
        this.watch = watch;
        this.listFn = listFn;
```

So this is just a cleanup, no changes in the logic.

brendandburns:
/approve

Thanks for the PR!

PR #492, opened 2020-08-31
fix basic auth error in WebSocketHandler (#488)

### LINKED ISSUE #488: portForward generates Unexpected server response: 401 on websocket with sample
Hi All,

im trying to find out what im doing wrong or maybe its just my config/setup

the sample for the portFoward isnt working as i keep getting an error saying `Unexpected server response: 401` in the console

the error only appears when i try and connect to the localport on my laptop

however when i call the listnamespacedpod command, it returns the pods listed in that namespace so im abit confused?

my code is very simple
```
const k8s = require("@kubernetes/client-node");
const net = require("net");
kc.loadFromDefault();
k8sApi.listNamespacedPod("kubenav").then((res) => {
  console.log(res.body);
});
const forward = new k8s.PortForward(kc);
const server = net.createServer((socket) => {
  forward.portForward("kubenav","kubenav-5b9bdbb9cc-b4qtd",[14122],socket,null,socket);
});
server.listen(14122, 'localhost');
```
any my kubeconfig looks like this
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxxxxx
    server: https://kubernetes.xxxxxx.com:6443
  name: default
contexts:
- context:
    cluster: default
    user: default
  name: default
current-context: default
kind: Config
preferences: {}
users:
- name: default
  user:
    password: xxxxxx
    username: admin
```

edvald:
Updates the `ws` library, to pull in this fix: https://github.com/websockets/ws/commit/eaf2ad1335ff7b502a7b119c576ea3c611c7a70c

Fixes #488

### DIFF
--- a/src/yaml.ts
+++ b/src/yaml.ts
@@ -1,7 +1,7 @@
 import * as yaml from 'js-yaml';
 
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    return yaml.safeLoad(data, opts) as T;
+export function loadYaml<T = any>(data: string, opts?: yaml.LoadOptions): T | undefined {
+    return (yaml.safeLoad(data, opts) as unknown) as T | undefined;
 }
 
 export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
--- a/package-lock.json
+++ b/package-lock.json
@@ -382,11 +382,6 @@
       "integrity": "sha512-jgsaNduz+ndvGyFt3uSuWqvy4lCnIJiovtouQN5JZHOKCS2QuhEdbcQHFhVksz2N2U9hXJo8odG7ETyWlEeuDw==",
       "dev": true
     },
-    "async-limiter": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/async-limiter/-/async-limiter-1.0.0.tgz",
-      "integrity": "sha512-jp/uFnooOiO+L211eZOoSyzpOITMXx1rBITauYykG3BRYPu8h0UcxsPNB04RR5vo4Tyz3+ay17tR6JVf9qzYWg=="
-    },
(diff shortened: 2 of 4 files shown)


brendandburns:
Thanks for the PR.

Can you separate the YAML fixes out and create a second PR with those changes as I don't think they're related to the version bump. Then we can merge.

Thanks!

si458:
@edvald can you fix the PR at all please?

edvald:
I believe those type errors came up due to package updates, so I'm not sure I can peel that out without breaking tests. Would a separate commit suffice?

si458:
thanks @edvald for the fix, it does indeed work perfectly! i could of spend months trying to find out why it wasn't working!
@brendandburns i have created my own PR #493 with just the update to the WS module 👍

src/yaml.ts
```diff
@@ -1,7 +1,7 @@
 import * as yaml from 'js-yaml';
 
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    return yaml.safeLoad(data, opts) as T;
+export function loadYaml<T = any>(data: string, opts?: yaml.LoadOptions): T | undefined {
+    return (yaml.safeLoad(data, opts) as unknown) as T | undefined;
```
drubin, line 4: Can you please add a test case that tests for these condition changes? I think I understand how this behaves differently but it would be good to add a test to make sure we don't break this behaviour in future.

drubin (changes requested):

src/yaml.ts
edvald: I suppose I could, but it's completely trivial tbh, and the compiler will catch any improper use of the function (which would basically involve not handling the undefined case). Anyway, since this type error may only have come up on my side, possibly due to TypeScript version differences, I can just close the PR in favor of #493.

PR #500, opened 2020-09-14
Add option to filter invalid items when loading a config

### LINKED ISSUE #499: loadFromString is more restrictive then what is available in kubectl
I have tried to see what `kubectl` does when a system config has some bad entries (like an entry with `""` as the context). It simply skips them. 

I think that there should be an option for `loadFromString`, `loadFromDefault`, and `loadFromFile` to filter out any bad entries

Nokel81:
- Default is "Throw"
- Could be later extended to return a list of errors along side the
  config

Fixes #499

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3352,9 +3352,9 @@
       }
     },
     "underscore": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.9.1.tgz",
-      "integrity": "sha512-5/4etnCkd9c8gwgowi5/om/mYO5ajCaOgdzj/oW+0eQV9WxKBDZw5+ycmKmeaTXjInS/W0BzpGLo2xR2aBwZdg=="
+      "version": "1.11.0",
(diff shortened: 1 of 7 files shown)


Nokel81:
I signed it

package-lock.json
```diff
@@ -3352,9 +3352,9 @@
       }
     },
     "underscore": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.9.1.tgz",
-      "integrity": "sha512-5/4etnCkd9c8gwgowi5/om/mYO5ajCaOgdzj/oW+0eQV9WxKBDZw5+ycmKmeaTXjInS/W0BzpGLo2xR2aBwZdg=="
+      "version": "1.11.0",
```
brendandburns, line 3355: Is there a reason we need to upgrade `underscore`?

brendandburns:
Generally looks good, but you need to sign the CLA.

package-lock.json
Nokel81: I thought there was, but looking again. No. Reverted.

PR #507, opened 2020-09-20
Require typedef in the style. Add some missing return types.

### LINKED ISSUE #449: typescript examples are actually javascript examples
Hi, thanks for this library first of all. I have one suggestion though:

Please add typescrip features to your examples, as they are now plain ecmascript mostly.

Most important feature that I am after: return type for `makeApiClient`, so that we get autocompletion in our editors.

brendandburns:
Fixes #449


brendandburns:
Friendly ping to @drubin if you have time (totally get it if you don't)

brendandburns:
(self-merging)

PR #508, opened 2020-09-20
Add some constants and an example of patching resources.

### LINKED ISSUE #450: Add an example of how to use patch.
See https://github.com/kubernetes-client/javascript/issues/443#issuecomment-628423914

brendandburns:
Fixes #450


brendandburns:
(self-merge)

PR #518, opened 2020-10-02
Create joiningstrings

rudrabisht:
In Java, we can also join two strings using the + operator.


brendandburns:
This isn't a valid PR. Closing.

PR #521, opened 2020-10-06
Update cp.ts

Nebojsa92:
It's faster to copy compressed data, and, in my case, it prevented following error:
"UnhandledPromiseRejectionWarning: Error: TAR_BAD_ARCHIVE: Truncated input (needed 31232 more bytes, only 0 available)"

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -28,7 +28,7 @@ export class Cp {
     ): Promise<void> {
         const tmpFile = tmp.fileSync();
         const tmpFileName = tmpFile.name;
-        const command = ['tar', 'cf', '-', srcPath];
+        const command = ['tar', 'zcf', '-', srcPath];
         const writerStream = fs.createWriteStream(tmpFileName);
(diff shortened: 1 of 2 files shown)


brendandburns:
This looks reasonable to me, but we need you to sign the CLA.

Tests also failed, but that looks like a flake and I've poked them to re-run.

Thanks!

brendandburns:
Actually the unit tests are failing w/ a legit problem, please fix the test too.

vip30:
It looks make sense to me too

Believe just change this line will make the test work
https://github.com/kubernetes-client/javascript/blob/master/src/cp_test.ts#L24

Nebojsa92 (commented):
Test file updated

Nebojsa92:
> This looks reasonable to me, but we need you to sign the CLA.
> 
> Tests also failed, but that looks like a flake and I've poked them to re-run.
> 
> Thanks!

I've changed cp_test.ts file.
Also I've signed CLA.

That should do it.

PR #526, opened 2020-10-15
Fix timeout on watch.js requestOptions

jhagestedt:
The watching of Kubernetes object stops after 60 seconds - as the default timeout of 60 seconds is used.
It should be set to 0 - so infinite or at least be more configureable.

Using the RequestInterface to configure this is not the best option as in line 70 the request options gets applied to the config but not taken from any passed configuration.

Also the forever flag could be set.

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -64,6 +64,8 @@ export class Watch {
             uri: url,
             useQuerystring: true,
             json: true,
+            forever: true,
+            timeout: 0,
         };
         await this.config.applyToRequest(requestOptions);
 


jhagestedt:
a new comment :)

brendandburns:
change looks ok, but validation tests are failing

jhagestedt:
> change looks ok, but validation tests are failing

I think because of other changes not mine :D

brendandburns:
Yep, looks like a different problem :( I will fix and then we can merge this. Not sure how things got broken.

brendandburns:
@jhagestedt I fixed the CI/CD errors, please re-base and I think this will be ready to go.

brendandburns:
@jhagestedt thanks for updating the PR, but your rebase got messed up somehow (there are like 14 commits that aren't part of this PR)

Once you fix up the PR, I'm happy to merge this PR.

jkryl:
Hi @jhagestedt, I was looking at the fix as I was fixing another problem in the same area. It's not 100% clear to me if these newly added options have any impact on the client. I'm looking at request library source code from master:

The `forever` option sets keep-alive for HTTP connection, which seems to be the exact opposite of what we want. AFAIK keep-alive keeps a connection alive between the requests. However, a watcher connection lasts until closed from either side. We should never get into a situation when we want to keep it alive for another request.

We set the `timeout` option however the first thing that request lib seems to do is delete the option and implement its own timeout (https://github.com/request/request/blob/3c0cddc7c8eb60b470e9519da85896ed7ee0081e/request.js#L748). So I'm wondering if that has any impact on behaviour of the client at all. I have seen issues with inactive watcher connections when using AKS on azure, so there seems to be a problem with that for sure. I'm just not clear how this can fix the problem.

Issue #530, opened 2020-10-22
regenerate api with kubernetes v1.19.x


weikinhuang:
Hi, can someone please regenerate the spec to the latest version of kubernetes, I'm trying to make use of v1 ingress. Thanks!

brendandburns:
You can use the instructions here:
https://github.com/kubernetes-client/javascript#re-generating-code

Otherwise we'll get to this eventually.

brendandburns:
Closing since #534 has merged.

I will cut a 0.13.0 release soon (once #535  merges)

weikinhuang:
Thanks!

brendandburns:
0.13.0 has been pushed to npm, closing this issue.

PR #534, opened 2020-10-23
Upgrade to 1.19 API

weikinhuang:
Regenerate for kubernetes v1.19 to include newly graduated schemas such as `networking.k8s.io/v1 Ingress`. A few alpha level specs were also removed with this regeneration.

Ref https://github.com/kubernetes-client/javascript/issues/530


brendandburns:
Thanks @weikinhuang !

PR #538, opened 2020-10-31
Get ready for the 0.13.0 release.

brendandburns:
cc @drubin @itowlson

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,11 +118,11 @@ Generally speaking newer clients will work with older Kubernetes, but compatabil
 
 Key: 
 
-* `✓` Exactly the same features / API objects in both java-client and the Kubernetes
+* `✓` Exactly the same features / API objects in both javascript-client and the Kubernetes
   version.
-* `+` java-client has features or api objects that may not be present in the
+* `+` javascript-client has features or api objects that may not be present in the
   Kubernetes cluster, but everything they have in common will work.
-* `-` The Kubernetes cluster has features the java-client library can't use
+* `-` The Kubernetes cluster has features the javascript-client library can't use
   (additional API objects, etc).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -103,6 +103,33 @@ const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
...
+* `✓` Exactly the same features / API objects in both java-client and the Kubernetes
+  version.
+* `+` java-client has features or api objects that may not be present in the
+  Kubernetes cluster, but everything they have in common will work.
+* `-` The Kubernetes cluster has features the java-client library can't use
+  (additional API objects, etc).
+* `x` The Kubernetes cluster has no guarantees to support the API client of
+  this version, as it only promises _n_-2 version support. It is not tested,
+  and operations using API versions that have been deprecated and removed in
+  later server versions won't function correctly.
```
Flydiverny, line 130: java -> javascript

```suggestion
* `✓` Exactly the same features / API objects in both javascript-client and the Kubernetes
  version.
* `+` javascript-client has features or api objects that may not be present in the
  Kubernetes cluster, but everything they have in common will work.
* `-` The Kubernetes cluster has features the javascript-client library can't use
  (additional API objects, etc).
* `x` The Kubernetes cluster has no guarantees to support the API client of
  this version, as it only promises _n_-2 version support. It is not tested,
  and operations using API versions that have been deprecated and removed in
  later server versions won't function correctly.
```

drubin:
Looks good with the suggested fixes. Tag me again and I will add the tag.

If I do it now it will get auto merged

README.md
brendandburns: The perils of cut and paste! (thanks for the edits, applied)

brendandburns:
@drubin comments addressed, please re-check.

Thanks!

PR #539, opened 2020-11-01
types(config): improve loadFromOptions

moshfeu:
Also, as far as I know, if an array typed, there is no need to declare the type of the item in functions such as `forEach`.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -164,7 +164,12 @@ export class KubeConfig {
         this.currentContext = obj['current-context'];
     }
 
-    public loadFromOptions(options: any): void {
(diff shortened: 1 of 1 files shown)


moshfeu:
NM, it seems more complicated that I thought. Locally it build failed but because different (existing) code such as  https://github.com/kubernetes-client/javascript/blob/6d53ddcfa96418ec9409d8bce0e6e5a01400dc90/src/config.ts#L161

> Object is possibly 'undefined'.

So if someone wants to take it from here, that will be great, I'm not a big fan of `any` 🙄

PR #546, opened 2020-11-03
fix tests on windows.

brendandburns:
where "fix" mostly means "disable that test" *sigh*


drubin:
I don't really get this. These tests fail on Windows which is expected since the code doesn't actually work does it?

PR #555, opened 2020-11-13
split watch error and watch done handlers.

### LINKED ISSUE #552: Done handler gets called twice when watch aborts
When trying to do a continuous watch on kubernetes resources I noticed that the done handler gets called twice when the connection to the kubernetes api server aborts unexpectedly.

I don't have time to look deeper into this at the moment, but I guess it's because the `error` and `close` events both get fired on the request and both of them trigger the callback function.

Currently I just use a debounce function as a workaround, but we should find a better way of handling this.

Reference:

https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts#L82
https://github.com/kubernetes-client/javascript/blob/master/src/watch.ts#L92

My code:

```
watchFunction(): void {
    this.watchClient.watch(
      '/api/v1/nodes',
      {},
      async (phase: string, node: V1Node) => {
        // do some magic
      },
      (err) => {
        console.log(`Failed to watch nodes.. restarting watch`);
        this.watchFunction()
      },
    );
}
```

Workaround:

```
watchFunction(): void {
    this.watchClient.watch(
      '/api/v1/nodes',
      {},
      async (phase: string, node: V1Node) => {
        // do some magic
      },
      debounce(() => {
        console.log(`Failed to watch nodes.. restarting watch`);
        this.watchFunction();
      }, 1000),
    );
}
```

brendandburns:
fixes #552

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -12,6 +12,7 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     private resourceVersion: string;
     private readonly indexCache: { [key: string]: T[] } = {};
     private readonly callbackCache: { [key: string]: Array<ObjectCallback<T>> } = {};
+    private stopped: boolean;
 
     public constructor(
         private readonly path: string,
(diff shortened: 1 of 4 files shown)


brendandburns:
self LGTM given LGTM in the issue thread (https://github.com/kubernetes-client/javascript/issues/552#issuecomment-727074337)

PR #569, opened 2020-12-16
Add getAPIResources to apisApi

Nokel81:
Currently this API is lacking from this library but a similar one exists within kubernetes's client-go.


brendandburns:
This code is generated. We can't hand-edit it.

Also, the API is defined here:
https://github.com/kubernetes-client/javascript/blob/master/src/gen/api/apiextensionsV1Api.ts#L408

Please let us know if there are any problems using that API.

Nokel81:
Thanks for the pointer that the code is generated, I missed that in the file.

The problem with using `ApiextensionsV1Api` is that only get resources under `/apis/apiextensions.k8s.io/` but I want just `/apis/`. I see that `getApiResources` is a common method for all the `*Api` types except `ApisApi` which is why I wanted to add this.

This functionality is available in `client-go` under the `DiscoveryClient` struct, so I would assume that it should also be here.

Nokel81:
@brendandburns Though I see that my PR (even if it wasn't against an auto-gen-ed file) doesn't work, I misunderstood the `client-go` code and this code needs to list the `apiVersions` first.

brendandburns:
Since this code is generated, and so if you want changes we need to update the OpenAPI specification and regenerate rather than edit this file.

The spec is here:
https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json

brendandburns:
Alternately, you could make a "raw" request, and move your code to a different file (`discovery.ts`?)

There's an example of a raw request here:
https://github.com/kubernetes-client/javascript/blob/master/examples/raw-example.js

Nokel81:
Thanks for the pointers. I think a combination might be good. Should I close this PR and open another one?

brendandburns:
@Nokel81 it's up to you if you want to close this PR and open a new one, or just update/modify this PR with the alternate approach.

Thanks!

brendandburns:
For reference, there is an approximately equivalent Java implementation here:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/Discovery.java

TattdCodeMonkey:
@Nokel81 The code you added does not return  a `V1APIResourceList` you would need to add a `groupVersion` to the `/apis/` uri to get a `V1APIResourceList`. Making a request to `/apis/` returns a `V1APIGroupList` which you can use `getAPIVersions` for.

If you want a `V1APIResourceList` you will need to either use the Api for the group, for example to get all the resources in networking group  you would call [getAPIResources](https://github.com/kubernetes-client/javascript/blob/master/src/gen/api/networkingV1Api.ts#L1074) with the `NetworkingV1Api`. 

Or if you don't know the group until runtime, you can build a raw request to do the same like this:

```typescript
const k8s = require('@kubernetes/client-node');
const request = require('request');

kc.loadFromDefault();

const opts = {};
kc.applyToRequest(opts);

request.get(`${kc.getCurrentCluster().server}/apis/networking.k8s.io/v1`, opts,
    (error, response, body) => {
        if (error) {
            console.log(`error: ${error}`);
        }
        if (response) {
            console.log(`statusCode: ${response.statusCode}`);
        }
        console.log(`body: ${body}`);
  });
```

Nokel81:
I am going to close this as what needs to be done is something different and I don't know enough about how the API is generated.

PR #580, opened 2021-01-24
Preserve context from starting file

### LINKED ISSUE #579: KubeConfig.loadFromDefault currentContext is inconsistent with kubectl with multiple kubeconfigs
When using multiple kubeconfigs through the `KUBECONFIG` environment variable, `kubectl` always uses the first file in the `KUBECONFIG` environment variable to grab the current context.

One example found here
https://github.com/kubernetes/kubectl/blob/952f50e545b14e0a2a0781ef47c342cdba5a8c54/pkg/cmd/config/current_context.go#L62-L76

However, because `loadFromDefault` uses `mergeConfig`, it always overwrites the `currentContext` with next kubeconfig it merges, causing it to use the last kubeconfig's context.
https://github.com/kubernetes-client/javascript/blob/a8d4f1c7bea2b27403d380e22e492b6680f80b31/src/config.ts#L234-L245

This causes using `kubectl` to change the context to not affect `loadFromDefault` when making CLI tools where the users sets the `KUBECONFIG` environment variable with multiple kubeconfigs since `kubectl use-context` sets the context in the first kubeconfig in the environment variable.
https://github.com/kubernetes/kubectl/blob/952f50e545b14e0a2a0781ef47c342cdba5a8c54/pkg/cmd/config/use_context.go#L65-L79

adrianord:
fixes  #579

I wasn't sure whether to put this argument in the `ConfigOptions` as it looks like that object gets passed down to `loadFromFile`. Let me know and I can change it.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -231,8 +231,10 @@ export class KubeConfig {
         this.currentContext = contextName;
     }
 
-    public mergeConfig(config: KubeConfig): void {
-        this.currentContext = config.currentContext;
+    public mergeConfig(config: KubeConfig, preserveContext: boolean = false): void {
+        if (!preserveContext) {
+            this.currentContext = config.currentContext;
+        }
         config.clusters.forEach((cluster: Cluster) => {
             this.addCluster(cluster);
         });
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for the PR!

/lgtm
/approve

PR #584, opened 2021-02-10
Expose types from config_types.ts

### LINKED ISSUE #583: Expose types from config_types.ts
It would be cool to access types that are needed for `loadFromClusterAndUser`  like e.g. `Cluster` or `User`.
Those are defined in [config_types.ts](https://github.com/kubernetes-client/javascript/blob/master/src/config_types.ts) which is not exported at the moment.

Agreon:
Closes #583

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,3 +13,4 @@ export * from './top';
 export * from './object';
 export * from './cp';
 export * from './patch';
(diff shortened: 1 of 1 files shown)


Agreon:
Signed cla

brendandburns:
/lgtm
/approve

Thanks!

PR #585, opened 2021-02-12
Regenerate client for Kubernetes 1.20.0, Rev package version to 0.14.0

brendandburns:
cc @drubin @jkryl


drubin:
@brendandburns the tests seem to fail but not 100% sure why. So something seems funky.

in general, the changes seem great but I would suggest splitting the manual changes and auto generated changes into 2 commits.

brendandburns:
@drubin thanks for review. I forgot to include some new files (forgot `git add`) so it worked locally, but obviously not remotely (thank goodness for CI/CD :)

Anyway, it's fixed, and I split out the manual changes into their own commit.  Please take another look.

Thanks!

drubin:
/lgtm

Thanks for the quick turnaround and for improving the commits.

PR #588, opened 2021-02-17
add list method

doguh:
Add `list` method to `KubernetesObjectApi`.

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -20,7 +20,7 @@ type KubernetesObjectResponseBody =
     | V1APIResourceList;
 
 /** Kubernetes API verbs. */
-type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'replace';
+type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'replace' | 'list';
 
 /**
  * Valid Content-Type header values for patch operations.  See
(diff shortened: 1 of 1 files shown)


src/object.ts
```diff
@@ -315,6 +315,57 @@ export class KubernetesObjectApi extends ApisApi {
...
+     * @param pretty If \&#39;true\&#39;, then the output is pretty printed.
+     * @param exact Should the export be exact.  Exact export maintains cluster-specific fields like
+     *        \&#39;Namespace\&#39;. Deprecated. Planned for removal in 1.18.
+     * @param exportt Should this value be exported.  Export strips fields that a user can not
+     *        specify. Deprecated. Planned for removal in 1.18.
+     * @param options Optional headers to use in the request.
+     * @return Promise containing the request response and list of [[KubernetesObject]].
+     */
+    public async list(
+        spec: KubernetesObject,
```
brendandburns, line 330: I don't think we want a `KubernetesObject` here, instead pass in `apiGroup` `apiVersion` & `kind` as well as a `namespaced` parameter.

We'll need to refactor the specUriPath code also.

brendandburns:
One comment, thanks for the PR.

We need you to sign the CNCF CLA.

Thanks!

PR #597, opened 2021-02-25
cache: skip callback upon restart if object did not change

jkryl:
Currently MOD callbacks are called for all objects in the cache when
the watcher is restarted. This commit makes the algorithm smarter
and call MOD callback only if the object has really changed. The decision
is based on "resourceVersion" metadata field in the object.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -197,15 +197,25 @@ export function addOrUpdateObject<T extends KubernetesObject>(
             addCallback.forEach((elt: ObjectCallback<T>) => elt(obj));
         }
     } else {
-        objects[ix] = obj;
-        if (updateCallback) {
-            updateCallback.forEach((elt: ObjectCallback<T>) => elt(obj));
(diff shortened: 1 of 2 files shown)


src/cache.ts
```diff
@@ -208,6 +210,13 @@ function isSameObject<T extends KubernetesObject>(o1: T, o2: T): boolean {
     return o1.metadata!.name === o2.metadata!.name && o1.metadata!.namespace === o2.metadata!.namespace;
 }
 
+function isSameVersion<T extends KubernetesObject>(o1: T, o2: T): boolean {
+    return (
+        o1.metadata!.resourceVersion !== undefined &&
```
brendandburns, line 215: probably worth testing for null also?

brendandburns:
Thanks! One comment wrt null checking

src/cache.ts
jkryl: thanks for the review! I will add the null check.

PR #510, opened 2020-09-22
Update openid-client version to 4.1.1

NicolasFloquet:
Fixes #497

PR #605, opened 2021-03-15
Use the public npm registry

### LINKED ISSUE #594: Documentation not available
When trying to generate the documentation as per https://github.com/kubernetes-client/javascript/blob/master/README.md#documentation, I get the following errors from `npm install` :
image

Deleting the lockfile allows me to go further with the installation, however I am later met with a number of typescript error preventing me from running  `npm run build` successfully.

In order to reproduce the problem:
Clone the repository
Run `npm install`

Expected result:
Being able to `npm run docs`

Actual result:
`npm install` failure(s)

dominykas:
Should fix #594

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -351,17 +351,17 @@
     },
     "@panva/asn1.js": {
       "version": "1.0.0",
-      "resolved": "https://artifactory.thalesdigital.io/artifactory/api/npm/npm/@panva/asn1.js/-/asn1.js-1.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/@panva/asn1.js/-/asn1.js-1.0.0.tgz",
       "integrity": "sha1-3VWue4Ep4CBJ8AlAi5fGHM+QMvY="
     },
     "@sindresorhus/is": {
       "version": "3.1.2",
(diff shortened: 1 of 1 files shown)


dominykas:
I signed it

drubin:
hi @dominykas welcome :) and thanks for jumping in and fixing this. 

As you can see this was added in https://github.com/kubernetes-client/javascript/pull/510 which both @brendandburns and I reviewed. I think this must have slipped through the cracks. We should add a CI check to ensure only the public registry is used but quick googling didn't show up any built in configs to ensure this.

Similar to https://docs.npmjs.com/cli/v6/configuring-npm/package-json#publishconfig but for pulling of NPM packages.

I am curious to know why these tests kept passing in CI though. 

/lgtm 
/assign @brendandburns

dominykas:
> We should add a CI check to ensure only the public registry is used but quick googling didn't show up any built in configs to ensure this.

Indeed I'm not aware of a thing that can enforce this, short of writing your own lint tool... Which is why I created an RRFC in npm to cover this case: https://github.com/npm/rfcs/issues/340.

Maybe someone already has a linter for that 🤷‍♂️ It shouldn't be hard to create one.

brendandburns:
Thanks for the PR, apologies for letting this slip in.

/lgtm
/approve

PR #607, opened 2021-03-15
Use npm ci as safer install option for CI tests

drubin:
References https://github.com/kubernetes-client/javascript/pull/605

An attempt to get CI to fail when using a non public registry

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,4 +24,5 @@ jobs:
       - run: npm ci
       - run: npm test
       - run: npm run lint
+      - run: npm audit –audit-level=critical
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,5 +24,5 @@ jobs:
       - run: npm ci
       - run: npm test
       - run: npm run lint
-      - run: npm audit –audit-level=critical
+      - run: npm audit --audit-level=critical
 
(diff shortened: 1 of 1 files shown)


brendandburns:
Seems to work. Now that the fix is merged, let's rebase this and see if it goes green.

drubin:
@brendandburns this is actually failing for another reason, but I will continue to try figure it out.. and see if I can get this to fail with invalid registries. 

```
npm ERR! arg Argument starts with non-ascii dash, this is probably invalid: [ '–audit-level=critical' ]
npm ERR! Invalid audit subcommand: `–audit-level=critical`
npm ERR! 
npm ERR! Usage:
npm ERR! 
npm ERR! npm audit [--json] [--production]
npm ERR! npm audit fix [--force|--package-lock-only|--dry-run|--production|--only=(dev|prod)]
```

brendandburns:
aha! your editor (OS X?) converted `--` to dashes to unicode long dash `–`

drubin:
@brendandburns I am not 100% sure how we want to handle this PR. 

It doesn't work as expected

It doesn't hurt to have these changes in any way, but I am not sure we want the security issues to be "blocking" ? 

I can split/rebase this so the changes are more logical and split between different commits, but I think we should add an issue for enforcing only public registry is used as we need to build something custom here as nothing seems to exist. 

Also, I am not sure if we want `audit-level=critical` or `audit-level=warn`

brendandburns:
I added https://github.com/kubernetes-client/javascript/issues/612

I think I like merging this, because I think that critical issues probably should be addressed quickly.

drubin:
Feel free to squash and merge or I will rebase the commits into a single commit later.
/assign @brendandburns

PR #613, opened 2021-03-19
Add a unit test to require registry.npm.js as the package registry.

### LINKED ISSUE #612: Enforce only public NPM registries in package-lock.json
A non-public NPM repository snuck into our config accidentally which broke things.

We should add a unit test that scans package-lock.json and requires npm as the registry.

brendandburns:
Fixes #612

### DIFF
--- a/src/package_test.ts
+++ b/src/package_test.ts
@@ -7,4 +7,21 @@ describe('package', () => {
         const v2 = require('../package-lock.json').version;
         expect(v1).to.equal(v2);
     });
+
+    it('package-lock should only reference npm', () => {
+        const validateDependencies = (deps) => {
+            if (!deps.dependencies) {
+                return;
+            }
+            for (const key in deps.dependencies) {
+                const dep = deps.dependencies[key];
+                const resolved = new URL(dep.resolved);
+                expect(resolved.hostname).to.equal('registry.npmjs.org');
+                expect(resolved.protocol).to.equal('https:');
+                validateDependencies(dep);
+            }
+        };
+        const deps = require('../package-lock.json');
+        validateDependencies(deps);
+    });
 });


src/package_test.ts
```diff
@@ -7,4 +7,13 @@ describe('package', () => {
         const v2 = require('../package-lock.json').version;
         expect(v1).to.equal(v2);
     });
+
+    it('package-lock should only reference npm', () => {
+        const deps = require('../package-lock.json').dependencies;
+        for (const key in deps) {
+            const dep = deps[key];
+            const resolved = new URL(dep.resolved);
+            expect(resolved.hostname).to.equal('registry.npmjs.org');
```
drubin, line 16: Can we also add a HTTPS check here too while we at it. 

```suggestion
            expect(resolved.hostname).to.equal('registry.npmjs.org');
            expect(resolved.protocol ).to.equal('https);
```

Its too easy to mistakenly switch to the HTTP registry and I doubt either you or I would notice it.

src/package_test.ts
```diff
@@ -7,4 +7,13 @@ describe('package', () => {
         const v2 = require('../package-lock.json').version;
         expect(v1).to.equal(v2);
     });
+
+    it('package-lock should only reference npm', () => {
```
drubin, line 11:
```suggestion
    it('package-lock should only reference official public npm registry', () => {
```

lets be more exact and specific here too.

drubin (changes requested):
Minor suggestion, but it looks good. if you make the changes I will approve.

Also thanks for this! Its surprisingly clean

brendandburns:
Comments addressed.

src/package_test.ts
```diff
@@ -7,4 +7,14 @@ describe('package', () => {
...
         expect(v1).to.equal(v2);
     });
+
+    it('package-lock should only reference official public npm registry', () => {
+        const deps = require('../package-lock.json').dependencies;
+        for (const key in deps) {
+            const dep = deps[key];
+            const resolved = new URL(dep.resolved);
+            expect(resolved.hostname).to.equal('registry.npmjs.org');
+            expect(resolved.protocol ).to.equal('https);
```
drubin, line 17:
```suggestion
            expect(resolved.protocol).to.equal('https);
```

not sure how this space got in here.

src/package_test.ts
```diff
@@ -7,4 +7,14 @@ describe('package', () => {
         const v2 = require('../package-lock.json').version;
         expect(v1).to.equal(v2);
     });
+
+    it('package-lock should only reference official public npm registry', () => {
+        const deps = require('../package-lock.json').dependencies;
+        for (const key in deps) {
```
dominykas, line 13: It's not enough to loop through the top level - each of `dependencies` can have its own `dependencies`, e.g. https://github.com/kubernetes-client/javascript/blob/d18e69d531580db628a4e925f0ece42a80a20412/package-lock.json#L38

src/package_test.ts
brendandburns: fixed.

src/package_test.ts
brendandburns: done.

brendandburns:

Thanks!

brendandburns:
@drubin @dominykas 

Friendly ping on this one.

Thanks!

dominykas (approved):

drubin:
/lgtm 

Thanks!

PR #614, opened 2021-03-19
Add a CONNECT event to informer.

### LINKED ISSUE #609: Emit an event when the informer reconnects the watch feed
My primary use case for this is just logging - there's an issue #596 about the watcher silently stopping and this might help debug the problem.

We're already operating a controller written in Node.js, and we've spotted similar issues - having some logs around the reconnection (and having them missing when things go south) has helped in trying to at 
least determine a rough range of time when the problems started occurring (fwiw - it seems to be related to `master` node rotation, although not easily reproducible).

It would be nice if `Informer` was able to emit something like `connected` each time it [restarts the watcher](https://github.com/kubernetes-client/javascript/blob/c13bf36994962843d9c81852165e169622e327f9/src/cache.ts#L116)?

brendandburns:
Fixes #609

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -4,6 +4,7 @@ import {
     CONNECT,
     DELETE,
     ERROR,
+    ErrorCallback,
     Informer,
     ListPromise,
     ObjectCallback,
@@ -21,7 +22,7 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     private objects: T[] = [];
     private resourceVersion: string;
     private readonly indexCache: { [key: string]: T[] } = {};
-    private readonly callbackCache: { [key: string]: Array<ObjectCallback<T>> } = {};
+    private readonly callbackCache: { [key: string]: Array<ObjectCallback<T> | ErrorCallback> } = {};
     private request: RequestResult | undefined;
     private stopped: boolean = false;
 
@@ -52,11 +53,13 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
         this._stop();
     }
 
-    public on(verb: string, cb: ObjectCallback<T>): void {
+    public on(verb: 'add' | 'update' | 'delete' | 'change', cb: ObjectCallback<T>): void;
+    public on(verb: 'error' | 'connect', cb: ErrorCallback): void;
+    public on(verb: string, cb: any): void {
         if (verb === CHANGE) {
-            this.on(ADD, cb);
-            this.on(UPDATE, cb);
-            this.on(DELETE, cb);
+            this.on('add', cb);
+            this.on('update', cb);
+            this.on('delete', cb);
             return;
         }
         if (this.callbackCache[verb] === undefined) {
@@ -65,11 +68,13 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
         this.callbackCache[verb].push(cb);
     }
 
-    public off(verb: string, cb: ObjectCallback<T>): void {
+    public off(verb: 'add' | 'update' | 'delete' | 'change', cb: ObjectCallback<T>): void;
+    public off(verb: 'error' | 'connect', cb: ErrorCallback): void;
+    public off(verb: string, cb: any): void {
         if (verb === CHANGE) {
-            this.off(ADD, cb);
-            this.off(UPDATE, cb);
-            this.off(DELETE, cb);
+            this.off('add', cb);
+            this.off('update', cb);
+            this.off('delete', cb);
             return;
         }
         if (this.callbackCache[verb] === undefined) {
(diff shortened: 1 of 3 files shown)


src/cache.ts
```diff
@@ -1,4 +1,4 @@
-import { ADD, DELETE, ERROR, Informer, ListPromise, ObjectCallback, UPDATE } from './informer';
+import { ADD, CONNECT, DELETE, ERROR, Informer, ListPromise, ObjectCallback, UPDATE } from './informer';
```
drubin, line 1: Does it make sense to add a disconnect here too while we are at it?

src/informer.ts
```diff
@@ -12,9 +12,13 @@ export type ListPromise<T extends KubernetesObject> = () => Promise<{
     body: KubernetesListObject<T>;
 }>;
 
+// These are issued per object
 export const ADD: string = 'add';
 export const UPDATE: string = 'update';
 export const DELETE: string = 'delete';
+
+// These are issued when a watch connects, reconnects or sees an error
```
drubin, line 20: I see we are only testing for one of these conditions, what about reconnect or errors?

src/cache.ts
```diff
@@ -97,6 +98,7 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
             // do not auto-restart
             return;
         }
+        this.callbackCache[CONNECT].forEach((elt: ObjectCallback<T>) => elt({} as T));
```
dominykas, line 101: Is `{} as T` the right thing to do? I would expect no params passed to the callback here (not sure if TypeScript allows expressing that though...)

This also caught me off guard for the `ERROR` event - `error` is an instance of `Error`, not `T`.

src/informer.ts
brendandburns: Added a test for reconnect. There was already a test for errors. I clarified this comment also.

src/cache.ts
brendandburns: I think connect + error is sufficient. There shouldn't be a disconnect w/o an error that isn't followed immediately by another connect.

src/cache.ts
brendandburns: This was legacy left over from when the callbacks only dealt with objects not errors and connects. I refactored the callback type signature to make it better.

brendandburns:

Thanks!

src/cache.ts
```diff
@@ -102,13 +113,14 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     private async doneHandler(err: any): Promise<any> {
         this._stop();
         if (err) {
-            this.callbackCache[ERROR].forEach((elt: ObjectCallback<T>) => elt(err));
+            this.callbackCache[ERROR].forEach((elt: ObjectCallback<T>) => elt(undefined, err));
```
dominykas, line 116: Hmmm. This makes it a breaking change? It also has the error at the end, which may be unintuitive to folks used to Node's error-first callbacks... But moving the error first would be even more breaking...

Not sure what's the best approach here...

Would be a travesty in TypeScript world to allow `on()` to have overloads to accept different parameter types? i.e. allow `on(verb: string, fn: ObjectCallback<T>): void;` as well as `on(verb: string, fn: ErrorCallback): void;`? While it wouldn't enforce the correct callback type based on the `verb`, it would at least be non-breaking? Not sure if there's a better way to express this in TS...

brendandburns: I did also try it that way:

`on(verb: string, fn: ObjectCallback<T> | ErrorCallback | ConnectCallback)`

But I felt like that was even messier, because suddenly the user can accidentally pass a `ConnectCallback` to a `on('add', ...)`

Maybe we could do:

```typescript
export type UntypedCallback = (err?: any, obj?: any) => void;
export type TypedCallback<T extends KubernetesObject> = (err?: any, obj?: T) => void;
export type ObjectCallback<T extends KubernetesObject> = (obj?: T, err?: any) => void;

...

on(verb: string, fn: ObjectCallback | TypedCallback | UntypedCallback)
...
```

wdyt? That has the benefit of not being a breaking change, and also adding in the more node-style callbacks.

brendandburns: Although, I'm not actually sure that TS can handle that since those two callback types may be indistinguishable...

dominykas: Would TS handle that at runtime or just compile time?

Did TS ever get enums? If the first param for `on()` was enum, rather than a `string`, then it could have multiple signatures based on enum type... Which would probably still be a breaking change from TS perspective, but possibly the old `string` signature could still be there, just with `any` callback so that people can upgrade their code eventually, even if they would lose out a little bit in terms of potentially accidentally passing the wrong callback?

Also I wonder if passing the wrong callback is even a problem for people using TS, because surely someone must have noticed by now that the callback for the `error` event is not right?

This is a common pattern (`on()` with different callbacks for different event types) - surely someone must have solved this by now? This is used in node itself (in streams and elsewhere) - I wonder how do they define the types 🤔 Let's see... 👀

drubin: was there a follow-up on this @brendandburns ?

brendandburns: Thanks for the pointer! I fixed this using type overloading.

brendandburns:
Comments addressed, please take another look!

dominykas (approved):
👏

drubin:
/lgtm

Thanks for fixing the comments

PR #630, opened 2021-04-07
send keep-alive probes every 30 seconds to the server

bacongobbler:
This sends keep-alive probes to the server every 30 seconds, similar to client-go. This should help in samples such as [the watch example](https://github.com/kubernetes-client/javascript/blob/master/examples/typescript/watch/watch-example.ts), as now a keepalive header should be sent by the client to keep the connection open.

Note that this option will apply to *only* watch requests. Kubernetes' client-go sets this up during the construction of the `transport.New` with a given client config, so I think theirs applies to all requests:

https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/transport/transport.go#L46-L53

https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/transport/cache.go#L90-L96

It is possible that alternative features like the `Log` with the `follow` option enabled might need this flag as well.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1014,7 +1014,6 @@
       "version": "4.1.1",
       "resolved": "https://registry.npmjs.org/debug/-/debug-4.1.1.tgz",
       "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
-      "dev": true,
       "requires": {
         "ms": "^2.1.1"
       },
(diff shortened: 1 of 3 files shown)


bacongobbler:
Hmm... Reading further into node's documentation, the http agent's `keepalive` feature is different than actually setting the `keep-alive` header. It just keeps old sockets around to be reused for future requests.

I'll see where I can find out how to send keep-alive probe requests to the server.

bacongobbler:
Done! Just had to do it at the socket level. Pushing up the lint fixes now and it should be ready for testing.

brendandburns:
/lgtm
/approve

Thanks!

PR #641, opened 2021-04-20
changed or of to for in in serialize

dudilugasi:
I added a function to the Array prototype in my app. when running the function createNamespacedJob i'm getting an error that says : 
`"Job.batch \"...\" is invalid: [spec.template.spec.containers[0].ports[1].containerPort: Required value, spec.template.spec.containers[1].name: Required value, spec.template.spec.containers[1].image: Required value, spec.template.spec.tolerations[1].operator: Invalid value: \"\": operator must be Exists when `key` is empty, which means \"match all values and all keys\"]",`

after researching your code i found out that in the serialize function in the file [models.ts](https://github.com/kubernetes-client/javascript/blob/0403427cada5cab3ee7e1e01002cebe4738ae9fb/src/gen/model/models.ts) theres a for...in loop that running on array instead of a for...of loop resulting in the loop reading the function as an element in the array and thus creating a new empty element. this is the json retruned from the serialize
{
  "apiVersion": "batch/v1",
  "kind": "Job",
  "metadata": {
    "labels": {
      "app": "...."
    },
    "name": "...."
  },
  "spec": {
    "backoffLimit": 0,
    "parallelism": 1,
    "template": {
      "spec": {
        "containers": [
          {
            "image": "...",
            "imagePullPolicy": "...",
            "name": "...",
            "ports": [
              {
                "containerPort": ...,
                "name": "example",
                "protocol": "TCP"
              },
              {
                "name": "" <--- this is the problematic area
              }
            ],
            "resources": {
              "limits": {
                "cpu": "1.0"
              },
              "requests": {
                "cpu": "0.5"
              }
            }
          },
          {
            "name": "" <--- this is the problematic area
          }
        ],
       
        "tolerations": [
          {
            "effect": "...",
            "key": "...",
            "operator": "...",
            "value": "..."
          },
          {} <--- this is the problematic area
        ]
      }
    }
  }
}

### DIFF
--- a/src/gen/model/models.ts
+++ b/src/gen/model/models.ts
@@ -1973,8 +1973,7 @@ export class ObjectSerializer {
             let subType: string = type.replace("Array<", ""); // Array<Type> => Type>
             subType = subType.substring(0, subType.length - 1); // Type> => Type
             let transformedData: any[] = [];
-            for (let index in data) {
-                let date = data[index];
+            for (let date of data) {
                 transformedData.push(ObjectSerializer.serialize(date, subType));
             }
             return transformedData;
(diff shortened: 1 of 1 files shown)


brendandburns:
This code is generated, so we don't change it here. If it is going to change it needs to change in the upstream generator (open-api-generator project)

But, I don't think we want to make this change anyway, see #643

PR #647, opened 2021-04-29
Remove node 10, it has reached EOL, add node 16 to get us ready for the next LTS.

brendandburns:
cc @drubin


drubin:
/lgtm 
/Approve 

Thanks for getting to this.

PR #651, opened 2021-05-06
bump execa to v5.0.0

TattdCodeMonkey:
execa added typescript types in v2, but the latest is v5. There were a
few type errors to fix once valid type checks were available. This
PR updates to the latest version and fixes the type errors.


brendandburns:
Thanks for the PR. We need you to sign the CLA.

TattdCodeMonkey:
@brendanburns just saw that, went through and signed. thanks

PR #657, opened 2021-05-14
Regenerate for Kubernetes 1.21.1

brendandburns:


drubin:
@brendandburns  can you rebase to remove the odd double version bump/downgrade?

settings
```diff
@@ -21,7 +21,7 @@ export GEN_COMMIT=d71ff1efd
 export USERNAME=kubernetes
 
 # Kubernetes branch/tag to get the OpenAPI spec from.
-export KUBERNETES_BRANCH="v1.20.2"
+export KUBERNETES_BRANCH="v1.21.1"
```
drubin, line 24: Do we need a changelog update?

drubin:
@brendandburns the commits still seemed odd there are 2 version bumps in this MR and the duplicate MR to version bump? which is correct?

drubin:
I approved the other MR first, but this also looks good, not sure when you want to do the release though so I will leave you to add the `/lgtm` flag if you are happy else just retag me.

brendandburns:
I think the policy is to keep the version in `package[-lock].json` as the next release version. So this is fine to merge, and then eventually I will cut a release.

(still figuring out the right way to do versioning w/ npm...)

PR #661, opened 2021-05-28
Changed Content-Type header on method patchNamespacedDeployment

fabioph:
I was trying to update deployments with this method, but it always returned error code 415 - Unsupported Media Type. 
So, to make it work, it was needed  to change the Content-Type header to "application/merge-patch+json".

### DIFF
--- a/src/gen/api/appsV1Api.ts
+++ b/src/gen/api/appsV1Api.ts
@@ -3326,6 +3326,10 @@ export class AppsV1Api {
             .replace('{' + 'namespace' + '}', encodeURIComponent(String(namespace)));
         let localVarQueryParameters: any = {};
         let localVarHeaderParams: any = (<any>Object).assign({}, this._defaultHeaders);
+        
+        //Change Content-Type header because the K8s endpoint needs it
+        localVarHeaderParams["Content-Type"] = 'application/merge-patch+json';
+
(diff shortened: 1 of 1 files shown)


brendandburns:
This is not necessary, please see the example here:

https://github.com/kubernetes-client/javascript/blob/master/examples/patch-example.js

Which shows how to pass in the header.

Also, this code is generated, so we can't hand edit it, as changes will be over-written next time we generate.

Also, there are three different patch content-types that Kubernetes supports, so it needs to be a parameter, not hard coded.

Closing this PR.

fabioph:
Hello,Thank you for your explanation.I was just trying to help, the documentation doesn’t explain a lot of things, at least they are not easy to find.  From: Brendan BurnsSent: Saturday, May 29, 2021 3:42 PMTo: kubernetes-client/javascriptCc: Fábio Henriques; MentionSubject: Re: [kubernetes-client/javascript] Changed Content-Type header on method patchNamespacedDeployment (#661) This is not necessary, please see the example here:https://github.com/kubernetes-client/javascript/blob/master/examples/patch-example.jsWhich shows how to pass in the header.Also, this code is generated, so we can't hand edit it, as changes will be over-written next time we generate.Also, there are three different patch content-types that Kubernetes supports, so it needs to be a parameter, not hard coded.Closing this PR.—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #688, opened 2021-06-30
Upgrade js-yaml to 4.1.0

### LINKED ISSUE #638: Function yaml.safeLoad is removed in js-yaml 4. Use yaml.load instead, which is now safe by default
when using the library with js-yaml 4 , i'm getting this warning. Not sure if you want to upgrade one time to js-yaml 4, but you will get this error. It4s being used in 2 places

```
Function yaml.safeLoad is removed in js-yaml 4. Use yaml.load instead, which is now safe by default
```

ps i got this error using webpack

brendandburns:
Closes #638

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.15.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@types/js-yaml": "^3.12.1",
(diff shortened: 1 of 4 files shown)


huineng:
I guess this is an easy fix ?

PR #692, opened 2021-07-07
Set current context namespace in loadFromCluster()

briceruzand:
I use k8s js client within a pod and current context namespace is not set properly.

I read current from namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`

[Doc: Accessing the API from a Pod](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -224,7 +224,7 @@ export class KubeConfig {
         const namespaceFile = `${pathPrefix}${Config.SERVICEACCOUNT_NAMESPACE_PATH}`;
         let namespace: string | undefined;
         if (fileExists(namespaceFile)) {
-            namespace = fs.readFileSync(namespaceFile, 'utf8')
+            namespace = fs.readFileSync(namespaceFile, 'utf8');
         }
         this.contexts = [
             {
--- a/src/config.ts
+++ b/src/config.ts
@@ -221,11 +221,17 @@ export class KubeConfig {
                 },
             },
         ];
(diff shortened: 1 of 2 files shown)


briceruzand:
I signed it

src/config.ts
```diff
@@ -221,11 +221,17 @@ export class KubeConfig {
...
+        let namespace: string | undefined;
+        if (fileExists(namespaceFile)) {
+            namespace = fs.readFileSync(namespaceFile, 'utf8')
+        }
         this.contexts = [
             {
                 cluster: clusterName,
                 name: contextName,
                 user: userName,
+                namespace,
```
brendandburns, line 234: Shouldn't this be `namespace: namespace`?

brendandburns: oh, right typescript just repeats it for you, nevermind...

src/config.ts
```diff
@@ -221,11 +221,17 @@ export class KubeConfig {
                 },
             },
         ];
+        const namespaceFile = `${pathPrefix}${Config.SERVICEACCOUNT_NAMESPACE_PATH}`;
+        let namespace: string | undefined;
+        if (fileExists(namespaceFile)) {
+            namespace = fs.readFileSync(namespaceFile, 'utf8')
```
brendandburns, line 227: missing a semi-colon.

brendandburns:
There's one style violation that you need to address, then I think this is good to merge.

src/config.ts
briceruzand: Sorry I miss the notification.
I just push it.
Thx for the review

Issue #700, opened 2021-07-19
[feature request] Support GOOGLE_APPLICATION_CREDENTIALS


mansona:
In https://github.com/kubernetes-client/javascript/issues/623#issuecomment-810653868 you mention that `The current GCP authentication doesn't support that environment variable (GOOGLE_APPLICATION_CREDENTIALS)` and that you wouldn't mind a PR. 

I'm opening this PR to keep track of the feature request and maybe to open up the discussion on how to achieve this. I'm happy to take a crack at it but I really don't have a clue where to start and would love a few pointers 👍

brendandburns:
The current GCP auth is here:

https://github.com/kubernetes-client/javascript/blob/master/src/cloud_auth.ts#L21

The right thing to do is probably to break GCP out into it's own file (that authenticator supports both Azure and GCP currently) and create a `gcp_auth.ts` update the registrations, etc.

Then you will need to figure out the right implementation of the authenticator interface.

mansona:
So I did a minimal amount of work on this yesterday just to check to see if you were happy with the direction. I noticed that there weren't any tests for the cloudauth class so I copied some of the basic tests from `oicd_auth_test.ts` to at least get this process started.

I could do with some guidance on the testing front, even just giving me some pointers on things to test. I am not all that familiar with k8s authentication methods so I don't know if the placeholder tests that I copied from `oicd_auth_test.ts` apply to either GCP or Azure 🤔

brendandburns:
There are a bunch of tests for the cloud provider in `config_test.ts` for historic reasons
e.g. https://github.com/kubernetes-client/javascript/blob/master/src/config_test.ts#L622

If you split out all of the tests with `'azure'` in them into `gcp/azue_test.ts` and take them out of `config_test.ts` that's probably the best bet.

Thanks!

PR #702, opened 2021-07-21
split cloud_auth into gcp and azure

mansona:
This PR is to track work on https://github.com/kubernetes-client/javascript/issues/700

This is probably going to be **very** WIP for a long time because I don't really know what I'm doing with this repo 😂  Please be as harsh as possible and tell me what I need to change to move forward 🙇


mansona:
I've signed the CLA

brendandburns:
@mansona this looks totally reasonable. Please use the tests from `config_test.ts` instead of the ones from oidc and I think that this is good to merge.

mansona:
@brendanburns I pulled the config tests that you mentioned but I only applied them to the azure file since they were all azure. Do you want me to duplicate them for gcp too?

brendanburns:
Probably worth duplicating them since we're cloning the code. It will help
prevent regressions.

The only reason that they're only azure is because most of the code was
shared.

mansona:
I have added the tests for GCP now too 👍  

Would you mind approving the workflows on this? I had a strange test failure on my local machine that I can't explain and I want to know if it's something in my environment or if it's something that I've broken 🙈

brendandburns:
Looks like ci/cd is failing....

mansona:
that's actually perfect! it's failing with the same error that I got locally so I know it's not my environment 🎉 

I'll take a look and get this fixed in a bit 👍

mansona:
I found the issue 🎉  there was a test that depended on the exact order and length of the authenticators in config instead of searching for the right one. 

I fixed that: https://github.com/kubernetes-client/javascript/pull/702/files#diff-f7d16ccc3d41b45b1511b3e65724c67009d49174a7a290c4034487e89e13141dL1001-R752

brendandburns:
/lgtm
/approve

Thanks!

PR #703, opened 2021-07-24
Replace callback with promise in Log

Legion2:
This replaces the done callback in the `Log.log` function with a promise, which is awaited as part of the log function. This ensures that when the promise of the log function resolves a response **was** received (success or error) from kubernetes. If the response is a 200, the body of the response is streamed into the provided stream. Else the promise is rejected with an error or the complete parsed status response as `HttpError`.

This is a breaking change as the `done` callback was removed from the function. Also the behavior of the returned Promise changed. Before, the Promise was resolved when the request was send, now it resolved after the response headers are received or an error occurred.

I also updated the `follow-logs.js` example.

### DIFF
--- a/examples/follow-logs.js
+++ b/examples/follow-logs.js
@@ -13,11 +13,11 @@ logStream.on('data', (chunk) => {
 	process.stdout.write(chunk);
 });
 
-log.log('default', 'pod1', 'container1', logStream, (err) => {console.log(err)}, {follow: true, tailLines: 50, pretty: false, timestamps: false})
+log.log('default', 'pod1', 'container1', logStream, {follow: true, tailLines: 50, pretty: false, timestamps: false})
+.catch(err => {console.log(err)})
 .then(req => {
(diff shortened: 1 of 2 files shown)


Legion2:
I signed it

brendandburns:
Thanks for the PR, I don't think we should break the contract for existing library users, can you use typescript overloading to retain both function signatures?

Thanks

Legion2:
@brendandburns I thinks its a better idea to make it a separate function, because the semantic of the returned promise changed. The current implementation is broken in sense that it can not be used to handle errors properly while streaming the response. It mixes the different events into the promise and done callback. So I consider my changes a fix for the inconsistent handling of callbacks and promises of the log function.

Legion2:
@brendandburns I added the typescript overload and deprecated the signature with the done callback. However, the fixed promise resolve behavior is also included in the old signature.

brendandburns:
Thanks!

/lgtm
/approve

PR #709, opened 2021-08-16
Change 'port' in v1HTTPGetAction to reflect proper type options

danthegoodman1:
`Object` is not the correct type, and makes typescript usage annoying since you need to do something like: `port: 'http' as Object` to allow it to work


brendandburns:
Unfortunately, this code is auto-generated, so we can't do manual changes like this. There is a fix needed in either the code generator or the Kubernetes swagger specification.

brendandburns:
https://github.com/kubernetes-client/javascript/issues/349

PR #710, opened 2021-08-16
Update v1ServicePort['targetPort'] to correct type

danthegoodman1:
`object` is not the correct type, it should be of `number` just like `port` is.


brendandburns:
This code is auto-generated, we can't make changes to it. We need to update the code generator or the swagger specification.

See https://github.com/kubernetes-client/javascript/issues/349

PR #726, opened 2021-09-30
Fix create empty array items in serializer

### LINKED ISSUE #725: Client change data to invalid before post
`.createNamespacedJob` autoadd two objects `{name:''}` to spec.containers array, and throw error `spec.template.spec.containers[2].name: Required value`

```javascript
const batchV1Api = k8s.makeApiClient(k8s.BatchV1Api);

const newJob = await batchV1Api.createNamespacedJob(this.namespace, {
			kind: 'Job',
			apiVersion: 'batch/v1',
			metadata: {
				name: 'q7-tasker-migrations',
				annotations: {
					'cronjob.kubernetes.io/instantiate': 'manual',
				},
			},
			spec: {
				backoffLimit: 1, 
				template: {
					spec: {
						restartPolicy: 'Never',
						containers: [ { name:''q7-tasker-migrations'', image: dockerimage } ],
					},
				},
			},
		}).catch(err=>{
			console.log(err.response?.body?.details?.causes);
		});

```
this code send to server 
```
 '{"apiVersion":"batch/v1","kind":"Job","metadata":{"annotations":{"cronjob.kubernetes.io/instantiate":"manual"},"name":"q7-tasker-migrations"},"spec":{"backoffLimit":1,"completions":1,"parallelism":1,"template":{"spec":{
 "containers":[
  {"image":"***/services/tasker:1.2.1-init", "name":"q7-tasker-migrations"},
  {"name":""},
  {"name":""}
],
"restartPolicy":"Never"}}}}',
```

returns error 
```js[
{
  reason: 'FieldValueRequired',
  message: 'Required value',
  field: 'spec.template.spec.containers[1].name'
},
{
  reason: 'FieldValueRequired',
  message: 'Required value',
  field: 'spec.template.spec.containers[1].image'
},
{
  reason: 'FieldValueRequired',
  message: 'Required value',
  field: 'spec.template.spec.containers[2].name'
},
{
  reason: 'FieldValueDuplicate',
  message: 'Duplicate value: ""',
  field: 'spec.template.spec.containers[2].name'
},
{
  reason: 'FieldValueRequired',
  message: 'Required value',
  field: 'spec.template.spec.containers[2].image'
}]
```

I think problem in ObjectSerializer.serialize function ... 

**Environment:**
 - OS: Linux
 - NodeJS :14
 - @kubernetes/client-node : "^0.15.0"

alexstep:
Close #725 

replace `for` to `.map`

### DIFF
--- a/src/gen/model/models.ts
+++ b/src/gen/model/models.ts
@@ -1966,12 +1966,7 @@ export class ObjectSerializer {
         } else if (type.lastIndexOf("Array<", 0) === 0) { // string.startsWith pre es6
             let subType: string = type.replace("Array<", ""); // Array<Type> => Type>
             subType = subType.substring(0, subType.length - 1); // Type> => Type
-            let transformedData: any[] = [];
-            for (let index in data) {
-                let date = data[index];
-                transformedData.push(ObjectSerializer.serialize(date, subType));
-            }
-            return transformedData;
+            return data.map(date=>ObjectSerializer.serialize(date, subType))
         } else if (type === "Date") {
             return data.toISOString();
         } else {
--- a/src/gen/model/models.ts
+++ b/src/gen/model/models.ts
@@ -1964,8 +1964,8 @@ export class ObjectSerializer {
         } else if (primitives.indexOf(type.toLowerCase()) !== -1) {
             return data;
         } else if (type.lastIndexOf("Array<", 0) === 0) { // string.startsWith pre es6
-            let subType: string = type.replace("Array<", ""); // Array<Type> => Type>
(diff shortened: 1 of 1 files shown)


brendandburns:
Is there an example of where this causes a problem?

Also, this code is generated, so the change will need to be made here, instead:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-node/models.mustache#L116

And then we will vendor it in when we re-generate the code.

brendandburns:
As I mentioned in #725 I can't reproduce this, and since it would require a generator fix to fix anyway, I'm going to close this PR.

PR #727, opened 2021-10-01
Update code generation to Kubernetes 1.22.2, Rev package to 0.16.0

brendandburns:


TattdCodeMonkey:
I tried to test this but ran into an issue, but the issue is something I've run into previously. Sometimes a CRD api gets skipped and it's Api client isn't generated. 

I'll try to set this up with a vanilla kind cluster and test against that, instead of with a cluster with lots of CRDs, then report back.

brendandburns:
Rebased and ready to merge I think.

Thanks!

PR #729, opened 2021-10-06
update kubernetes-client/gen

mstruebing:
We had problems regenerating our javascript client. After digging into it I've found this place where it uses a commit of `kubernetes-client/gen` which is over a year old and I updated it to the most recent commit hash.

I have no clue about all the implications this would have, but it works for us if we use this commit.


mstruebing:
I signed it

mstruebing:
I signed it

PR #737, opened 2021-11-16
chore: update typescript k8s  apply example

### LINKED ISSUE #736: Typescript apply example is out of date
**Describe the bug**
`js-yaml` doesn't support `safeLoadAll` anymore as it is safe by default.

**To Reproduce**
run this code: https://github.com/kubernetes-client/javascript/blob/master/examples/typescript/apply/apply-example.ts

djamaile:
should fix #736


djamaile:
@brendandburns alright everything should be good to go now

Issue #666, opened 2021-06-03
V1ServicePort with wrong type for targetPort


Saeger:
**Describe the bug**
The "targetPort" type is defined as "object" while the possibility of values can be both number and string.
With this said, you can't set a type for a variable and fill targetPort with a number the way it's implemented.
"Type 'number' is not assignable to type 'object'.ts(2322)"

** Client Version **
`0.14.3`

** Server Version **
`1.18.0`

**To Reproduce**
Define a variable of type V1Service and try to make the specs for ports.

**Expected behavior**
The type definition could be defined as number | string since both are supported.

** Example Code**
`      ports: [
        {
          name: 'http',
          port: 80,
          targetPort: 3000,
          protocol: 'TCP',
        },
      ],`

**Environment (please complete the following information):**
 - OS: OSX
 - NodeJS Version 14

brendandburns:
Officially, the targetPort is an IntOrString so we need to implement special handlingwhich currently doesn't exist.

Saeger:
@brendanburns where should I start looking and see if I could help with a fix ?

brendandburns:
@Saeger thanks for offering to help.

We haven't done this for Javascript before, so it's likely to be a little rocky, but here's what (I think) you need to do.

We need to add `typeMappings` and `importMappings`, similar to this one in the Java client:
https://github.com/kubernetes-client/gen/blob/master/openapi/java.xml#L96

To the Javascript client config:
https://github.com/kubernetes-client/gen/blob/master/openapi/typescript.xml

Then you need to regenerate the client:
`npm run generate`

This code should not compile because it will be missing the `IntOrString` type, so then you'll need to add `custom.ts` (or somesuch) and define the `IntOrString` class in that file.

As I mentioned, complicated, please feel free to ping this issue if you try and see errors!

alex-bezek:
@brendandburns thanks for pointing me to this issue, I've created a couple of pull requests which I think address the issue:
* https://github.com/kubernetes-client/javascript/pull/741
* https://github.com/kubernetes-client/gen/pull/213

brendandburns:
I have just pushed an `0.16.1` release to npm that contains a fix for IntOrString types. Many thanks to @alex-bezek for sending the PRs.

PR #741, opened 2021-11-18
Update typescript open API generator to handle the IntOrString type

alex-bezek:
Resolves https://github.com/kubernetes-client/javascript/issues/666

The 1 main change here is the addition of the `IntOrString` type to the `src/types.ts` file. The rest is the result of running
`GEN_ROOT=/home/alexbezek/code/gen npm run generate` and running `npm run build` to verify it compiles

Depends on https://github.com/kubernetes-client/gen/pull/213

### DIFF
--- a/src/gen/model/v1HTTPGetAction.ts
+++ b/src/gen/model/v1HTTPGetAction.ts
@@ -11,6 +11,7 @@
  */
 
 import { RequestFile } from './models';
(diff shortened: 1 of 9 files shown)


brendandburns:
/lgtm
/approve

Thank you for taking the time to do this!!

PR #743, opened 2021-11-22
Ensure copied files are finished copying

gingermusketeer:
Why: Otherwise errors such as "Error: TAR_BAD_ARCHIVE: Unrecognized archive format" come up.


gingermusketeer:
Didn't have much luck writing tests for this so would happily take some pointers

brendandburns:
Thanks for the PR.

We need you to sign the CLA.

The tests that exist for copy are here:

https://github.com/kubernetes-client/javascript/blob/master/src/cp_test.ts

PR #753, opened 2021-12-08
Fix expires-on case on Azure auth

laudibert:
Fix for #752

PR #756, opened 2021-12-17
Package publish workflow

laudibert:
This is a proposition to add automatic package publish on `master` as well as a publish dry run on PRs.

I modified the validation workflow to add a follow-up job on it that will publish the package. It will fail if the version given has already been published. It will also only do a dry un on PRs.

The job requires `NPM_TOKEN` to be [added as a secret on this repository](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository).

I am aware that this may be up to debate, will be happy to do so.


brendandburns:
I'm not sure that I'm in favor of automatic publishing of npms on push. I think publishing should be an intentional act.

I guess you could argue that rev-ing the version in `package.json` is intentional, but I like an explicit action to be taken.

Regardless, it also doesn't match up with the existing way that we use the version in `package.json` which is that the checked in version is always the _next_ version, so this would end up with us publishing a new version before we were ready.

Additionally, we need to run a build (e.g. compile typescript to javascript) before we publish, I don't think your current workflow does this.

If you are open to it, can you refactor this as follows:

1) Don't delete test.yaml, leave it as-is, you can add the dry-run publish here if you want, though I'm not sure it adds much.
2) Add `ci.yaml` but set it as a `workflow_dispatch` so that it can be manually triggered.
3) [optional] figure out the right way to have a snapshot library on npm. I'm not sure the right approach, we could use semantic versioning (e.g. `alpha1`).

laudibert:
Yes ok, understood. Yes I think that step 3 is important. I honestly got motivated to do this PR because I am waiting for my merged PR to be published on a version that I can use. Therefore I do think it is important to have some way to give a alpha/release candidate/snapshot version.

I'll look into it in the next few days.

brendandburns:
In the meantime, would it be useful for me to cut a new version to pick up your PR? (which PR is it)

laudibert:
That would be very nice, thank you. It's PR #753

Issue #653, opened 2021-05-10
quantityToScalar does not support all of the SI suffixes


zephyrxvxx7:
**Describe the bug**
In the file `src/utils.ts`, the function `quantityToScalar` does not support all of the Kubernetes canonical quantity.
It just supports "m" and "Ki" SI suffixes. 

**Client Version**
`0.14.3`

**Server Version**
`1.19.3`

**Expected behavior**
Support all of the Kubernetes canonical quantity:
base1024: Ki | Mi | Gi | Ti | Pi | Ei
base1000: n | u | m | "" | k | M | G | T | P | E

**Environment (please complete the following information):**
 - OS: `macOS`
 - NodeJS Version `v16.0.0`

**Additional context**
The official repo [apimachinery](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go) have more detail about it.

brendandburns:
We'd be happy to take a PR to expand that functionality.

yash-saini275:
I was reading [Managing Resources for Containers](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu) and found that Kubernetes doesn't use **n(nano)** and **u(micro)**. So I don't think there is a need to include it in `quantityToScalar` function.

yash-saini275:
Also `const num = parseInt(quantity, 10);
    if (isNaN(num)) {
        throw new Error('Unknown quantity ' + quantity);
    }`

I doubt if this logic is correct or not. Because if we pass a string starting with the alphabetic character to `parseInt` function it returns NaN but if we pass a string starting with the number it returns the number even though the suffix contains random characters.
Example:- 
`parseInt("100asdf") = 100`
`parseInt("asdf1000asdf") = NaN`

yash-saini275:
Hey guys,
I was working on the test cases for this and found that we are using the previous version of typescript which doesn't support ```BigInt```. And the number for ```Ei```, which is ```2^60=1152921504606846976```, But JS Math.pow function returns ```1152921504606847000``` is going beyond the safe limit for JS. 
What should be the approach to solve this?

brendandburns:
This library is using TypeScript 4.1.3 and BigInt was introduced in TS 3.2, so I think we should be able to use BigInt, right?

yash-saini275:
It should not cause the problem but the tsconfig.json targets es6 which is lower than es2020. This is causing the error.

[...]

PR #757, opened 2021-12-28
Add support for more quantity units

wejendorp:
Adds support for missing decimal units: u, k, M, G, T, P and E.
Fixes multiplier for nano from 1e-8 to 1e-9.

Adds more units from #653, and should cover the full range of decimal units now.
Prevents the new `top` commands from crashing if any resource is using decimal units.

### DIFF
--- a/src/top_test.ts
+++ b/src/top_test.ts
@@ -29,7 +29,7 @@ const mockedPodMetrics: PodMetricsList = {
             },
             timestamp: '2021-09-26T11:57:21Z',
             window: '30s',
-            containers: [{ name: 'nginx', usage: { cpu: '5000000n', memory: '3912Ki' } }],
+            containers: [{ name: 'nginx', usage: { cpu: '50000000n', memory: '3912Ki' } }],
         },
         {
             metadata: {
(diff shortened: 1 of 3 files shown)


src/top_test.ts
```diff
@@ -42,7 +42,7 @@ const mockedPodMetrics: PodMetricsList = {
             window: '30s',
             containers: [
                 { name: 'nginx', usage: { cpu: '0', memory: '4012Ki' } },
-                { name: 'sidecar', usage: { cpu: '140000000n', memory: '3012Ki' } },
+                { name: 'sidecar', usage: { cpu: '1400000000n', memory: '3012Ki' } },
```
wejendorp, line 45: Adding missing 0 to make tests pass w/ the nano 1e-8 -> 1e-9 change.

wejendorp:
@brendandburns when do you think it'll be possible to make a release from this?

sorenmat:
Please note I created this https://github.com/kubernetes-client/javascript/issues/758 since merging to master triggers a job that doesn't work

brendandburns:
@wejendorp I will push a release this week.

wejendorp:
> @wejendorp I will push a release this week.

@brendandburns bump, in case you forgot.

mycrEEpy:
A release would be super helpful as we are currently hitting the "Unknown suffix" exception.

Issue #769, opened 2022-01-28
Vulnerability from tar


Tenaria:
**Describe the bug**
Security Vulnerability in package imported by kubernetes-client. The vulnerable package is `tar 6.0.2`. To fix, upgrade to `6.1.9` or higher.

Vuln details at: https://snyk.io/vuln/npm:tar

** Client Version **
`0.15.0`

Tenaria:
Fixed in PR above.

PR #767, opened 2022-01-18
package.json: Add override for colors@1.40.0

justaugustus:
ref:
- https://github.com/advisories/GHSA-5rqg-jm4f-cqx7
- https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides
- https://twitter.com/bitandbang/status/1478786515660648448
- https://twitter.com/bitandbang/status/1480605181792370693

Signed-off-by: Stephen Augustus <foo@auggie.dev>

PR #770, opened 2022-01-28
upgrades tar version from 6.0.2 to 6.1.11

Tenaria:
Upgrade tar version to 6.1.11 and run `npm install` to resolve security vulnerability. This is a minor bump and should not have breaking changes. `npm test` passes.

ticket: https://github.com/kubernetes-client/javascript/issues/769.


package-lock.json
```diff
@@ -5688,8 +5688,7 @@
       "dev": true
     },
     "colors": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/colors/-/colors-1.4.0.tgz",
+      "version": "https://registry.npmjs.org/colors/-/colors-1.4.0.tgz",
```
brendandburns, line 5691: Any idea why the colors version is changing too?

package-lock.json
```diff
@@ -8195,7 +8194,7 @@
       "integrity": "sha512-ZQZnjJPrt0rjp216gp6FQC1QC4ojcoKikhfOJ/51CqaJunVDilRLlIO5tCGWj1tzlYYT9eOGhJv7MF3t7rxSmw==",
       "dev": true,
       "requires": {
-        "colors": "^1.4.0",
```
brendandburns, line 8198: This seems unrelated.

brendandburns:
This looks fine, but there appears to be an unrelated change to the colors dependency.

Tenaria:
@brendandburns I believe there was a typo introduced by this PR https://github.com/kubernetes-client/javascript/pull/767/files (see the first link on the PR description which recommends to downgrade colors to 1.4.0 but we pinned it to 1.40.0 which doesn't exist https://www.npmjs.com/package/colors). This is probably the root cause. I've pushed a fix.

package-lock.json
Tenaria: (see comment in PR)

Tenaria:
Thanks! @brendandburns 🙇

Tenaria:
How often / how do you determine when to create a release btw? @brendandburns

Just wondering when I can expect this change to officially get released. Thanks!

PR #772, opened 2022-01-31
Add list method to generic KubernetesObjectApi

schrodit:
This PR implements a list method on the generic `KubernetesObjectApi`.
I created a new PR for that feature as the original PR (https://github.com/kubernetes-client/javascript/pull/588), that wanted to implement this feature, has been closed due to inactivity.

In addition the list method supports:
- an optional namespace selector
- fieldselector
- labelselector
- limit + continue token

### DIFF
--- a/src/object_test.ts
+++ b/src/object_test.ts
@@ -1747,6 +1747,82 @@ describe('KubernetesObject', () => {
             await client.delete(s, undefined, undefined, 7, undefined, 'Foreground');
             scope.done();
         });
+
+        it('should list resources in a namespace', async () => {
+            const scope = nock('https://d.i.y')
+                .get(
+                    '/api/v1/namespaces/default/secrets?fieldSelector=metadata.name%3Dtest-secret1&labelSelector=app%3Dmy-app&limit=5',
+                )
+                .reply(200, {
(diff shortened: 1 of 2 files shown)


src/object_test.ts
```diff
@@ -1921,5 +1921,31 @@ describe('KubernetesObject', () => {
...
+            let thrown = false;
+            try {
+                await (client.list as any)('', undefined);
+                expect.fail('should have thrown an error');
+            } catch (e) {
+                thrown = true;
+                expect(e.message).to.contain('Required parameter kind was null or undefined when calling ');
+            }
+            expect(thrown).to.be.true;
+        });
```
brendandburns, line 1949: Can you add a test that validates that you correctly construct the URL for the GET then I think this is ready to merge.

brendandburns:
Thanks for the PR! One additional unit test requested then we can merge this.

src/object_test.ts
schrodit: Added tests for namespaced and non namespaced list requests.
And also added support for the continue feature

brendandburns:
/lgtm
/approve

Thanks!

Issue #754, opened 2021-12-09
Proposal to swap deprecated request for node-fetch


OliverMKing:
# Proposal to swap deprecated request for node-fetch

Request is fully deprecated requiring us to switch libraries (see #414 for more information).

@brendanburns outlined some of the clear options [forward](https://github.com/kubernetes-client/javascript/issues/414#issuecomment-978031677). This change should ideally be a long-term solution leaving us with the option of either modifying the node-typescript open-api-generator to use a different request library or migrate to one of the other generator options.

Modifying an outdated node-typescript library does not seem like the right path forward. It would require us to still use a new library which would likely change the api and still be a breaking change. We might as well migrate to a newer version of openapi-generator while we do that and acquire the advantages of a more up-to-date version.

## Choosing a new library

@d-cs posted a nice [summary](https://github.com/kubernetes-client/javascript/issues/414#issuecomment-648847415) of the different choices. We narrowed it down to Axios or Fetch. The other libraries didn't match how widely adopted and focused these libraries are.

Fetch is simply the standard JavaScript as a whole is moving to, which brings more longterm support aligning us better with our goal of migrating to a long-term solution. [node-fetch](https://www.npmjs.com/package/node-fetch) has significantly more weekly downloads than [axios](https://github.com/axios/axios). The package size of node-fetch is also noticeably smaller than axios.

Additionally, future improvements could likely be done to allow us to also support the browser #165 since Fetch is native to the browser.

## Node-fetch

Fetch is a native browser API that is not implemented in Node.js by default. Node-fetch is the natural choice for the Node implementation because it is the most widely adopted.

The openapi-generator uses the browser implementation of Fetch so we must substitute node-fetch in manually.

## Implementation steps

@davidgamero and I will work to implement the following changes.

### Other repositories

-   [x] Update [kubernetes-client/gen](https://github.com/kubernetes-client/gen)'s typescript-fetch files to let us pass in the `typescriptThreePlus` config option <sup>[1](https://github.com/OpenAPITools/openapi-generator/issues/9973) [2](https://github.com/OpenAPITools/openapi-generator/issues/3869#issuecomment-584152932)</sub>
-   [x] Update [openapi-generator](https://github.com/OpenAPITools/openapi-generator)'s typescript-fetch flavor to mark parameters as optional if all parameters are optional <sup>[3](https://github.com/OpenAPITools/openapi-generator/issues/6440)</sup>

### Kubernetes-client repository

[...]

PR #773, opened 2022-02-03
generated code migration from request to node-fetch/isomorphic-fetch

davidgamero:
migration of OpenAPIGenerator code to current typescript generator that no longer uses `request`
in reference to #754

Note: this will leave the master branch in a development state as the migration doesn't cover the /src/*.ts files that still use request.
Merging this would allow the work of updating all the non-generated code to be parallelized on master.

Notable remaining migration work:
-watch.ts, log.ts, and metrics.ts all directly implement `request` without using the generated apiclient
-config.ts needs to be refactored from the `applyToRequest()` method to the new `applySecurityAuthentication()` method now that the generated code uses a `SecurityAuthentication` interface for creating a `RequestContext` instead of directly passing `https.Options`


brendandburns:
Thanks!

Can you split this PR into two commits: 1) generated code changes. 2) any hand-edited changes.

It's basically impossible to review the generated code because there are so many files, but I would like to review any hand-edited changes.

Also you need to sign the CLA. Just do the EasyCLA one b/c there is migration happening to that today-ish.

davidgamero:
The only hand-edited change in this PR is the GEN_COMMIT hash, and I will be making a separate PR for the hand-edited changes after this one for converting the `config` to the new interfaces.
Should I do the GEN_COMMIT hash change in a separate commit from the code generated by altering it for clarity?

Sounds good I'll do the CLA

brendandburns:
@davidgamero I created a `release-1.x` repo. Let's move this PR to target that repo and we can merge and iterate.

davidgamero:
perfect. just retargeted the PR

davidgamero:
looks like the PR might need a rebase- I could just open a new one at this point

PR #780, opened 2022-03-01
fix(cache): update cache with O(1) data structures

bbatha:
The prior implementation used arrays to store cached objects. This meant
that updates were O(n). In a controller I developed that managed ~70k PV
and ~70k pvc 99.9% of CPU time was spent in cache updates pegging the
entire process. This new implementation doesn't even have cache updates
show up in the profiles and is using ~25m cpu for the same number of
objects.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -20,19 +20,23 @@ export interface ObjectCache<T> {
     list(namespace?: string): ReadonlyArray<T>;
 }
 
+// exported for testing
+export type CacheMap<T extends KubernetesObject> = Map<string, Map<string, T>>;
+
 export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, Informer<T> {
-    private objects: T[] = [];
+    private objects: CacheMap<T> = new Map();
     private resourceVersion: string;
-    private readonly indexCache: { [key: string]: T[] } = {};
-    private readonly callbackCache: { [key: string]: Array<ObjectCallback<T> | ErrorCallback> } = {};
+    private readonly callbackCache: {
+        [key: string]: Array<ObjectCallback<T> | ErrorCallback>;
+    } = {};
     private request: RequestResult | undefined;
-    private stopped: boolean = false;
+    private stopped = false;
 
     public constructor(
         private readonly path: string,
         private readonly watch: Watch,
         private readonly listFn: ListPromise<T>,
-        autoStart: boolean = true,
+        autoStart = true,
         private readonly labelSelector?: string,
     ) {
         this.callbackCache[ADD] = [];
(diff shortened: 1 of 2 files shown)


src/cache_test.ts
```diff
@@ -143,11 +144,11 @@ describe('ListWatchCache', () => {
...
         expect(cache.list()).to.deep.equal(list);
 
-        expect(cache.get('name1')).to.equal(list[0]);
-        expect(cache.get('name2')).to.equal(list[1]);
+        expect(cache.get('name1', 'default')).to.equal(list[0]);
+        expect(cache.get('name2', 'default')).to.equal(list[1]);
 
         expect(cache.list('default')).to.deep.equal(list);
-        expect(cache.list('non-existent')).to.be.undefined;
+        expect(cache.list('non-existent')).to.deep.equal([]);
```
bbatha, line 151: The test and the type of the `list` return disagreed, I went with returning an empty array if its not found. I can leave the diagreement in place or update the types an undefined is preferred to be returned.

src/cache_test.ts
```diff
@@ -143,11 +143,11 @@ describe('ListWatchCache', () => {
         expect(pathOut).to.equal('/some/path');
         expect(cache.list()).to.deep.equal(list);
 
-        expect(cache.get('name1')).to.equal(list[0]);
-        expect(cache.get('name2')).to.equal(list[1]);
+        expect(cache.get('name1', 'default')).to.equal(list[0]);
+        expect(cache.get('name2', 'default')).to.equal(list[1]);
```
bbatha, line 147: I changed the test here to take the `default` namespace. The old logic didn't make much sense and it seemed like it was supposed to return `undefined` if the name and namespace did not match. From a consumer perspective I would expect the namespace to only be optional if the resource is global like a `namespace` or `pvc`. Let me know I should restore the old behavior of finding the first object with a matching name regardless of what namespace its in if no namespace is provided.

src/cache.ts
```diff
@@ -93,18 +97,25 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
...
+        const nsObjects = this.objects.get(namespace || '');
+        if (nsObjects) {
+            return nsObjects.get(name);
+        }
     }
 
     public list(namespace?: string | undefined): ReadonlyArray<T> {
         if (!namespace) {
-            return this.objects;
+            const allObjects: T[] = [];
```
bbatha, line 108: If it is unacceptable for performance to do this concatting on each call I can modify it to store a secondary array with all of the objects. However, based on using the go client, I'm not sure that I'd expect this `list` to return everything in the cache without a namespace. Like my comment `get` functionality my expectation is that `list` with no namespace would only return objects that don't have namespaces like pvcs and namespaces rather than all objects. But a number of tests rely on the later behavior and this would be a breaking change.

brendandburns: The standard client libraries have a 'listAllResources' function. Given that a catch is tied to a specific Kubernetes type, if I create a watch for a namespaced resource (e.g. a Pod) I would expect that I can get a list of all pods regardless of namespace. This is useful (for example) when writing a scheduler, where I really do want to watch all pods in all namespaces without having to create N different watches for N namespaces (and manage what happens when namespaces come and go)

So I would prefer to retain the pre-existing behavior. From a performance standpoint though, I think it is fine to just concatenate all of the various lists together into a single list.

src/cache_test.ts
brendandburns: In the case of this test, the objects are global, they are `V1Namespace` so I think this test shouldn't change.

src/cache_test.ts
brendandburns: Empty array seems fine.

src/cache.ts
```diff
@@ -93,18 +97,25 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
...
-        return this.objects.find(
-            (obj: T): boolean => {
-                return obj.metadata!.name === name && (!namespace || obj.metadata!.namespace === namespace);
-            },
-        );
+        const nsObjects = this.objects.get(namespace || '');
+        if (nsObjects) {
+            return nsObjects.get(name);
+        }
     }
```
brendandburns, line 104: add an explicit `return undefined` here.

src/cache_test.ts
bbatha: Ah I see this test incorrectly puts `namespace` in the mocked responses so the namespace objects have namespaces. I'll split this into two tests one for namespace objects and one for unnamespaced objects.

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #781, opened 2022-03-02
fix(cache): watch errors must call done handler

bbatha:
The type of `watchObject` was incorrect and has been updated to match
the actual request body.

Using this info it was clear that 'ERROR' events were not being handled
correctly. When the watch receives an error it is not always an http
status code, because the status code can only be sent when the stream is
starting. This means that `410` resourceVersion out of date errors could
only be handled if they were detected before the watch stream started
leaving watches running on channels that would never receive more events
and not notifying `ListWatch` consumers of the error.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -120,7 +120,10 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
 
     private async doneHandler(err: any): Promise<any> {
         this._stop();
-        if (err && err.statusCode === 410) {
+        if (
+            err &&
+            ((err as { statusCode?: number }).statusCode === 410 || (err as { code?: number }).code === 410)
+        ) {
             this.resourceVersion = '';
         } else if (err) {
             this.callbackCache[ERROR].forEach((elt: ErrorCallback) => elt(err));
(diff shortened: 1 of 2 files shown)


brendandburns:
/lgtm
/approve

Thanks for the PR!

Kaggggggga:
hello, I wonder when this fix will be published to next release?
----
some background of my case:
after long running, my operator using v0.16.3 sdk on aws eks 1.21 will stuck into repeatedly watch request call without long polling
according to api server access log, their response is 200, but when i test with the request path, it is showing http200 with 410 error object, and no list api is called
even i implement something like if no event for 15min, will stop and start again, but the present of old resourceVersion will skip  the list api call to refresh resourceVersion

my concluded timeline: 
therefore without this fix, it will 
run for long time => 
until watching initial resourceVersion is too old, the long polling will push a error object without handling => 
15min no new event, watch got restarted without list api call => 
got status 200 but 410 error, request stream closed, recreate watch without list => 
got status 200 but 410 error, request  stream closed, recreate watch without list => .
...........repeatedly one by one when 410 watch request is done, causing frequent api request to api server

you could reproduce by setting a expired initial value of resourceVersion at ListWatch for see my case

PR #786, opened 2022-03-22
Security authentication support and watch migration

davidgamero:
migrate `config.ts` and related code to support the new SecurityAuthentication interface for the generated code that now will rely on fetch instead of fequest. this replaces the old `applyToRequest` interface
also includes a sample migration of `watch.ts` using node-fetch outside the generated code

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -27,7 +27,16 @@ import { ExecAuth } from './exec_auth';
 import { FileAuth } from './file_auth';
 import { GoogleCloudPlatformAuth } from './gcp_auth';
 import { OpenIDConnectAuth } from './oidc_auth';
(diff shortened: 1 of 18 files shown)


src/config.ts
```diff
@@ -138,24 +153,64 @@ export class KubeConfig {
...
-                password: user.password,
-                username: user.username,
-            };
+            httpsOptions.auth = `${user.username}:${user.password}`;
         }
+
+        // Copy headers from httpsOptions to RequestContext
+        const headers = httpsOptions.headers || {};
+        Object.entries(headers).forEach(([key, value]) => {
+            context.setHeaderParam(key, '' + value);
```
brendandburns, line 194: nit: `${value}` instead of '' + value

brendandburns:
One small nit, also please squash and merge.

PR #791, opened 2022-04-01
fix problems with WSL + Windows

### LINKED ISSUE #788: Fails to start on windows with WSL in some cases
Docker Desktop on Windows with WSL2 fails to start (just hangs without any information) due to this client running:

```sh
wsl.exe bash -ic "printenv KUBECONFIG"
```

This happens in [`loadFromDefault()` in config][1].

I'm guessing that this client does this on start, thus the client itself fails to start.

The problem with the command is that bash is ran in interactive mode (the `-i` flag), while it's run as a non-interactive command/shell.
And anything needing interaction in ones `.bashrc` will just wait and hang since there is no way for the user to interact with it.

**Solution**
The `-i` flag to bash should be removed. As the shell is non-interactive, and should not be flagged as interactive.

**Background**
In my case I'm using [`keychain` to automatically add SSH keys to WSL2][2]. This asks for the key passphrase when bash starts interactively. So it (Docker Desktop in my case) just gets stuck there and is unable to continue. (As a workaround for this it's possible to add a `timeout` to the interactive command to kill it after some time, but it shouldn't be needed in the first place.)

**Version**
[`master`][1]

**Environment:**
 - OS: Windows 10, WSL2
 - NodeJS Version: Docker desktop 4.6.1

[1]: https://github.com/kubernetes-client/javascript/blob/master/src/config.ts#L314 
[2]: https://devblogs.microsoft.com/commandline/sharing-ssh-keys-between-windows-and-wsl-2/

brendandburns:
Fixes #788 (and another problem I saw on my WSL machine)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -311,13 +311,9 @@ export class KubeConfig {
         }
         if (process.platform === 'win32' && shelljs.which('wsl.exe')) {
             try {
-                const envKubeconfigPathResult = execa.sync('wsl.exe', ['bash', '-ic', 'printenv KUBECONFIG']);
+                const envKubeconfigPathResult = execa.sync('wsl.exe', ['bash', '-c', 'printenv KUBECONFIG']);
                 if (envKubeconfigPathResult.exitCode === 0 && envKubeconfigPathResult.stdout.length > 0) {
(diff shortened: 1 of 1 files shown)


drubin:
I can't verify this with my current setup but the changes look good
/lgtm

PR #803, opened 2022-05-20
Fix the github action also rev package version to the next version.

brendandburns:


brendandburns:
cc @davidgamero thx!

davidgamero (approved):

davidgamero:
oh i guess i can't use that, but i tried to indicate approval

PR #818, opened 2022-06-13
makeInformer returns Informer & ObjectCache

uesyn:
makeInformer currently returns the ListWatch object implements Informer interface.
In addition that, it also implements ObjectCache interface.

ObjectCache is very useful to implements kubernetes controller, but to use ObjectCache methods, we can't achieve the benefit of typescript type checking, because makeInformer returns the value as Informer interface.
In this PR, makeInformer returns the value as Informer & ObjectCache in order to accurate the typescript type check.
Maybe we should define new interface for that.

### DIFF
--- a/src/informer.ts
+++ b/src/informer.ts
@@ -1,4 +1,4 @@
-import { ListWatch } from './cache';
+import { ListWatch, ObjectCache } from './cache';
 import { KubeConfig } from './config';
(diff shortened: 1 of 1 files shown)


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #820, opened 2022-06-14
Add support for tls-server-name

### LINKED ISSUE #799: The tls-server-name KUBECONFIG cluster flag is not taken into account
**Describe the bug**
The `tls-server-name` flag is not respected by the k8 client even if the SNI was explicitly set like in the following example: 
```yaml
clusters:
- cluster:
    certificate-authority-data: Q0FEQVRB 
    server: http://example.com
    tls-server-name: kube.example.com 
  name: cluster1 
```
Right now the k8 client will use the  `example.com` value for SNI (TLS servername) instead of `kube.example.com`

**Expected behavior**
If the `tls-server-name` flag if present in the KUBECONFIG cluster configuration the `tls-server-name` flag should be properly propagated and set in https request (HTTPSOptions.servername) to the Kubernetes API server.

brendandburns:
Closes #799

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -132,11 +132,16 @@ export class KubeConfig {
 
     public async applytoHTTPSOptions(opts: https.RequestOptions): Promise<void> {
         const user = this.getCurrentUser();
+        const cluster = this.getCurrentCluster();
 
         await this.applyOptions(opts);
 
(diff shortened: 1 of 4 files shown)


davidgamero (approved):
\lgtm

PR #822, opened 2022-06-17
convert log to use fetch streaming

davidgamero:
remove another remaining usage of the request library
ref #754

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -121,11 +121,11 @@ export class Log {
         AddOptionsToSearchParams(options, searchParams);
 
         await this.config.applytoHTTPSOptions(requestOptions);
-        console.log(requestURL.toString())
+
         const req = await fetch(requestURL.toString(), requestOptions)
             .then(response => {
                 const status = response.status;
-                if (status == 200) {
+                if (status === 200) {
                     response.body.pipe(stream) // TODO: the follow search param still has the stream close prematurely based on my testing
                 } else if (status === 500) {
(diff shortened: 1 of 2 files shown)


src/log.ts
```diff
@@ -90,38 +113,33 @@ export class Log {
...
-                        reject(new HttpError(response, body, response.statusCode));
+        const requestOptions: RequestOptions = {};
+
+        const requestURL = new URL(cluster.server + path);
+
+        var searchParams = requestURL.searchParams;
+        AddOptionsToSearchParams(options, searchParams);
+
+        await this.config.applytoHTTPSOptions(requestOptions);
+        console.log(requestURL.toString())
```
brendandburns, line 124: Remove this stray debug code.

src/log.ts
```diff
@@ -90,38 +113,33 @@ export class Log {
...
+        const req = await fetch(requestURL.toString(), requestOptions)
+            .then(response => {
+                const status = response.status;
+                if (status == 200) {
```
brendandburns, line 128: probably not strictly necessary, but triple equals here?

brendandburns:
Couple of comments, is there a unit test that needs updating also?

src/log.ts
davidgamero: got it

davidgamero:
> Couple of comments, is there a unit test that needs updating also?

i don't believe there was a log_test, but i added one for param parsing

Issue #825, opened 2022-06-19
Empty `client-secret` is not refreshed


uesyn:
**Describe the bug**
When using OIDC for client authentication, if `client-secret` is empty, token refresh will fail.
It is allowed for `client-secret` to be an empty value. It is specified in [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1).
However, if client-secret is empty, an error occurs and the token cannot be refreshed because its value is empty.
The reason for the failure is that when refreshing a token with empty `client-secret`, the node-openid-client client must set the `token_endpoint_auth_method` to `none`. ref: https://github.com/panva/node-openid-client/issues/402#issuecomment-916203372

error message:
```
TypeError: client_secret_basic client authentication method requires a client_secret
    at Client.authFor (/Users/test/src/tmp/nodejs/node_modules/openid-client/lib/helpers/client.js:124:15)
    at Client.authenticatedPost (/Users/test/src/tmp/nodejs/node_modules/openid-client/lib/helpers/client.js:163:30)
    at Client.grant (/Users/test/src/tmp/nodejs/node_modules/openid-client/lib/client.js:1336:46)
    at Client.refresh (/Users/test/src/tmp/nodejs/node_modules/openid-client/lib/client.js:1091:33)
    at OpenIDConnectAuth.refresh (/Users/test/src/tmp/nodejs/node_modules/@kubernetes/client-node/dist/oidc_auth.js:74:43)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async OpenIDConnectAuth.applyAuthentication (/Users/test/src/tmp/nodejs/node_modules/@kubernetes/client-node/dist/oidc_auth.js:46:23)
    at async KubeConfig.applyAuthorizationHeader (/Users/test/src/tmp/nodejs/node_modules/@kubernetes/client-node/dist/config.js:349:13)
    at async KubeConfig.applyOptions (/Users/test/src/tmp/nodejs/node_modules/@kubernetes/client-node/dist/config.js:357:9)
    at async KubeConfig.applyToRequest (/Users/test/src/tmp/nodejs/node_modules/@kubernetes/client-node/dist/config.js:90:9)
```

**Client Version**
e.g. `0.16.3`

**Server Version**
e.g. `1.23.7`

**To Reproduce**
1. Prepare OIDC provider accept empty client-secret.
[...]

PR #826, opened 2022-06-19
Support to refresh the tokens correctly when client-secret is set empty

uesyn:
Fixes https://github.com/kubernetes-client/javascript/issues/825


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #755, opened 2021-12-16
Fetch migration docs

OliverMKing:
Add fetch migration docs to repo. #754

PR #829, opened 2022-06-24
remove Object.ts and get unit tests runnable again

davidgamero:
Object.ts is too deeply entangled with the old ObjectSerializer and generated authentication patterns and would require a full re-write, so I am removing it from the initial migration release to streamline us back towards the primarily generated code from openapigenerator.

Otherwise mostly just import fixes and auth interface updates.
With this PR the unit tests are runnable (albeit prevalently failing) which sets up the next steps of migration in a much more manageable way for distributing work.
#755 #754


clintonmedbery:
@brendandburns @davidgamero I am trying to get the apply example to work again. This seems kind of impossible after removing this. Do we have plans to reinstate something similar?

davidgamero:
Hey @clintonmedbery we removed object.ts to be able to fully remove the request dependency since it's so heavily dependent on it. The plan is definitely to bring back a new version built on fetch, and we'd welcome a PR!

Issue #828, opened 2022-06-23
V1Lease fails to serialize Date


kurtb:
**Describe the bug**
Attempting to create/update a Lease using the CoordinationV1Api fails due to the Date serializer calling toISOString(). Kubernetes expects this string to use the microsecond format (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#leasespec-v1-coordination-k8s-io) - not the millisecond format that the Date object provides. (i.e. Date creates 2022-06-23T18:43:04.180Z but Kubernetes expects 2022-06-23T18:43:04.180000Z).

The response from the Kubernetes API is
```
 "message": "Lease in version \"v1\" cannot be handled as a Lease: v1.Lease.Spec: v1.LeaseSpec.HolderIdentity: AcquireTime: unmarshalerDecoder: parsing time \"2022-06-23T18:10:50.278Z\" as \"2006-01-02T15:04:05.000000Z07:00\": cannot parse \".278Z\" as \".000000\", error found in #10 byte of ...|0:50.278Z\",\"holderId|..., bigger context ...|},\"spec\":{\"acquireTime\":\"2022-06-23T18:10:50.278Z\",\"holderIdentity\":\"test\",\"leaseDurationSeconds\":60|...",
```

As a workaround you can create a custom Date that overrides toISOString to convert to the correct format (see example code).

** Client Version **
0.16.3

** Server Version **
1.22.3

**Expected behavior**
Lease is created/updated

** Example Code**
```
import * as k8s from '@kubernetes/client-node';

export async function brokenTest() {
    const kc = new k8s.KubeConfig();
    kc.loadFromDefault();

    const leaseApi = kc.makeApiClient(k8s.CoordinationV1Api);
    await leaseApi.createNamespacedLease(
        'default', 
        {
            apiVersion: 'coordination.k8s.io/v1',
            kind: 'Lease',
            metadata: {
                name: 'test',
                namespace: 'default',
            },
            spec: {
                acquireTime: new Date(),
                holderIdentity: 'test',
                leaseDurationSeconds: 60,
                leaseTransitions: 0,
            },
        });
}

class MicroDate extends Date {
    public toISOString(): string {
        const str = super.toISOString();
        return str.slice(0, -1) + '000Z'; 
    }
}

export async function workingTest() {
    const kc = new k8s.KubeConfig();
    kc.loadFromDefault();

    const leaseApi = kc.makeApiClient(k8s.CoordinationV1Api);
[...]

PR #830, opened 2022-06-25
Support to serialize MicroTime fields

uesyn:
Some Kubernetes objects expect to have date fields represented RFC3339 format with microsecond.
However, Kubernetes is unable to represent microsecond field in openapi schema.
ref: https://github.com/kubernetes/kubernetes/issues/97904

In this PR, patch the kubernetes/gen which generate clients based on the open api schema generated by kubernetes, and define V1MicroTime to represent microseconds.
I confirmed the below code works.

```typescript
import * as k8s from "@kubernetes/client-node";

async function main(): Promise<void> {
  const config = new k8s.KubeConfig();
  config.loadFromDefault();

  const cli = config.makeApiClient(k8s.CoordinationV1Api);
  await cli.createNamespacedLease("default", {
    metadata: {
      name: "foo",
      namespace: "default",
    },
    spec: {
      acquireTime: new k8s.V1MicroTime(),
      holderIdentity: "foo",
    },
  });
}

(async () => {
  await main();
})()
```

ref: https://github.com/kubernetes-client/javascript/issues/828

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -14,3 +14,9 @@ export interface KubernetesListObject<T extends KubernetesObject> {
 }
 
 export type IntOrString = number | string;
+
+export class V1MicroTime extends Date {
+    public toISOString(): string {
+        return super.toISOString().slice(0, -1) + '000Z';
+    }
+}


uesyn:
@brendandburns Thank you for the comment!

I created [the PR for kuberentes/gen](https://github.com/kubernetes-client/gen/pull/224)!

I checked kubernetes/gen's master branch, and found that it's typescript's generatorName is `typescript`(for 1.x.y versions) not `typescript-node`(for 0.x.y versions).
ref: https://github.com/kubernetes-client/gen/pull/219

Therefore, to generate codes with `typescript-node`, we need to create a new branch from [this commit](https://github.com/kubernetes-client/javascript/blob/master/settings#L18), and cherry-pick the PR to it.

uesyn:
> Once that PR is in, we can rebase this one to pick up those changes and then merge this one.

I rebased this PR in advance that the PR is merged.
I dropped the commit to patch kubernetes/gen. 
Should I drop [the commit which regenerate the codes](https://github.com/kubernetes-client/javascript/pull/830/commits/7b9442776751e1267c2bffcd2c047a76c132427a) too?

brendandburns:
Thanks I LGTM'd the gen PR. You're right that we need to branch and pull it back to the commit that we use for generation, but for now let's just merge this PR also.

uesyn:
@brendandburns 
Thank you for the review of the kubernetes/gen PR!
I rebased this PR, and make it contain only V1MicroTime type definition. PTAL😄

brendandburns:
I think the next step is to run the Github action to regenerate the code.

uesyn:
@brendandburns 
I have the same understanding.
If you don't mind, I will make PR 😄.

Issue #838, opened 2022-07-03
Patch fails on custom resouce with UnsupportedMediaType error


dominic-p:
**Describe the bug**
When I try to run a patch operation on a custom resource, I get an error like this:

> the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml

Some Googling took me to [this issue](https://github.com/kubernetes/kubernetes/issues/97423) which indicates that the solution is to use a different patch type for custom resources. It looks like right now the strategic merge patch is [the default](https://github.com/kubernetes-client/javascript/blob/0.17.0/src/object.ts#L535) for all patch requests.

Of course, I can override the headers which is great. But, could that code be made "smarter" to recognize if we are patching a custom resource and then default to a different patch strategy?

**Client Version**
`0.16.3`

**Server Version**
`1.23.7`

brendandburns:
@dominic-p can you send an example of the code that you are using? I'm not sure if there is a way to make this more automatic but we can look into it.

dominic-p:
Thanks for the response. My code right now it pretty custom. I'm using a function like the [apply example](https://github.com/kubernetes-client/javascript/blob/0.17.0/examples/typescript/apply/apply-example.ts) to deploy YAML to the cluster. When I call that function with a Custom Resource (in my case a ClusterIssuer from cert-manager), I simply use the headers override on the patch function to use a different `content-type`.

I'm still pretty new to this code base (and kubernetes in general), so I'm not sure what would be involved in this. In my mind, if there were a magic function like `isCustomResource()` that could look at a spec and return true or false, then we could simply check that on the patch function and chose a default `content-type` header appropriately.

If that function doesn't exist or would be brittle/hard to write then this probably doesn't make a lot of sense to attempt.

pillumina:
> Thanks for the response. My code right now is pretty custom. I'm using a function like the [apply example](https://github.com/kubernetes-client/javascript/blob/0.17.0/examples/typescript/apply/apply-example.ts) to deploy YAML to the cluster. When I call that function with a Custom Resource (in my case a ClusterIssuer from cert-manager), I simply use the headers override on the patch function to use a different `content-type`.
> 
> I'm still pretty new to this code base (and kubernetes in general), so I'm not sure what would be involved in this. In my mind, if there were a magic function like `isCustomResource()` that could look at a spec and return true or false, then we could simply check that on the patch function and chose a default `content-type` header appropriately.
> 
[...]

PR #840, opened 2022-07-11
Add comment about patching custom resources on apply example

dominic-p:
When running a `patch` operation on a custom resource the default strategic merge strategy can fail. Since it's difficult to identify custom resources, we leave a note to the developer that they need to be aware of this when using this example code.

See: #838


brendandburns:
@dominic-p looks like CI/CD is failing.

dominic-p:
It looks like one of the lines was a bit too long. It should be good to go now.

brendandburns:
/lgtm
/approve

Thanks!

PR #847, opened 2022-07-26
Move types to devDependencies

### LINKED ISSUE #846: Types bundled as dependencies
**Describe the bug**
Some of this package's dependencies include types (e.g. `@types/tar`). These dependencies should be moved to the `devDependencies` section in the package.json

** Client Version **
`0.17.0`

** Server Version **
N/R

**To Reproduce**
`npm ci`

**Expected behavior**
The `dependencies` section should include only runtime dependencies

** Example Code**
N/R

 - OS: N/R
 - NodeJS Version: N/R
 - Cloud runtime: N/R

**Additional context**
N/A

yotamselementor:
Fixes #846

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "ts-node": "^8.2.0",
     "tslint": "^5.17.0",
     "typedoc": "^0.22.15",
-    "typescript": "~4.3.5"
+    "typescript": "~4.1.3"
   },
   "optionalDependencies": {
     "openid-client": "^5.1.6"
--- a/package.json
+++ b/package.json
@@ -54,13 +54,6 @@
   "author": "Kubernetes Authors",
   "license": "Apache-2.0",
   "dependencies": {
-    "@types/js-yaml": "^4.0.1",
-    "@types/node": "^10.12.0",
-    "@types/request": "^2.47.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -96,7 +96,7 @@
     "ts-node": "^8.2.0",
     "tslint": "^5.17.0",
     "typedoc": "^0.22.15",
-    "typescript": "^4.1.3"
+    "typescript": "~4.3.5"
```
brendandburns, line 99: Can you do the typescript upgrade as a separate PR?

yotamselementor: @brendandburns - before this change, if I was to regenerate the package-lock file, it would fetch the latest typescript version, which fails the build.
I see 3 options here:
1-Keep the version currently in the lock file (4.1.3), which will require me to modify the TS version to '~4.13' or '4.1.3' in this PR
2-Follow the implementation of '^4.1.3' and try to fetch the latest compatible version of TS without migrating code (this is what I did)
3-Follow the implementation of '^4.1.3' and try to fetch the latest compatible version of TS, while migrating the code as necessary 

I believe that option 1 will be the best way to go, is that what you want me to do?

brendandburns: Let's do option 1. Thanks

PR #848, opened 2022-07-26
Add single pod metrics

RCCSilva:

PR #849, opened 2022-07-27
deploy

j0pgrm:


brendandburns:
This seems like a spam PR that is a duplicate of #848

PR #850, opened 2022-07-28
Regenerate with new openapi generator

uesyn:
**This PR is for release-1.x branch.**

openapi-generator supported import-mapping for typescript client generator.
ref: https://github.com/OpenAPITools/openapi-generator/pull/12957

With it, we don't need to do workaround like copying IntOrString type to IntOrString.ts.
In this PR, re-generate codes using openapi generator which supports import-mapping.
In addition, cherry-pick https://github.com/kubernetes-client/javascript/pull/830.


brendandburns:
cc @davidgamero 

looks like a number of the paths changed here? (e.g. the `models` directory was added) do we know why things moved around? It doesn't necessarily bother me, but I want to make sure that it is known/intentional.

brendandburns:
Also, we really aren't running any unit tests on this branch so we need to be extra careful.

davidgamero:
hello @uesyn, nice to see you again!
the input mapping support would make this transition much easier compared to the workarounds I've been using. 
It seems worth it to me to figure out the changes and jump to this commit for the gen.

davidgamero:
let me take a shot at getting a couple more unit tests running this weekend so we can verify we can retain progress.
It'd be nice to at least keep our current progress on unit tests when making this kind of generated file change

uesyn:
@brendandburns 
cc: @davidgamero

> looks like a number of the paths changed here? (e.g. the models directory was added) do we know why things moved around? It doesn't necessarily bother me, but I want to make sure that it is known/intentional.

When import mapping feature is added to openapi-generator, the template of generated codes was changed to simplify openapi-generator implementation.
As a result, generated import path become redundant. This is the why that `models` is added in some lines.
for example `import { V1CronJobSpec } from './V1CronJobSpec';` was converted to `import { V1CronJobSpec } from '../models/V1CronJobSpec';`

ref: https://github.com/OpenAPITools/openapi-generator/pull/12957#issuecomment-1196736939

The reason why that the lines related to IntOrString and V1MicroTime is changed is the result of generating with import mapping feature. (main purpose of this PR)

The other lines changed is the result that `supportES6` option was reflected.
ref: https://github.com/kubernetes-client/gen/blob/master/openapi/typescript.xml#L31

davidgamero:
i tried running the tests on your branch and it looks like you forked at #44f5170046d7cecd0738935942ce2d92f7422690 which is a couple commits behind the current state of the release-1.x branch. 

Could you merge the release-1.x branch into yours or regenerate on top of the current head so I can give it a try?
@uesyn

davidgamero:
yep i can confirm that the unit tests run and the progress is retained
/lgtm
@uesyn @brendandburns

brendandburns:
/lgtm
/appprove

Thanks! (apologies for the delay, I was on vacation)

brendandburns:
/approve

(instead of /appprove :P)

Issue #852, opened 2022-07-28
KubernetesObjectApi does not properly return typed objects


schrodit:
**Describe the bug**
When using the general `KubernetesObjectApi` the objects are not returned as typed objects as defined in the api definitions (https://github.com/kubernetes-client/javascript/tree/master/src/gen/api), but rather the plain json value is returned.

** Client Version **
e.g. `0.16.3`

** Server Version **
e.g. `1.21.0`

**To Reproduce**

see code snippet

**Expected behavior**
I expect that known objects are returned with their proper types.

** Example Code**
```
const k8s = require('@kubernetes/client-node');

const kc = new k8s.KubeConfig();
kc.loadFromDefault();

const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
const client = kc.makeApiClient(k8s.KubernetesObjectApi);

client.list('v1', 'Pod', 'default').then(res => {
    const p = res.body.items[0];
    console.log(typeof p.metadata.creationTimestamp); // "string"
})

k8sApi.listNamespacedPod('default').then((res) => {
    const p = res.body.items[0];
    console.log(typeof p.metadata.creationTimestamp.getTime()); // "Object" --> Date
});
```

**Environment (please complete the following information):**
 - OS: [e.g. Linux]
 - NodeJS Version [eg. 16]

**Enhancement Proposal**

I guess that the current behavior is intended to also support unknown objects.
Also one can currently work around the described issue by using the ObjectSerializer like that

```
const k8s = require('@kubernetes/client-node');
const res = await client.read({apiVersion: 'v1', 'kind': 'Pod', metadata: {'name': 'a', 'namespace': 'default'}});
const pod = res.body; // type: KubernetesObject
pod.metadata.creationTimestamp // type "string"

const p = ObjectSerializer.serialize(pod, 'V1Pod');
```

But I think there could be a good opportunity to simplify the way the generic client can be used and improve the typing
Maybe something like that:
```
class KubernetesObjectApi {
    public read<T extends KubernetesObject | KubernetesObject>(spec: T, ...): {res: http.response, body: T} {
        // do the stuff that is currently done
        return this.requestPromise(localVarRequestOptions, this.getKnownType(spec));
    }
    
    private getKnownType(spec: KubernetesObject): string {
      // get type from spec + V1APIResource
    }

}

const spec: V1Pod = {apiVersion: 'v1', 'kind': 'Pod', metadata: {'name': 'a', 'namespace': 'default'}};
[...]

PR #854, opened 2022-08-01
Return strong types in generic client

schrodit:
- Also use generics to make the generic client easier usable without casting.
  e.g.
  ``` typescript
  const client = kc.makeApiClient(k8s.KubernetesObjectApi);
  const pod = (await client.read<V1Pod>({apiVersion: 'v1', kind: 'Pod', metadata: {name: 'test', namespace: 'default'}})).body;
  ```
- The `read` function is more strongly typed to better represent that only the header (apiVersion, kind, metadata.name) is needed
- The usage of the client is backwards compatible so all current code should work as before. But some types might change.
   e.g. the following types differ when dealing with pods (or other known types)
   
``` typescript
const client = kc.makeApiClient(k8s.KubernetesObjectApi);
const pod = (await client.read({apiVersion: 'v1', kind: 'Pod', metadata: {name: 'test', namespace: 'default'}})).body;

// before
pod.metadata.creationTimestamp // --> string
// with this change
pod.metadata.creationTimestamp // --> Date
```

Issue: https://github.com/kubernetes-client/javascript/issues/852

⚠️ There is still an issue that custom resources or `KubernetesObject` is not correctly parsed, due to the current `ObjectSerialzer.deserializer` implementation. I will provide a separate PR to fix this issue

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -22,6 +22,21 @@ type KubernetesObjectResponseBody =
 /** Kubernetes API verbs. */
 type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'list' | 'replace';
 
+type KubernetesObjectHeader<T extends KubernetesObject | KubernetesObject> = Pick<
+    T,
+    'apiVersion' | 'kind'
+> & {
+    metadata: {
+        name: string;
+        namespace: string;
+    };
+};
+
+interface GroupVersion {
+    group: string;
+    version: string;
+}
+
 /**
  * Valid Content-Type header values for patch operations.  See
  * https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/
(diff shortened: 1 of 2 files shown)


src/object.ts
```diff
@@ -275,17 +290,24 @@ export class KubernetesObjectApi extends ApisApi {
...
+    ): Promise<{ body: T; response: http.IncomingMessage }> {
         // verify required parameter 'spec' is not null or undefined
         if (spec === null || spec === undefined) {
             throw new Error('Required parameter spec was null or undefined when calling read.');
         }
+        // verify required parameter 'kind' is not null or undefined
+        if (spec.kind === null || spec.kind === undefined) {
+            throw new Error('Required parameter spec.kind was null or undefined when calling read.');
+        }
+        if (!spec.apiVersion) {
```
brendandburns, line 308: I think we should throw an Error here instead of defaulting, I suspect that this will confuse people.

brendandburns:
One minor comment, otherwise lgtm.

src/object.ts
schrodit: Done

Issue #856, opened 2022-08-17
Errors from watch callback function are caught


Nicklason:
**Describe the bug**
When using the Watch class and watch function, if there is an error in the callback function, then the error is caught by the watch class, instead of being thrown.

https://github.com/kubernetes-client/javascript/blob/eaf1b23537df96a4f6e93bdd43a971d86bf5576a/src/watch.ts#L123-L130

The function is not only ignoring parsing errors, but also errors made by the callback function. The callback should be called outside the try-catch to solve the problem.

**Client Version**
`0.17.0`

**Server Version**
`n/a`

**To Reproduce**

1. Create instance of Watch class and call watch function using a callback function
2. Throw an error inside the callback function
3. Error is caught by watch function instead of resulting in an uncaught exception

**Expected behavior**
The error should not be caught, an uncaught exception should occur.

**Example Code**

```js
const k8s = require('@kubernetes/client-node');

const kc = new k8s.KubeConfig();
kc.loadFromDefault();

const watch = new k8s.Watch(kc);
watch.watch('/api/v1/namespaces', {}, () => {
    // Error is thrown but caught by watch function 
    throw new Error('something bad happened');
}, () => {});
```

**Environment (please complete the following information):**
 - OS: Windows 11
 - NodeJS Version: v18.3.0

brendandburns:
Closing since #857 merged.

PR #857, opened 2022-08-17
Move watch callback out of try-catch

Nicklason:
Issue: #856


brendandburns:
Thanks for the PR!

Can you add a unit test to cover this behavior? You should be able to adapt the code from here: https://github.com/kubernetes-client/javascript/blob/master/src/watch_test.ts#L320 but remove the parse error and set your watch handler to throw.

Nicklason:
@brendandburns I have added the test. It tests if the handler receives the correct values and if the promise fails because of the handler error without calling the done callback.

PR #860, opened 2022-08-22
Fix formatting

Nicklason:
Looks like tests are failing because I didn't format the test code from my previous pull request #857.

https://github.com/kubernetes-client/javascript/actions/runs/2905160040

https://github.com/kubernetes-client/javascript/runs/7955503529?check_suite_focus=true#step:7:11

### DIFF
--- a/src/watch_test.ts
+++ b/src/watch_test.ts
@@ -396,22 +396,24 @@ describe('Watch', () => {
 
         const handlerError = new Error('handler error');
 
-        await watch.watch(
-            path,
-            {},
-            (recievedType: string, recievedObject: string) => {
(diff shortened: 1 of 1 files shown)


brendandburns:
@Nicklason thanks for fixing this. I don't know why the pre-merge tests didn't catch this...

/lgtm
/approve

Nicklason:
@brendandburns can it have something to do with this? https://github.com/kubernetes-client/javascript/blob/master/.github/workflows/test.yml#L6-L7

It looks like tests are only ran when a pull request is made, and not when a new commit is added to the pull request. I can make a PR for this.

PR #872, opened 2022-09-07
Update example with correct parameters

dominykas:


brendandburns:
/lgtm
/approve

Thanks! We should probably add some sort of tests for these examples so that we catch such problems in the future...

PR #879, opened 2022-09-18
consider tar return code when cpToPod

qweritos:
`tar` also using stderr for notifications if it writes output into stdout pipe

It helps to avoid several "errors", such as `tar removing leading '/' from member names`, which's not actually an error.

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -40,8 +40,8 @@ export class Cp {
             errStream,
             null,
             false,
-            async () => {
-                if (errStream.size()) {
+            async ({status}) => {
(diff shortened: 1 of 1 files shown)


src/cp.ts
```diff
@@ -40,8 +40,8 @@ export class Cp {
             errStream,
             null,
             false,
-            async () => {
-                if (errStream.size()) {
+            async ({status}) => {
+                if (status === 'Failure' && errStream.size()) {
```
brendandburns, line 44: Perhaps this should be `||` not `&&`?

brendandburns:
Closing in favor of #943 since this went quiet.

Issue #876, opened 2022-09-15
Unable to load @kubernetes/client-node since 0.17.1 due to missing openid-client dependency


dcoles15:
Since upgrading to 0.17.1 we're no longer able to load the @kubernetes/client-node module as we don't install optional dependencies and openid-client has been changed to be optional, yet it still seems to be required by the code.
All you need to do is require the module with:

`const k8s = require('@kubernetes/client-node')`

and then you see this require error:

```
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module 'openid-client'
Require stack:
- /test/node_modules/@kubernetes/client-node/dist/oidc_auth.js
- /test/node_modules/@kubernetes/client-node/dist/config.js
- /test/node_modules/@kubernetes/client-node/dist/index.js
- /test/go.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/test/node_modules/@kubernetes/client-node/dist/oidc_auth.js:4:25)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/test/node_modules/@kubernetes/client-node/dist/oidc_auth.js',
    '/test/node_modules/@kubernetes/client-node/dist/config.js',
    '/test/node_modules/@kubernetes/client-node/dist/index.js',
    '/test/go.js'
  ]
}
```

It seems like if the dependency is really optional then you should have to opt into the function that requires the optional dependency. Just requiring the module should not require optional dependencies be installed.

brendandburns:
Yeah, this is a bug, thanks for reporting. We should handle this correctly. I will work on a fix.

felix-gohla:
Hi :) Thank you very much for the fix! Can we expect a new release including that fix in the near future or did I miss something?

nomagick:
[...]

PR #882, opened 2022-09-20
Defer loading of OIDC until it's actually used.

brendandburns:
This should address #876

cc @dominykas


lchrennew (approved):
Nice, maybe there should be a common utility for "deferred-loadification"

Issue #887, opened 2022-09-28
`tls-server-name` is not populated for listing requests.


tigrato:
**Describe the bug**
`tls-server-name` is not set on requests other than WebSockets.

This PR #820 added support for `tls-server-name`, however running `listNamespacedPod` the `tls-server-name` doesn't seem to be configured causing the request to fail.
The request calls this function
https://github.com/kubernetes-client/javascript/blob/4ea2b0f60ba724eee1f2ab1485e39e9ff1390507/src/config.ts#L148-L167
that does not define the `tls-server-name` property.

** Client Version **
e.g. `0.17.1`

** Server Version **
e.g. `1.24.3`

**To Reproduce**

Run the following command against any cluster that requires `tls-server-name` and it will fail.

** Example Code**
```js
const k8s = require('@kubernetes/client-node');

const kc = new k8s.KubeConfig();
kc.loadFromDefault();

const k8sApi = kc.makeApiClient(k8s.CoreV1Api);

k8sApi.listNamespacedPod('default')
    .then((res) => {
	console.log(res.body);
    });
```

**Expected behavior**
The request should be successful 

**Environment (please complete the following information):**
 - OS:  macOS & Linux
 - NodeJS Version: 18

PR #888, opened 2022-09-28
Add `tls-server-name` to https requests if provided

tigrato:
This should address #887

cc @brendandburns

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -161,6 +161,10 @@ export class KubeConfig {
                 username: user.username,
             };
         }
+
+        if (cluster && cluster.tlsServerName) {
+            opts.agentOptions = { servername: cluster.tlsServerName } as https.RequestOptions;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -161,6 +161,10 @@ export class KubeConfig {
                 username: user.username,
             };
         }
+
+        if (cluster && cluster.tlsServerName) {
+            opts.agentOptions = { servername: cluster.tlsServerName } as https.RequestOptions;
```
brendandburns, line 166: I think this should just be `opts.servername = cluster.tlsServerName` instead of `opts.agentOptions`

tigrato:
@brendandburns, thanks for the review.
I pushed unit tests for  `applyToRequest` and `applytoHTTPSOptions` public functions!

src/config.ts
brendandburns: Ah, looks like you're right. Apologies for the mis-direction.

Issue #883, opened 2022-09-22
Missing type dependencies in 0.17.1


Rugvip:
**Describe the bug**
Dependencies are missing for type declarations, leading to type errors for users of this package.

** Client Version **
`0.17.1`

** Server Version **
N/A

**To Reproduce**

- `npm init -y`
- `npm install typescript`
- `npm install @kubernetes/client-node`
- `echo 'import "@kubernetes/client-node"' > index.ts`
- `yarn tsc index.ts` <- Lots of errors

**Expected behavior**
I expect `@kubernetes/client-node` to declare direct or peer dependencies for all types that are referenced in the published type declarations.

**Example Code**
Just a simple `import "@kubernetes/client-node"'`

**Environment (please complete the following information):**
 - OS: Linux
 - NodeJS Version: 16
 - Cloud runtime [e.g. Azure Functions, Lambda]

**Additional context**
Essentially a revert of #847, although not necessarily all of those dependencies.

matteosilv:
The problem is outlined here too https://github.com/kubernetes-client/javascript/issues/864#issuecomment-1243216296

Rugvip:
Fixed by #890 👍

PR #890, opened 2022-10-13
Fix missing types

danmontgomery:
This fixes compilation errors introduced when types all moved to devDependencies - described in #883

By default, there will still be some errors because of default imports, which are all fixed by setting `esModuleInterop` to true (maybe worth documenting, or addressing in a separate PR?)

```bash
npm init -y
npm install typescript
npm install @kubernetes/client-node
echo 'import "@kubernetes/client-node"' > index.ts
npx tsc index.ts
```

Before:
```bash
node ➜ /workspaces/k8s-client $ npx tsc index.ts
node_modules/@kubernetes/client-node/dist/attach.d.ts:1:23 - error TS2688: Cannot find type definition file for 'node'.

1 /// <reference types="node" />
                        ~~~~

node_modules/@kubernetes/client-node/dist/attach.d.ts:2:23 - error TS2688: Cannot find type definition file for 'ws'.

2 /// <reference types="ws" />
                        ~~

node_modules/@kubernetes/client-node/dist/attach.d.ts:4:25 - error TS2307: Cannot find module 'stream' or its corresponding type declarations.

4 import stream = require('stream');

...

Found 166 errors in 71 files.
```

After:
```bash
node ➜ /workspaces/k8s-client $ npx tsc index.ts --esModuleInterop
node ➜ /workspaces/k8s-client $ 
```


brendandburns:
@danmontgomery thanks for the PR, but from my reading of this npm document:

https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file

It seems that `@typings` files should not be included in the dependency section as they are not required for production, they are only required for development.

I'm open to arguments that I'm wrong, but this is the best source I could find on how `package.json` should be set up.

danmontgomery:
@brendandburns That's true, except when publishing a typescript library... "production" in this case are users of the package (anyone adding @kubernetes/client-node to their package.json). With these types not included in `dependencies`, they aren't grabbed by npm/yarn when installing the package, resulting in the errors posted above (and in the issue I linked to), so developers using this package need to manually add these undocumented dependencies to their own package.json

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies

The alternative (at the bottom of the page in the above link) would be to submit a PR to https://github.com/DefinitelyTyped/DefinitelyTyped declaring type dependencies :)

brendandburns:
@danmontgomery thanks for the additional details. I'm ok to merge this.

/lgtm
/approve

rbnayax:
@brendandburns thanks for fixing both typescript issues (Im referring to the types dependencies). When can we expect a new version? We are eager to start working with the library and this is currently a show stopper for us :-)

matteosilv:
Any news on the release?

PR #900, opened 2022-11-16
fix possible RCE in jsonpath-plus

rluvaton:
https://github.com/JSONPath-Plus/JSONPath/issues/182

### DIFF
--- a/src/json_path_test.ts
+++ b/src/json_path_test.ts
@@ -1,0 +1,10 @@
+import { expect } from 'chai';
+import { jsonpath } from './json_path';
+
+describe('jsonpath', () => {
+    it('should throw if vulnerable for RCE (remote code execution)', () => {
+        expect(() => {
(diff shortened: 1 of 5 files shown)


src/json_path_test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect } from 'chai';
+import { jsonpath } from './json_path';
+
+describe('jsonpath', () => {
```
brendandburns, line 4: Can you add a test like this to the gcp_auth test also?

Thanks

rluvaton:
to be honest after I found that issue I looked at the top most packages that use that and fixed that in them to have better security in our ecosystem

The RCE was meant to be in the package itself (even though it's a feature)...

src/json_path_test.ts
rluvaton: done 😄

PR #911, opened 2022-12-14
Add a dependabot configuration.

brendandburns:
About time :P


brendandburns:
Self LGTM so that we can make progress on rev-ing dependencies.

PR #941, opened 2022-12-21
Rev execa to 6.1.0 and fix breaking changes.

brendandburns:
See #934


brendandburns:
There is something weird going on here, where mocha is losing track of how to load .ts files.

/hold

brendandburns:
Ah, I believe it is due to this:

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#faq

We may need to move to ESM also, but that's likely a significant change.

brendandburns:
Closing in favor of #946

PR #942, opened 2022-12-21
Update ws and @types/node, update code to match.

brendandburns:
See #933 and #935


brendandburns:
Misbehaving kubernetes bot won't let this merge, so manual merge.

PR #880, opened 2022-09-18
cpFromPod: close writerStream syncronously

qweritos:
helps to eliminate such non-persistent error:
(node:445240) UnhandledPromiseRejectionWarning: ZlibError: zlib: unexpected end of file

caused by possibly-still-dirty writes from a previous step

PR #943, opened 2022-12-22
Close the writestream before heading over to untar also read failure from status.

brendandburns:
This replicates the work in #880 and #879 with some fixes, since those PRs are a good idea, but have been sitting idle.

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -87,8 +87,8 @@ export class Cp {
             errStream,
             readStream,
             false,
-            async ({status}) => {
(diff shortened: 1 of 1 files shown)


rubenjoy (commented):
it would be great if we check the status failure [line 91](https://github.com/kubernetes-client/javascript/blob/739dce15fa16ba9a59bac92221fedbad4817fede/src/cp.ts#L91) or cpToPod function

brendandburns:
@rubenjoy done.

PR #945, opened 2022-12-22
only import shelljs on windows

webratz:
this should avoid certain bundling issues with esbuild


brendandburns:
Closing in favor of #946

Issue #954, opened 2023-01-04
exec_auth.ts getCredential function mutates entire nodejs process.env when run


edify42:
**Describe the bug**
The `exec.env`  adds all the environment variables passed in through the `KUBECONFIG` to the `process.env` global variable.

For our usage, it introduces a race condition depending on when the call to the k8s API client occurs.

e.g. the KUBECONFIG `users` block below changes the entire nodejs process env:
```yaml
users:
- name: arn:aws:eks:us-east-2:702880000000:cluster/our-cluster  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - ap-southeast-2
      - eks
      - get-token
      - --cluster-name
      - our-cluster
      command: aws
      env:
        - name: teddy
          value: testing
```

It's not really necessary to do this, instead we can pass a copy of the `process.env` with the appended `env` array from the `KUBECONFIG` to the `child_process`.

** Client Version **
e.g. `0.18.0`

** Server Version **
e.g. `1.23.0`

**To Reproduce**

Run the code with the `env` specified in the KUBECONFIG:

```typescript
...
const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
console.log(`how come ${process.env.teddy} is not set?`)
const nodes = await k8sApi.listNode();
console.log(`how come ${process.env.teddy} is set?`)
```

**Expected behavior**
Entire process.env should not be mutated (unknowingly depending on what is stored in `env` of the KUBECONFIG).

**Environment (please complete the following information):**
 - OS: Debian Linux
 - NodeJS Version 16/18

edify42:
Fixed pending release/publish of new version -> https://github.com/kubernetes-client/javascript/pull/955

PR #955, opened 2023-01-04
Copy the nodejs process.env rather than mutate the global state

edify42:
Quick change which fixes the issue I just raised around modifying the global process.env state https://github.com/kubernetes-client/javascript/issues/954


brendandburns:
/lgtm
/approve

Thanks for the PR!

edify42:
@brendandburns failed CI prettier check which I've just changed

PR #962, opened 2023-01-13
add NetworkingV1 Ingress example

vBitza:
Hi,

I recently needed to add a Ingress on the newer "networking.k8s.io/v1" since the example you have is on "networking.k8s.io/v1beta" I decided to add my work here aswell. I've also renamed the first example you had to properly differentiate the two files based on apiVersion.

Have a nice day!

### DIFF
--- a/examples/ingress.js
+++ b/examples/ingress.js
@@ -1,27 +1,43 @@
-const k8s = require('@kubernetes/client-node')
-const kc = new k8s.KubeConfig()
-kc.loadFromDefault()
+const k8s = require('@kubernetes/client-node');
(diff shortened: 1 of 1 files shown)


brendandburns:
Thanks for the PR, we need to have you sign the CLA.

Also, since the v1beta1 Ingress API is deprecated, I think you can just modify the existing ingress example to use `v1` instead of `v1beta1`

Thanks!

vBitza:
CLA signed and PR updated as per request.

brendandburns:
/lgtm
/approve

Thanks!

PR #966, opened 2023-01-21
Fix push script to be able to run without input and skip tagging.

brendandburns:


.github/workflows/release.yml
```diff
@@ -22,3 +22,5 @@ jobs:
         run: ./push-package.sh
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          confirm: "y"
+          SKIP_TAG: "y"
```
mstruebing, line 26: Why don't we want to create a git tag for every npm release? Wouldn't that be better?

PR #988, opened 2023-02-15
fix: resolve copy to and from pod when the copying is completed

### LINKED ISSUE #982: cp.cpToPod() does not work
**Describe the bug**
[`cp.cpToPod()`](https://github.com/kubernetes-client/javascript/blob/master/src/cp.ts#L63) does not actually copy the file to the pod. No error is raised and the function resolves, that is, there is no indication that the copying had failed. However, the file does not exist in the pod.

After inspecting the source code, I found that the function resolves because a `Promise` is not used. 
I found that the root cause is because the `const command = ['tar', 'xf', '-', '-C', tgtPath];` hangs forever and displays no errors. It seems to me that the command `tar xf -C` is waiting for something and hangs infinitely.

**Client Version**
`"@kubernetes/client-node": "^0.18.0"`

**To Reproduce**
1. Create a pod called in `nginx-4217019353-9gl4s` in `default` namespace with container called `nginx`.
2. Use example cp code
```javascript

kc.loadFromDefault();

const cp = new k8s.Cp(kc);
cp.cpFromPod('default', 'nginx-4217019353-9gl4s', 'nginx', './test.txt', '/tmp');
```
3. Connect to the pod and inspect for the file
```bash
kubectl exec --stdin --tty nginx-4217019353-9gl4s -- /bin/bash
cd /tmp
ls test.txt
```

**Expected behavior**
I expect the file to be copied into the pod, and for the file to exist in the pod.

**Environment:**
 - OS: MacOS
 - NodeJS Version: v16.17.0
 - Cloud runtime: DigitalOcean Kubernetes

**Additional context**
I'm currently using a hack to work around this by spawning a process to call `kubectl cp`.
However, it's not ideal since it depends on the `kubectl` command line tool:
```typescript
import * as k8s from "@kubernetes/client-node";
import { Cp } from "@kubernetes/client-node";
import { spawn } from "child_process";
import * as fs from "fs";
import * as tmp from "tmp-promise";

export class CustomCp extends Cp {
   /**
   * Modified version of cpToPod() to copy files from a string to pod without writing files
   * NOTE: the cpToPod() does not seem to work (no copy made). It resolves but the tar command hangs.
   * @param {string} namespace - The namespace of the pod to exec the command inside.
   * @param {string} podName - The name of the pod to exec the command inside.
   * @param {string} containerName - The name of the container in the pod to exec the command inside.
   * @param {string} srcString - The source string to copy
   * @param {string} tgtPath - The target path in the pod
   */
  public async cpStringToPod(
    namespace: string,
    podName: string,
    containerName: string,
    srcString: string,
    tgtPath: string
  ): Promise<void> {
    return new Promise(async function (resolve, reject) {
      // (1) Create a temporary file
      const tmpFile = tmp.fileSync();
      const tmpFilename = tmpFile.name;
      // (2) Write the string to the temporary file
      await fs.writeFileSync(tmpFilename, srcString);
      // (3) Copy the temporary file to the pod using `kubectl cp` command line
      const cp = await spawn("kubectl", ["cp", tmpFilename, `${namespace}/${podName}:${tgtPath}`, "-c", containerName]);
      cp.stderr.on("data", (data) => {
        tmpFile.removeCallback(); // remove file
        reject(data);
      });

      cp.on("close", (code) => {
        tmpFile.removeCallback(); // remove file
        if (code == 0) resolve();
        else reject(`exited with code ${code}`);
      });
    });
  }
}
export const cp = new CustomCp(kc, exec);
cp.cpStringToPod(...)
```

SayakMukhopadhyay:
This PR fixes #982

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -31,7 +31,7 @@ export class Cp {
         const command = ['tar', 'zcf', '-', srcPath];
         const writerStream = fs.createWriteStream(tmpFileName);
         const errStream = new WritableStreamBuffer();
-        this.execInstance.exec(
+        const conn = await this.execInstance.exec(
             namespace,
             podName,
             containerName,
@@ -45,12 +45,18 @@ export class Cp {
                 if (status === 'Failure' || errStream.size()) {
                     throw new Error(`Error from cpFromPod - details: \n ${errStream.getContentsAsString()}`);
                 }
-                await tar.x({
-                    file: tmpFileName,
-                    cwd: tgtPath,
-                });
             },
         );
+        return await new Promise((resolve) => {
+            conn.addEventListener('close', (event) => {
+                resolve(
+                    tar.x({
+                        file: tmpFileName,
+                        cwd: tgtPath,
+                    }),
+                );
+            });
+        });
     }
 
     /**
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -78,7 +78,7 @@ export class Cp {
         );
         const readStream = fs.createReadStream(tmpFileName);
         const errStream = new WritableStreamBuffer();
-        this.execInstance.exec(
+        const conn = await this.execInstance.exec(
             namespace,
             podName,
             containerName,
@@ -93,5 +93,10 @@ export class Cp {
                 }
             },
         );
+        return new Promise((resolve) => {
+            conn.addEventListener('close', (event) => {
+                resolve();
+            });
+        });
     }
 }


SayakMukhopadhyay:
I noticed that a similar issue as mentioned in #982 occurs when copying from a pod too. The copy resolves before the copying is done. So, I have added another commit. This implements the same functionality in `copyFromPod` too as implemented in `copyToPod` in the first commit.

SayakMukhopadhyay:
Let me see why the tests are failing

SayakMukhopadhyay:
I must say that I am having some difficulty with the tests. Can one of the reviewers review the existing code as is so that I know if its worth it to pursue working on the tests.

PR #989, opened 2023-02-15
feat: allow to set the cwd in `copyToPod` for use in tar

### LINKED ISSUE #987: Set the working directory while copying
**Describe the bug**
While copying items from a pod or to a pod, I want to set a working directory, the contents of which are copied and not the whole file path. Right now, if I call `cpToPod` for example as
```ts
await cp.cpToPod(
  'default',
  'pod-name',
  'copy',
  'path/to/the/folder/to/copy',
  '/workdir',      
```
Now, if I only want to copy the `copy` folder into `workdir` so that the contents in my pod look like `workdir/copy/...`, there is no way to achieve that. The above will result in `workdir/path/to/the/folder/to/copy/...` which is not the intention. Moreover, `tar` expects that the `srcPath` to be a sub directory of the current working directory, which by default is the source code directory.

**Client Version**
e.g. `0.18.1`

**Server Version**
e.g. `1.23`

**To Reproduce**
Try to copy the contents of a folder outside the working directory of the program such that the folder is copied and not its parent folder too.

**Expected behavior**
I want to be able to set the working directory for the underlying `tar` library used by `cp` so that when the tar is created to copy, it contains the folder tree I want and not every folder in the tree all the way to root.

**Example Code**
Shown above

 - OS: Windows/Linux
 - NodeJS Version: 18.x.x
 - Cloud runtime: None

**Additional context**
This can be handled by either taking an optional working directory or we can take an optional config to pass to the `tar` library. I want the former but the later would open up greater ways to use the underlying `tar`.

I am willing to submit a PR. But here is a snippet of what I want to do.

```ts
public async cpToPod(
    namespace: string,
    podName: string,
    containerName: string,
    srcPath: string,
    tgtPath: string,
    cwd?: string,
): Promise<void> {
    ...
    await tar.c(
        {
            file: tmpFile.name,
            cwd,
        },
        [srcPath],
    );
   ...
}
```

SayakMukhopadhyay:
This PR fixes #987

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -18,6 +18,7 @@ export class Cp {
      * @param {string} containerName - The name of the container in the pod to exec the command inside.
      * @param {string} srcPath - The source path in the pod
      * @param {string} tgtPath - The target path in local
+     * @param {string} cwd - The directory that is used as the parent in the pod when downloading
      */
     public async cpFromPod(
         namespace: string,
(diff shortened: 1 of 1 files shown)


leahjlou:
Hi @brendandburns, do you have an idea of when there will be a new release available on NPM? I'm interested in using the code in this commit in something I'm working on. Thank you!

PR #1002, opened 2023-03-02
maintenance: added types to `KubeConfig#loadFromOptions` input

emeraldsanto:
### Summary

This PR adds stronger types to the `KubeConfig#loadFromOptions` method which is currently accepting `any`, even though all the necessary types are already defined and available to use.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -180,7 +180,12 @@ export class KubeConfig {
         this.currentContext = obj['current-context'];
     }
 
-    public loadFromOptions(options: any): void {
+    public loadFromOptions(options: {
+        clusters: Cluster[];
+        contexts: Context[];
+        currentContext: Context['name'];
+        users: User[];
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for the PR, this looks fine but CI/CD is failing b/c of the stricter types. 

I think we need to make some fields in the existing types optional (e.g. `skipTlsVerify`)

emeraldsanto:
@brendandburns good catch! I've gone ahead with the change and confirmed the tests pass locally. However that type has not been changed in 5 years so I'm worried that marking this field as optional now might be a breaking change?

brendandburns:
I looked at the code and I think we'll be ok. Most people are probably not including it anyway...

brendandburns:
(and fwiw, adding stronger types is probably a breaking change anyway :) )

PR #1004, opened 2023-03-03
fixes the bad workflow file and removes the boilerplate

clintonmedbery:
I think my last PR broke the CodeQL workflow. Furthermore I didn't realize we could ignore some warnings from the console, making ignoring the tests a little redundant. Sorry! This should put things back the way they were. I also removed the boilerplate.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,9 +30,6 @@ jobs:
       fail-fast: false
       matrix:
         language: [ 'javascript' ]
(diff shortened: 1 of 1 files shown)


clintonmedbery:
@brendanburns @davidgamero Sorry about breaking that workflow. I think you can just go to that codeql warning and mark it as a test. If you want me to take another crack at ignoring it, let me know.

brendandburns:
/lgtm
/approve

Thanks!

PR #964, opened 2023-01-20
bump openapigen commit and mocha to get native javascript url wrappers

davidgamero:
get our URL parsing to use the native URL instead of a deprecated dependency we inherited from OpenAPIGen
https://github.com/OpenAPITools/openapi-generator/pull/14319

also remove the last of our shims 'models/IntOrString' residual file that isn't needed since the import mapping works now!

PR #1008, opened 2023-03-14
bump openapi gen commit for URL parsing and audit fix

davidgamero:
rewrite #964 to a smaller chunk of code since @clintonmedbery kindly fixed a bunch of dependency issues

get our URL parsing to use the native URL instead of a deprecated dependency we inherited from OpenAPIGen
https://github.com/OpenAPITools/openapi-generator/pull/14319

also remove the last of our shims 'models/IntOrString' residual file that isn't needed since the import mapping works now!


brendandburns:
/lgtm
/approve

Thanks!

PR #1022, opened 2023-03-24
remove request usages from auth & config

davidgamero:
in reference to #754, this PR removes usages of the request library (specifically the request.RequestOptions) in several interfaces to now fully rely on the https.RequestOptions instead which is compatible with fetch.


mstruebing (approved):
LGTM

PR #1033, opened 2023-03-31
fix: remove final refrences to request + fix tests

rbnayax:
I removed the request library (had to install form-data instead of it)
I added the node-fetch types
fixed log, watch, catch and metrics
Fixed all tests

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -189,8 +189,6 @@ export class KubeConfig implements SecurityAuthentication {
         agentOptions.passphrase = opts.passphrase;
         agentOptions.rejectUnauthorized = opts.rejectUnauthorized;
         agentOptions.timeout = opts.timeout;
-        agentOptions.host = opts.host;
-        agentOptions.path = opts.path;
         agentOptions.servername = opts.servername;
         agentOptions.ciphers = opts.ciphers;
         agentOptions.honorCipherOrder = opts.honorCipherOrder;
@@ -201,10 +199,6 @@ export class KubeConfig implements SecurityAuthentication {
         agentOptions.secureOptions = opts.secureOptions;
         agentOptions.secureProtocol = opts.secureProtocol;
         agentOptions.sessionIdContext = opts.sessionIdContext;
-        // setting undefined will cause tls error
-        if (opts.hasOwnProperty('port')) {
-            agentOptions.port = typeof opts.port === 'string' ? parseInt(opts.port, 10) : opts.port;
-        }
 
         opts.agent = new https.Agent(agentOptions);
     }
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,6 +7,7 @@ import path = require('path');
 
 import shelljs = require('shelljs');
 
+import { Headers, RequestInit } from 'node-fetch';
 import * as api from './api';
 import { Authenticator } from './auth';
 import { AzureAuth } from './azure_auth';
@@ -144,6 +145,26 @@ export class KubeConfig implements SecurityAuthentication {
         const rootDirectory = path.dirname(file);
         this.loadFromString(fs.readFileSync(file, 'utf8'), opts);
         this.makePathsAbsolute(rootDirectory);
+    }
+
+    public async applytoFetchOptions(opts: https.RequestOptions): Promise<RequestInit> {
+        await this.applytoHTTPSOptions(opts);
+        const headers = new Headers();
+        for (const [key, val] of Object.entries(opts.headers || {})) {
+            if (Array.isArray(val)) {
+                val.forEach((innerVal) => {
+                    headers.append(key, innerVal);
+                });
+            } else if (typeof val === 'number' || typeof val === 'string') {
+                headers.set(key, val.toString());
+            }
+        }
+        return {
+            agent: opts.agent,
+            headers,
+            method: opts.method,
+            timeout: opts.timeout,
+        };
     }
 
     public async applytoHTTPSOptions(opts: https.RequestOptions): Promise<void> {
--- a/src/config.ts
+++ b/src/config.ts
@@ -159,6 +159,9 @@ export class KubeConfig implements SecurityAuthentication {
                 headers.set(key, val.toString());
             }
         }
+        if (opts.auth) {
+            headers.set('Authorization', 'Basic ' + Buffer.from(opts.auth).toString('base64'));
+        }
         return {
             agent: opts.agent,
             headers,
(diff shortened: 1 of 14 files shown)


src/watch_test.ts
```diff
@@ -34,81 +33,26 @@ const fakeConfig: {
...
-    pipe(stream: Duplex): void {}
-    abort() {}
-}
+const systemUnderTest = (): [nock.Scope] => {
+    const scope = nock(fakeConfig.clusters[0].server);
+    return [scope];
+};
 
 describe('Watch', () => {
-    describe('DefaultRequest', () => {
```
brendandburns, line 44: I'm a little worried that we're deleting these two tests, are they really not needed?

src/watch_test.ts
```diff
@@ -238,90 +211,28 @@ describe('Watch', () => {
...
 
         expect(receivedTypes).to.deep.equal([obj1.type]);
         expect(receivedObjects).to.deep.equal([obj1.object]);
 
-        expect(doneCalled).to.equal(true);
         expect(doneErr.length).to.equal(1);
         expect(doneErr[0]).to.deep.equal(errIn);
-    });
-
-    it.skip('should handle server side close correctly', async () => {
```
brendandburns, line 263: Worried that we're removing this test too.

src/watch_test.ts
```diff
@@ -238,90 +211,28 @@ describe('Watch', () => {
...
             },
         );
 
-        verify(fakeRequestor.webRequest(anything()));
-
-        const [opts] = capture(fakeRequestor.webRequest).last();
-        const reqOpts: request.OptionsWithUri = opts as request.OptionsWithUri;
+        stream.emit('error', errIn);
 
-        expect(reqOpts.uri).to.equal(`${server}${path}`);
```
brendandburns, line 251: Are there equiavelent tests we could do for fetch?

brendandburns:
Generally this looks ok (and thanks for the PR!) but I'm a little bit worried that the test coverage is being reduced. Is there a way to maintain the test coverage?

src/watch_test.ts
rbnayax: The fact that the test passes, means the correct uri was used, as this is the URL nock will listen on. If you change the URL for some reason, you will get an error from nock

src/watch_test.ts
rbnayax: those 2 are not relevant anymore as we don't use the DefaultRequest class anymore

src/watch_test.ts
rbnayax: my bad, I readded but it is now skipped due to the node-fetch bug I mentioned (at least I think). To properly close a stream, one needs to call `destroy` which other tests are already performing and it works

rbnayax:
> Generally this looks ok (and thanks for the PR!) but I'm a little bit worried that the test coverage is being reduced. Is there a way to maintain the test coverage?

I did everything I can to test what is possible. If you can suggest other tests I will gladly implement

src/config.ts
```diff
@@ -146,6 +147,26 @@ export class KubeConfig implements SecurityAuthentication {
         this.makePathsAbsolute(rootDirectory);
     }
 
+    public async applytoFetchOptions(opts: https.RequestOptions): Promise<RequestInit> {
```
davidgamero, line 150: could we just merge this with applyToHTTPSOptions()?

src/watch_test.ts
```diff
@@ -34,81 +33,26 @@ const fakeConfig: {
...
-            const req = defaultRequest.webRequest({
-                uri: `http://${server}/testURI`,
-            });
-            req.on('error', done);
-            (<any>req).emit('response', {
-                statusCode: 200,
-            });
-        });
-
-        it('should handle non-200 error codes', (done) => {
```
davidgamero, line 63: i think we could convert this test to still validate this chunk of error wrapping by emitting a non-200 code and catching it on the stream?:
```
await fetch(watchURL, requestInit)
            .then((response) => {
                if (response.status === 200) {
                    response.body.on('error', doneCallOnce);
                    response.body.pipe(stream);
                } else {
                    const error = new Error(response.statusText) as Error & {
                        statusCode: number | undefined;
                    };
                    error.statusCode = response.status;
                    throw error;
                }
            })
            .catch(doneCallOnce);
```
i see you have it in other tests, but this one was isolated to just that functionality

src/config.ts
rbnayax: We can not, as `applyToHTTPSOptions` is being used by proto-client

src/log.ts
```diff
@@ -92,19 +92,18 @@ export class Log {
...
         stream: Writable,
         doneOrOptions?: ((err: any) => void) | LogOptions,
         options?: LogOptions,
-    ): Promise<request.Request> {
-        let done: (err: any) => void = () => undefined;
-        if (typeof doneOrOptions === 'function') {
-            done = doneOrOptions;
-        } else {
+    ): Promise<AbortController> {
+        const AbortControllerCtor = globalThis.AbortController || (await import('abort-controller'));
```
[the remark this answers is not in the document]
rbnayax: AbortController exists natively since node since v15, so we can use it, or the "shim"  in previous version

src/watch_test.ts
rbnayax: test `should handle error from request stream` does exactly that, unless I'm missing something here

src/watch_test.ts
rbnayax: I fixed the test

src/config_test.ts
```diff
@@ -237,6 +238,41 @@
...
+                    string: 'str',
+                    empty: undefined,
+                    list: ['a', 'b'],
+                },
+            };
+            const requestInit = await kc.applytoFetchOptions(opts);
+            const expectedCA = Buffer.from('CADATA2', 'utf-8');
+            const expectedAgent = new https.Agent({
+                ca: expectedCA,
+                rejectUnauthorized: false,
```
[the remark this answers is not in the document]
rbnayax: this is a test... not sure how to ignore that

rbnayax:
> Generally this looks ok (and thanks for the PR!) but I'm a little bit worried that the test coverage is being reduced. Is there a way to maintain the test coverage?

I managed to find a way to maintain all existing tests so coverage is the same!

brendandburns:
/lgtm
/approve

Thanks for the patience!

PR #1037, opened 2023-04-11
fix: casing of `applyToHttpsOptions`

mstruebing:
Technically this is a breaking change but I  guess a `fix` is fine for these kind of change?


brendandburns:
/lgtm
/approve

I bet no one else is actually calling this, so I'm ok with it.

Issue #893, opened 2022-10-18
[release-1.x] Patch namespace secret trigger an error due to unsupported media type


JBWatenbergScality:
**Describe the bug**

When trying to use `patchNamespacedSecret` method on release-1.x branch the following error is raised:

```
Error: None of the given media types are supported: application/json-patch+json, application/merge-patch+json, application/strategic-merge-patch+json, application/apply-patch+yaml
    at ObjectSerializer.getPreferredMediaType (ObjectSerializer.js:1760:1)
    at CoreV1ApiRequestFactory.patchNamespacedSecret (CoreV1Api.js:8653:1)
    at ObservableCoreV1Api.patchNamespacedSecret (ObservableAPI.js:9211:1)
    at ObjectCoreV1Api.patchNamespacedSecret (ObjectParamAPI.js:2661:1)
   //...
```

** Client Version **
release-1.x

** Server Version **
1.24.3

**To Reproduce**
Sample code triggering the error:

```js
const coreV1 = new CoreV1Api(config);
coreV1.patchNamespacedSecret(
        {
          namespace: "default",
          name: "my-secret",
          body: {"tls.crt": "newValue"},
        }
      )
```

Also kind of related to this issue, I think we are missing a way to instruct `patchNamespacedSecret` which patch strategy we expect it to use. For example in the above example I expect it to use `application/merge-patch+json` but have no way right now to provide the patch strategy.

brendandburns:
cc @davidgamero

davidgamero:
@JBWatenbergScality thanks for letting us know, did you have this working on the 0.x version prior?
I think you are correct that it would require a custom `Content-Type` header, let me check if we have a supported pattern for this.

davidgamero:
From a quick look through it seems like some of the media-types are missing in this [list](https://github.com/kubernetes-client/javascript/blob/912ceb5fca091959d0443330ec9fdb5f002fe3d8/src/gen/models/ObjectSerializer.ts#L1065), but you may be able to supply one through the `options` parameter or using the [customObjectsAPII](https://github.com/kubernetes-client/javascript/blob/912ceb5fca091959d0443330ec9fdb5f002fe3d8/src/gen/apis/CustomObjectsApi.ts)

vpeltola:
I had a similar problem with patching cronjobs and I was able to add headers via the options argument and get it working.  For example:
```
      const result = await batchV1Api.patchNamespacedCronJob(
[...]

PR #1074, opened 2023-05-04
chore: update examples node fetch

clintonmedbery:
This fixes a few of the examples in the `FETCH_MIGRATION` markdown file. The ones I have checked off I have tested. I have also updated a few more I will include in another PR that I have not been able to successfully test because of this issue:
https://github.com/kubernetes-client/javascript/issues/893

I also added the typescript examples to the document, as they will need to be upgraded as well.

### DIFF
--- a/FETCH_MIGRATION.md
+++ b/FETCH_MIGRATION.md
@@ -52,28 +52,40 @@ Code will be on the `master` branch.
 -   [x] Fix errors in /src folder (due to new api)
     -   [x] migrate src/auth.ts, the dependent implementations (ex: azure_auth, gcp_auth etc) and tests to fetch api from request
     -   [x] migrate src/log.ts and its tests to fetch api from request
-        -  [x] major remaining work is fixing up async signatures and return piping
+        -   [x] major remaining work is fixing up async signatures and return piping
     -   [x] migrate src/watch.ts and its tests to fetch api from request
-        -  [x] remove decprecated requestImpl and RequestInterface
-        -  [x] implement queryParams parameter in watch method by injecting them into the fetch call
-        -  [x] update tests in src/watch_test.ts
+        -   [x] remove decprecated requestImpl and RequestInterface
+        -   [x] implement queryParams parameter in watch method by injecting them into the fetch call
+        -   [x] update tests in src/watch_test.ts
 -   [x] Fix errors in test (due to new api)
 -   [ ] Test all features
--   [ ] Fix examples and validate their param signatures (due to new api)
+-   [ ] Fix JavaScript examples and validate their param signatures (due to new api)
 
-    -   [ ] cache-example
-    -   [ ] example
-    -   [ ] follow-logs
+    -   [x] cache-example
+    -   [x] example
+    -   [x] follow-logs
     -   [ ] in-cluster-create-job-from-cronjob
-    -   [ ] in-cluster
-    -   [ ] ingress
+    -   [x] in-cluster
+    -   [x] ingress
     -   [ ] namespace
     -   [ ] patch-example
     -   [ ] raw-example (note: uses request lib directly, will require full fetch migration not just client param swap)
     -   [ ] scale-deployment
-    -   [ ] top_pods
-    -   [ ] top
+    -   [x] top_pods
+    -   [x] top
     -   [ ] yaml-example
+
+-   [ ] Fix TypeScript examples and validate their param signatures (due to new api)
+
+    -   [ ] apply-example
+    -   [ ] attach-example
+    -   [ ] cp-example
+    -   [ ] exec-example
+    -   [ ] informer-with-label-selector
+    -   [ ] informer
+    -   [ ] port-forward
+    -   [ ] example
+    -   [ ] watch-example
 
 -   [ ] Update docs
     -   [ ] Update README examples
--- a/examples/cache-example.js
+++ b/examples/cache-example.js
@@ -1,4 +1,5 @@
-const k8s = require('@kubernetes/client-node');
+// in a real program use require('@kubernetes/client-node')
+const k8s = require('../dist/index');
 
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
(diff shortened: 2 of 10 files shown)


src/cache.ts
```diff
@@ -25,7 +25,7 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
     private objects: T[] = [];
     private resourceVersion: string;
     private readonly indexCache: { [key: string]: T[] } = {};
-    private readonly callbackCache: { [key: string]: (ObjectCallback<T> | ErrorCallback)[] } = {};
+    private readonly callbackCache: { [key: string]: Array<ObjectCallback<T> | ErrorCallback> } = {};
```
mstruebing, line 28: Could you explain why you need to make these changes in this file and the corresponding test?

examples/in-cluster.js
```diff
@@ -5,11 +5,11 @@ kc.loadFromCluster();
 
 const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
 
-k8sApi.listNamespacedPod('default')
+k8sApi
+    .listNamespacedPod({ namespace: 'default' })
     .then((res) => {
-	console.log(res.body);
+        console.log(res);
```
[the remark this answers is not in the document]
clintonmedbery: I didn't. That is what this function is now returning. This example now reflects that.

examples/example.js
```diff
@@ -5,8 +5,11 @@ kc.loadFromDefault();
...
 
-k8sApi.listNamespacedPod('default')
+k8sApi
+    .listNamespacedPod({ namespace: 'default' })
     .then((res) => {
-	console.log(res.body);
+        console.log(res);
+    })
+    .catch((err) => {
+        console.log(err);
```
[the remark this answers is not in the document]
clintonmedbery: I am just matching what the example was doing, but I can update that.

FETCH_MIGRATION.md
```diff
@@ -52,29 +52,41 @@ Code will be on the `master` branch.
...
     -   [ ] patch-example
     -   [ ] raw-example (note: uses request lib directly, will require full fetch migration not just client param swap)
     -   [ ] scale-deployment
-    -   [ ] top_pods
-    -   [ ] top
+    -   [x] top_pods
+    -   [x] top
     -   [ ] yaml-example
 
+-   [ ] Fix TypeScript examples and validate their param signatures (due to new api)
```
clintonmedbery, line 78: We didn't account for the TypeScript examples before.

examples/cache-example.js
```diff
@@ -8,7 +8,7 @@ const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
 const path = '/api/v1/pods';
 const watch = new k8s.Watch(kc);
 
-const listFn = () => k8sApi.listPodForAllNamespaces()
+const listFn = () => k8sApi.listPodForAllNamespaces();
```
clintonmedbery, line 11: Some of these examples got formatted with prettier.

examples/ingress.js
```diff
@@ -1,27 +1,39 @@
-const k8s = require('@kubernetes/client-node')
-const kc = new k8s.KubeConfig()
-kc.loadFromDefault()
+const k8s = require('@kubernetes/client-node');
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
 
-const k8sApi = kc.makeApiClient(k8s.NetworkingV1beta1Api) // before 1.14 use extensions/v1beta1
-const clientIdentifier = 'my-subdomain'
+const k8sApi = kc.makeApiClient(k8s.NetworkingV1Api); // before 1.14 use extensions/v1beta1
```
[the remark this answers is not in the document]
clintonmedbery: I think I will just remove it as it seems outdated.

src/cache.ts
clintonmedbery: Functions like `k8sApi.listPodForAllNamespaces()` no longer return the HTTP status and body. They return just the KubernetesObjects. I updated ListPromise to reflect this and updated the tests. Let me know if this is out of line, but the example and tests do not work otherwise.

clintonmedbery: Oh, and the Array updates are linting fixes.

examples/ingress.js
```diff
@@ -1,27 +1,39 @@
...
-      }
-    }],
-    tls: [{ hosts: [`${clientIdentifier}.example.com`] }]
-  }
-}).catch(e => console.log(e))
+k8sApi
+    .createNamespacedIngress({
+        namespace: 'default',
+        body: {
+            apiVersion: 'networking.k8s.io/v1beta1',
```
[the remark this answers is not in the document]
clintonmedbery: Good catch.

mstruebing (approved):
👍 LGTM

examples/top_pods.js
```diff
@@ -1,39 +1,35 @@
-const k8s = require('../dist/index');
+const k8s = require('@kubernetes/client-node');
```
brendandburns, line 1: this needs to be removed b/c otherwise we are testing against the published package vs `HEAD`

brendandburns:
One minor comment.

examples/top_pods.js
clintonmedbery: Most of the examples use this. I will updated the others to use `HEAD` to match your comment.

brendandburns: Now I'm actually conflicted :) Perhaps it is better to have a working example? The trouble with not using `HEAD` is that if we change the API then we don't have continuous validation that the examples are working...

Maybe we can add a comment to each of these that says:
`// in a real program use require('@kubernetes/client-node')`

or something like that.

clintonmedbery: Ok, the rest of these seem to use `require('@kubernetes/client-node');`. I will make them use `require('../dist/index');` but maybe it would be helpful to have a comment on how to import the package? Or maybe the readme does that already. Thinking out loud here. For now I will just import them.

brendandburns:
/lgtm
/approve

Thanks!

PR #1082, opened 2023-05-09
chore: update js and typescript examples

clintonmedbery:
This is part two of updating examples to node-fetch. A lot of these examples do not work do to this issue (https://github.com/kubernetes-client/javascript/issues/893) but I have updated the syntax to match what we need with fetch. Other examples do work. Let me know if you would like me to separate these out. 

Right now I just put a note in the `FETCH_MIGRATION` file. If you have another idea to best handle this then let me know and I will update.

### DIFF
--- a/examples/in-cluster-create-job-from-cronjob.js
+++ b/examples/in-cluster-create-job-from-cronjob.js
@@ -1,12 +1,13 @@
-const k8s = require('@kubernetes/client-node');
+// in a real program use require('@kubernetes/client-node')
+const k8s = require('../dist/index');
 
 const kc = new k8s.KubeConfig();
 kc.loadFromCluster();
 
 const batchV1Api = kc.makeApiClient(k8s.BatchV1Api);
 const batchV1beta1Api = kc.makeApiClient(k8s.BatchV1beta1Api);
-const cronJobName = 'myCronJob';
-const jobName = 'myJob';
+const cronJobName = 'cronjob';
+const jobName = 'myjob';
 
 const job = new k8s.V1Job();
 const metadata = new k8s.V1ObjectMeta();
@@ -18,12 +19,14 @@ metadata.annotations = {
 };
 job.metadata = metadata;
 
-batchV1beta1Api.readNamespacedCronJob(cronJobName, 'default')
+batchV1beta1Api
+    .readNamespacedCronJob({ name: cronJobName, namespace: 'default' })
     .then((cronJobRes) => {
-        job.spec = cronJobRes.body.spec.jobTemplate.spec;
-        batchV1Api.createNamespacedJob('default', job)
+        job.spec = cronJobRes?.spec?.jobTemplate.spec;
+        batchV1Api
+            .createNamespacedJob({ namespace: 'default', body: job })
             .then((res) => {
-                console.log(res.body);
+                console.log(res);
             })
             .catch((err) => {
                 console.log(err);
--- a/examples/raw-example.js
+++ b/examples/raw-example.js
@@ -1,20 +1,37 @@
-const k8s = require('@kubernetes/client-node');
-const request = require('request');
+// in a real program use require('@kubernetes/client-node')
+const k8s = require('../dist/index');
+const fetch = require('node-fetch');
+const https = require('https');
 
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
-const opts = {};
-kc.applyToRequest(opts);
+const currentUser = kc.getCurrentUser();
+const currentCluster = kc.getCurrentCluster();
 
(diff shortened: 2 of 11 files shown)


examples/in-cluster-create-job-from-cronjob.js
```diff
@@ -1,12 +1,13 @@
-const k8s = require('@kubernetes/client-node');
+// in a real program use require('@kubernetes/client-node')
+const k8s = require('../dist/index');
```
clintonmedbery, line 2: I did this to match the previous PR.

examples/raw-example.js
```diff
@@ -1,20 +1,49 @@
...
-  });
+const kubeconfig = yaml.load(fs.readFileSync(KUBECONFIG, 'utf-8'));
+console.log(kubeconfig);
+if (!kubeconfig || !kubeconfig.users || !kubeconfig.clusters || !kubeconfig['current-context']) {
+    throw Error('please give me a valid KUBECONFIG file to work with.');
+}
+const cluster = kubeconfig.clusters.find((c) => c.name === kubeconfig['current-context']).cluster;
+const user = kubeconfig.users.find((u) => u.name === kubeconfig['current-context']).user;
+
+const agent = new https.Agent({
```
clintonmedbery, line 24: If you have a better way to authenticate then LMK!

examples/raw-example.js
```diff
@@ -1,20 +1,49 @@
...
-        if (error) {
-            console.log(`error: ${error}`);
-        }
-        if (response) {
-            console.log(`statusCode: ${response.statusCode}`);
-        }
-        console.log(`body: ${body}`);
-  });
+const kubeconfig = yaml.load(fs.readFileSync(KUBECONFIG, 'utf-8'));
+console.log(kubeconfig);
```
mstruebing, line 17: Can we remove this `console.log`?

examples/raw-example.js
mstruebing: Did you try using `applyToHTTPSOptions` or `applyToFetchOptions`? 
I'm not sure it will work but I guess this should be the equivalent of `applyToRequest`.

* https://github.com/kubernetes-client/javascript/blob/release-1.x/src/config.ts#L150
* https://github.com/kubernetes-client/javascript/blob/release-1.x/src/config.ts#L173

examples/scale-deployment.js
```diff
@@ -9,16 +10,31 @@ const targetNamespaceName = 'default';
 const targetDeploymentName = 'docker-test-deployment';
 const numberOfTargetReplicas = 3;
 
-async function scale(namespace, name, replicas) {
-  // find the particular deployment
-  const res = await k8sApi.readNamespacedDeployment(name, namespace);
-  let deployment = res.body;
+async function scale(deploymentNamespace, deploymentName, replicas) {
```
mstruebing, line 13: Can we stick to`namespace` and `name` here instead of `deploymentNamespace` and `deploymentName`?

examples/typescript/informer/informer-with-label-selector.ts
```diff
@@ -34,25 +32,20 @@ const createPod = async (name, app) => {
...
         },
     } as k8s.V1Pod;
-    await k8sApi.createNamespacedPod('default', appPod).catch((e) => console.error(e));
+    await k8sApi.createNamespacedPod({ namespace: 'default', body: appPod }).catch((e) => console.error(e));
     console.log('create', name);
 };
 
-const deletePod = async (name, namespace) => {
-    await k8sApi.deleteNamespacedPod(name, namespace);
+const deletePod = async (podName: string, podNamespace: string) => {
```
mstruebing, line 39: `name`, `namespace`?

examples/raw-example.js
clintonmedbery: I did, but unsuccessfully. I might go back and try again. It wanted some context that I could get from the library. Not saying it can't be done though...

examples/scale-deployment.js
clintonmedbery: The reason I did this was because as an example, I think it is easier to read the named parameters as `{ name: variableName, namespace: anotherName }` rather than using object shorthand property names like `{ name, namespace }`. In general I prefer what you are saying, but as an example I find this more easier to digest.

examples/typescript/informer/informer-with-label-selector.ts
clintonmedbery: Same reason above.

examples/raw-example.js
clintonmedbery: Yes! 🤦

examples/raw-example.js
clintonmedbery: I figured it out! Thanks for the comment! Made me go back and try again 🥇

Issue #1075, opened 2023-05-04
[release-1.x] `getPreferredMediaType` doesn't handle empty arrays resulting in `None of the given media types are supported:`


per2plex:
**Describe the bug**

`getPreferredMediaType` only checks if `mediaTypes` is falsy and not if it's an empty array:
https://github.com/kubernetes-client/javascript/blob/release-1.x/src/gen/models/ObjectSerializer.ts#L1745-L1749

The API provides an empty array, e.g:
https://github.com/kubernetes-client/javascript/blob/release-1.x/src/gen/apis/CoreV1Api.ts#L2779

If the TypeScript types can be trusted, `mediaTypes` should never be `undefined` anyway.

I didn't know where to provide a fix, because those files are generated?

Might be related to https://github.com/kubernetes-client/javascript/issues/893

**Client Version**
`1.0.0-rc2`

**Server Version**
 `1.24.10`

**To Reproduce**

- Try creating a namespace via API
- Throws `None of the given media types are supported:`

**Expected behavior**
Not throw `None of the given media types are supported:`

**Example Code**
```
const kc = new k8s.KubeConfig()

kc.loadFromDefault()

const k8sApi = kc.makeApiClient(k8s.CoreV1Api)

await k8sApi.createNamespace({
  body: {
    metadata: {
      name: 'my-namespace',
    },
  },
})
```

felix-gohla:
Maybe this is rather a problem with either a) the OpenAPI specs over at the Kubernetes repo or b) in the OpenAPI generator.
The place to fix for the latter one would be here (I did not find any issue regarding that in the OpenAPI generator repo):

https://github.com/OpenAPITools/openapi-generator/blob/cc620d8ba22feedff4a2a4e4ef7c3a69a98eacff/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache#L198-L201

Another potential fix was in the `kubernetes-client/gen` repo by preprocessing the swagger documentation replacing all `"consumes": ["*/*"]` with `"consumes": ["application/json"]`. But I think fixing the generator would be better.

davidgamero:
after taking a look, I think you're correct. The current generated code doesn't respect the `*/*` media type as a pattern.
This is most likely best addressed with an upstream PR to add support.

I can take a shot at it this week. I don't know if it makes more sense to check explicitly for an `allow all` pattern or to use the existing default of an empty array (this could end up hiding error depending on the implementation`

[...]

PR #1120, opened 2023-06-09
[release-1.x] Fix media type error

WassimBenzarti:
Whenever calling createNamespacedSecret or deleteNamespacedSecret, there is an error occurs like the following:
```
Error: None of the given media types are supported: 
<Stack trace>
```
The error shows that there are no `mediaTypes` to the `getPreferredMediaType` function. After some digging, turns out that there is a condition that wasn't taking into account the fact that the array can be empty (not undefined) which in this case should default to `application/json`.

### DIFF
--- a/src/gen/models/ObjectSerializer.ts
+++ b/src/gen/models/ObjectSerializer.ts
@@ -1052,27 +1052,27 @@ import { VersionInfo } from '../models/VersionInfo';
 
 /* tslint:disable:no-unused-variable */
 let primitives = [
-                    "string",
-                    "boolean",
-                    "double",
-                    "integer",
-                    "long",
-                    "float",
-                    "number",
-                    "any"
-                 ];
+    "string",
+    "boolean",
+    "double",
+    "integer",
+    "long",
+    "float",
+    "number",
+    "any"
+];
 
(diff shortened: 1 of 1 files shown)


WassimBenzarti:
After having a look at [this](https://github.com/kubernetes-client/javascript/issues/1075), it seems that we need to change the upstream repo and regenerate everything to fix this.

Am I right?

brendandburns:
Yes, the change really needs to be made in the upstream code generator, and then we can regenerate.

If you wanted to make that change that would be awesome, the generator is here:
https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript/model/ObjectSerializer.mustache#L197

Thanks

felix-gohla:
There is a related issue: <https://github.com/OpenAPITools/openapi-generator/issues/15011>

mstruebing:
I've created a PR to start things to get moving: https://github.com/OpenAPITools/openapi-generator/pull/15833

mstruebing:
Sorry @WassimBenzarti I went ahead and created a PR including the openapi-generator fix: #1127

WassimBenzarti:
No problem thank you!

brendandburns:
Closing in favor of #1127

PR #1127, opened 2023-06-17
Regenerate client & test more example

### LINKED ISSUE #1075: [release-1.x] `getPreferredMediaType` doesn't handle empty arrays resulting in `None of the given media types are supported:`

mstruebing:
* Updated setting to match newer gen and openapi-generator versions
* Regenerated the client with https://github.com/OpenAPITools/openapi-generator/pull/15833 applied
* Fixed a type issue after regenerating the client
* tested more examples and updated `FETCH_MIGRATION.md` accordingly, updated examples to easier change related properties (namespace, podnames, etc.)

Can be reviewed as individual commits if the full diff is to big.

* Closes #1075  
* Closes #1120 

FYI: I created an upstream PR in `node-fetch` but I do not know if this will get accepted: https://github.com/node-fetch/node-fetch/pull/1757


brendandburns:
Nice, thank you!

/lgtm
/approve

PR #1134, opened 2023-06-21
Update node typings, remove 'abort-controller' and remove node 14

brendandburns:
Updating node typings added a definition for `AbortController` and `AbortSignal` this was added in node 15+.
Node 14 is EOL anyway, so remove it from support.

Cleanup the code a little b/c of some differences between Abort* definitions and also ClientOptions.


package.json
```diff
@@ -55,13 +55,12 @@
     "license": "Apache-2.0",
     "dependencies": {
         "@types/js-yaml": "^4.0.1",
-        "@types/node": "^10.12.0",
+        "@types/node": "^20.3.1",
```
mstruebing, line 58: That's a huge bump.

mstruebing (approved):

package.json
brendandburns: yeah, this is just the typings though so it doesn't really affect anything except compilation. Seems to be ok (aside from the fact that we'll no longer support node 14)

brendandburns: fwiw, this is also what we have in the `main` branch, in `release-1.x` we didn't have dependabot turned on, so it didn't get updated for a long time, and then there were some compilation fixes that were needed.

PR #1143, opened 2023-06-27
Remove execa and shelljs

brendandburns:
Similar to #946 but for the 1.x branch.

cc @mstruebing


mstruebing (approved):
Nice, I can not see any issues with your changes 👍

PR #1158, opened 2023-07-07
Remove underscore js

DavidIsa:
This pull request proposes the removal of the Underscore.js library from the project. After careful consideration and evaluation of the project's requirements, it has been determined that removing underscore.js will bring several benefits.
Underscore js is an external dependency that needs to be maintained and updated alongside the project. By removing the library, we reduce the number of dependencies we need to manage, simplify the maintenance process, and potentially avoid compatibility issues or security vulnerabilities. Including Underscore.js in the project adds unnecessary file size and overhead, as it introduces an extra dependency that needs to be loaded by the application. Removing Underscore.js will help reduce the overall size of the project and potentially improve performance, particularly if a significant portion of its features is not utilized.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,14 +17,12 @@
         "isomorphic-ws": "^5.0.0",
         "js-yaml": "^4.1.0",
         "jsonpath-plus": "^7.2.0",
-        "openid-client": "^5.3.0",
         "request": "^2.88.0",
         "rfc4648": "^1.3.0",
         "stream-buffers": "^3.0.2",
         "tar": "^6.1.11",
         "tmp-promise": "^3.0.2",
         "tslib": "^2.4.1",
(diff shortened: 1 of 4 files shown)


mstruebing (approved):
Thank you, removing deps is always a good idea I think and in this case we use the features of underscore very little.

brendandburns:
Thanks for the PR. It looks like tests are failing. You can run the tests with `npm run test`

DavidIsa:
Thanks for the PR review!

I've pushed a new commit with the fix

brendandburns:
Sorry, looks like formatting failed now :(

You can run `npm run format` and `npm run lint`

Thanks!

mstruebing:
Could you still run `npm run format` and commit the changes?

DavidIsa:
Done

PR #1164, opened 2023-07-12
Examples improvements

DavidIsa:
I suggest some improvements to the examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,17 @@ kc.loadFromDefault();
 
 const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
 
-k8sApi.listNamespacedPod('default').then((res) => {
-    console.log(res.body);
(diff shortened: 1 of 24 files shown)


brendandburns:
Thanks for the PR

/lgtm
/approve

arimgibson:
For whatever reason, the watch example no longer functions after this change. The code looks fine to me as well, but the watch method no longer reports anything nor does it call the `done` function after running.

The example before this PR works as expected. I'm happy to look into it more next week when I have some availability if the fix isn't clear before then. Also happy to open a new issue if that makes more sense too; again it might have to be next week though to make sure I can complete it with all the details!

arimgibson:
Nvm, super simple fix. The call to `watch.watch()` shouldn't be `await`ed. Submitting a quick PR.

PR #1169, opened 2023-07-17
Remove tmp-promise

DavidIsa:
Removing unnecessary dependency tmp-promise


src/cp.ts
```diff
@@ -65,7 +64,7 @@ export class Cp {
      * @param {string} containerName - The name of the container in the pod to exec the command inside.
      * @param {string} srcPath - The source path in local
      * @param {string} tgtPath - The target path in the pod
-     * @param {string} cwd - The directory that is used as the parent in the host when uploading
+     * @param {string} [cwd] - The directory that is used as the parent in the host when uploading
```
mstruebing, line 67: Why did you change the `cwd` to be in square brackets?
Doesn't that mean it's an array of strings while it isn't?

mstruebing (commented):
Nice change, but could we use this function instead of writing our own with a UUID? 
[fsPromises.mkdtemp](https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fspromises_mkdtemp_prefix_options) It's available in node 14 and according to our workflow, this is the oldest node version we are supporting in `master`.

src/cp.ts
DavidIsa: When a param name is in the square brackets, it means that it is optional

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

DavidIsa:
[fsPromises.mkdtemp](https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fspromises_mkdtemp_prefix_options) creates a temporarily directory, but in the code we need to create temporarily file.

src/cp.ts
mstruebing: Thank you I did not know this.

mstruebing:
You are right, I'm sorry I misunderstood the code. 👍

mstruebing (approved):

brendandburns:
One small nit, I know it is _very_ unlikely that the file already exists, but can we do a check to make sure (and generate a new random name if it does)

PR #1214, opened 2023-08-21
Update generate-javascript.yml

brendandburns:
Another cherry-pick to get this working again.


brendandburns:
Yeah, the main problem is that the `kubernetes-client/gen` repo got updated to support `fetch` instead of creating a branch. So we're trying to support two different generators after the same repo.

I probably should just create a branch rather than cherry-pick, but for now this is kind of a manual branch.

PR #1221, opened 2023-08-24
Add a spec to the Kubernetes Object type

### LINKED ISSUE #1218: KubernetesObject type should have a spec field
**Describe the bug**
This type here: https://github.com/kubernetes-client/javascript/blob/6305ecd4308543407910f8db5529f3639bd758c5/src/types.ts#L2C1-L8C1

It should have a `spec?: object` property on it that is compatible with all derived object's spec type.

** Client Version **
0.18.1

** Server Version **
1.24.1

**To Reproduce**
N/A

**Expected behavior**
A `spec?: object` property added. I've extended it in my own codebase and added the property, I can't see a reason consumers should be expected to do that.

** Example Code**
```typescript
import type { KubernetesObject } from '@kubernetes/client-node';

export interface CustomKubernetesObject extends KubernetesObject {
  spec?: object;
}
```

 - OS: macOS
 - NodeJS Versio: 18
 - Cloud runtime: Docker (Debian Bullseye)

bencooper222:
Adding an optional spec property to this object brings it into alignment with actual usage of Kubernetes objects. 

Fixes #1218


brendandburns:
@bencooper222 thanks for the PR. While most Kubernetes resources do have a Spec field, there are a number of others (ConfigMap is probably the most obvious example) which do not.

I think we could introduce a `KubernetesObjectWithSpec` if that would be useful, but I don't think that we can/should change KubernetesObject in this way.

bencooper222:
Thank you for the quick response! I've revised the PR to match your suggestion.

.npmrc
```diff
@@ -0,0 +1,2 @@
+# for contributors with a corporate registry, this will make sure we use the public one
```
brendandburns, line 1: Please remove this file as it is unrelated to this change.

Issue #1253, opened 2023-09-07
chore: refactor examples dir 


Ananya2001-an:
**Describe the bug**
We could refactor the examples folder to make it more readable and maintainable. Like creating a folder called `kubectl/equivalents`  inside examples to give example for the corresponding kubectl command.

For example, creating `NamespaceCreateEquivalent.js` file which shows you the code equivalent to kubectl command: `kubectl create namespace test`

** Client Version **
e.g. `0.18.1`

** Example Code**
PodsFilterByNamespace.js

```js
const k8s = require('@kubernetes/client-node');

const kc = new k8s.KubeConfig();
kc.loadFromDefault();

const k8sApi = kc.makeApiClient(k8s.CoreV1Api);

/*
This sample javascript code is equivalent to `kubectl get pod --namespace=demo-namespace`
*/

const main = async () => {
    try {
        const podRes = await k8sApi.listNamespacedPod('demo-namespace');
        console.log('Pod: ', podRes.body.items[0].metadata);
     } catch (err) {
         console.error(err);
     }
};

main();

```

**Additional context**
I have already started working on this and would love to get assigned for the same.

PR #1254, opened 2023-09-08
chore: add kubectl equivalent examples

Ananya2001-an:
Related to #1253 
Created a `kubectl/equivalents` folder inside examples. Each file corresponds to the relevant kubectl command making it easier for users to understand the usage of the client. Makes it more readable and maintainable.

Current files added:
- Create namespace
- Create Namespace from yaml
- Create Namespaced pod
- List namespaces
- List Pods
- Create Resource quotas
- List all Resource quotas

... more to be added

### DIFF
--- a/examples/kubectl/equivalents/namespace-create-yaml.js
+++ b/examples/kubectl/equivalents/namespace-create-yaml.js
@@ -1,0 +1,24 @@
+const k8s = require('@kubernetes/client-node');
+const fs = require('fs');
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+
+/**
+ * This sample code is Javascript equivalent to `kubectl apply -f namespace.yml`
+ * You can find the sample namespace.yml file in the examples/resources folder
+ */
+
+const namespaceCreateYaml = async () => {
+    try {
+        const namespaceYaml = k8s.loadYaml(fs.readFileSync('../../resources/namespace.yml'));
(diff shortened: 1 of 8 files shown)


brendandburns:
Thanks for this PR! I think it would be even better to introduce a `kubectl.js` as part of this library, so that people can simply call the `kubectl` equivalents.

It would be similar to https://github.com/kubernetes-client/java/tree/master/extended/src/main/java/io/kubernetes/client/extended/kubectl

Ananya2001-an:
> Thanks for this PR! I think it would be even better to introduce a `kubectl.js` as part of this library, so that people can simply call the `kubectl` equivalents.

Yeah that sounds good! I will work on it. I think we can even add some tests for the same with jest, wdyt? But I would work on it in a separate PR to keep things simple for now :)

Ananya2001-an:
Hey @brendandburns I created a `Kubectl.ts` file taking reference from the java client. Though, I am not very sure about how to create the generic execution calls for each function. Would like some help with that :) I have pushed the changes for your convenience.

brendandburns:
Yes, you're right, there's no good equivalent of the generic client in Java. We should add that :)

If that's too much work, feel free to revert this PR just to the examples and we can merge those and work on the generic `kubectl` implementation some other time.

brendandburns:
/lgtm
/approve

Thanks!

Ananya2001-an:
Hey @brendandburns I would love to be a member of the kubernetes-client org and I was wondering how to proceed with it...like how one can apply for it? I know that one has to do some significant contribution work and I am willing to put that effort ;)

PR #1268, opened 2023-09-17
Add an additional test for version synchronization

brendandburns:


mstruebing (commented):
Not sure if that can be applied in this context (when/where the test runs) but wouldn't an `npm install` with a relativlely recent `npm` version and a `git diff` achieve the same with a bit more variety of possible mismatches/errors?

brendandburns:
I think there's a separate lint failure that got introduce somehow, I will try to fix it in a different PR.

mstruebing (approved):
tests are passing 🙌

PR #1312, opened 2023-10-05
docs: exec example

cdaringe:
# Problem

Cmd+arg was not clearly recorded in the examples, leading me to try zany things, like piping my cmds via stdin, but then the websocket closing prematurely

# Solution

Make the default example clear that string[] is a permissible input for a req/res lifecycle of exec, despite being over a ws

### DIFF
--- a/examples/typescript/exec/exec-example.ts
+++ b/examples/typescript/exec/exec-example.ts
@@ -1,8 +1,6 @@
 import * as k8s from '@kubernetes/client-node';
 import * as stream from 'stream';
 
-const command = process.argv[2];
-
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
 
@@ -11,7 +9,7 @@ exec.exec(
     'default',
     'nginx-4217019353-9gl4s',
     'nginx',
-    command,
+    ["ls, "-al", "."],
     process.stdout as stream.Writable,
     process.stderr as stream.Writable,
     process.stdin as stream.Readable,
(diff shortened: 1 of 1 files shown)


examples/typescript/exec/exec-example.ts
```diff
@@ -11,7 +9,7 @@ exec.exec(
     'default',
     'nginx-4217019353-9gl4s',
     'nginx',
-    command,
+    ["ls, "-al", "."],
```
cdaringe, line 12:
```suggestion
    ['ls, '-al', '.'],
```

examples/typescript/exec/exec-example.ts
```diff
@@ -11,7 +9,7 @@ exec.exec(
     'default',
     'nginx-4217019353-9gl4s',
     'nginx',
-    command,
+    ['ls, '-al', '.'],
```
mstruebing, line 12: What would happen if you run it with: `<typescript> ./examples/typescript/exec/exec-example.ts 'ls -al'` - that should work, right?

cdaringe: I was getting “bin not found” errors doing as you suggest/ask. I presume the string vs array is encoded differently in the resultant url

mstruebing: I see, I think I understand.
I think this is fine as this is only an example. Thank you.

mstruebing:
/lgtm
/approve

thank you

PR #1336, opened 2023-10-12
feat: `kubectl` class for directly calling kubectl equivalent commands

Ananya2001-an:
WIP
- need to work on generic K8s api implementation
- currently focusing on just the `get` method that should invoke the list method of `genericKubernetesApi` for all `apiType` (pods, namespaces, etc..)
- will add necessary tests once implemented

### Notes for reviewers
I have created a draft PR for ease of reviewing....would make necessary changes as per feedback :)
This will be implemented in a similar way to the java client as discussed here #1254

### DIFF
--- a/src/extended/kubectl/kubectl-get.ts
+++ b/src/extended/kubectl/kubectl-get.ts
@@ -1,0 +1,9 @@
+import { KubernetesObject } from '../../types';
+
+export class kubectlGet<ApiType extends KubernetesObject> {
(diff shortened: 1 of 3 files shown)


Ananya2001-an:
Hey @brendanburns I have started working out on this but I would still need some help with the generic implementation...I would really appreciate your guidance on this...like how I should go about implementing it :) Any kind of feedback would be great!

PR #1341, opened 2023-10-13
bump openapi version to support for k8s media types

davidgamero:
regenerate after bumping the open api version to include https://github.com/OpenAPITools/openapi-generator/pull/16386 

ref https://github.com/kubernetes-client/javascript/issues/754
this should allow the routes that are currently blocked by missing media types, and i've included an example for patch showing how it can be done

the next priority would ideally be a fully custom object serializer, or simply adding the ability to merge header maps in the `options` parameter.

For now, this will provide a way to get things working even if it's a bit verbose


OliverMKing (approved):

davidgamero:
this is weird we need more than 2gb of memory to run our tests now?

examples/typescript/patch/patch-example.ts
```diff
@@ -0,0 +1,83 @@
...
+                },
+            },
+        ];
+        const options = { headers: { 'Content-type': 'application/json-patch+json' } };
+        const headerPatchMiddleware = new PromiseMiddlewareWrapper({
+            pre: async (requestContext: RequestContext) => {
+                requestContext.setHeaderParam('Content-type', 'application/json-patch+json');
+                return requestContext;
+            },
+            post: async (responseContext: ResponseContext) => responseContext,
```
mstruebing, line 39: Hey, @OliverMKing, one question: 
is it required to define both `pre` as well as `post` even if you do not do anything else than return the arguments like in your example with `post`?

davidgamero: i [don't see any default implementations ](https://github.com/kubernetes-client/javascript/blob/ca53fef7f65f8544a2e98db7caaff2d0e92e0152/src/gen/middleware.ts#L28)of the pre/post methods, so they are most likely required for now,  but we could easily add a wrapper that does default them or even a wrapper that makes middleware/configuration from a desired MIME media type since that's most of what we're after here

davidgamero: you can skip the middleware altogether if you want to use [the first supported content type](https://github.com/davidgamero/javascript/blob/c17ff3193dbd111b9e5b5db939647af70005381e/src/gen/apis/CoreV1Api.ts#L10138C10-L10138C10) which is `application/json-patch+json`, but for any other types you'll have to add it

davidgamero:
i was able to locally repo this using the ubuntu 22.04 docker image and running the container with a 2gb memory limit, but when limiting node memory with `node --max_old_space_size=25 $(which npm) test` i was able to run the tests on as little as 25mb.

I'm continuing to test so we can get his merged

davidgamero:
testing this out for now https://github.com/actions/runner-images/issues/70#issuecomment-1191708172

davidgamero:
@mstruebing i just merged your changes from main and it looks like it works now, but i suspect one of the packages just had an issue that's somehow consuming extra CI memory

examples/patch-example.js
```diff
@@ -19,9 +21,36 @@ k8sApi.listNamespacedPod({ namespace }).then((res) => {
         },
     ];
     // TODO this method of passing the content type will change when we figure out a way to properly do this
-    const options = { headers: { 'Content-type': k8s.PatchUtils.PATCH_FORMAT_JSON_PATCH } };
+    const headerPatchMiddleware = new PromiseMiddlewareWrapper({
```
mstruebing, line 24: Can the comment above be removed?

examples/typescript/patch/patch-example.ts
```diff
@@ -0,0 +1,83 @@
...
+    RequestContext,
+    ResponseContext,
+    KubeConfig,
+    createConfiguration,
+    Configuration,
+    ServerConfiguration,
+} from '../../../dist';
+import { PromiseMiddlewareWrapper } from '../../../dist/gen/middleware';
+
+// const k8s = require('@kubernetes/client-node');
```
mstruebing, line 12: Could we put this at the top and a comment like in the other files i.e:

```
// in a real program use require('@kubernetes/client-node')
```

examples/typescript/patch/patch-example.ts
```diff
@@ -0,0 +1,83 @@
...
+            )
+            .then(() => {
+                // tslint:disable-next-line:no-console
+                console.log('Patched.');
+            })
+            .catch((err) => {
+                // tslint:disable-next-line:no-console
+                console.log('Error: ');
+                // tslint:disable-next-line:no-console
+                console.log(err);
```
mstruebing, line 81:
```suggestion
                console.log('Error: ', err);
```

settings
```diff
@@ -30,4 +30,4 @@ export CLIENT_VERSION="0.8-SNAPSHOT"
 # Name of the release package
 export PACKAGE_NAME="@kubernetes/node-client"
 
-export OPENAPI_GENERATOR_COMMIT=5555137b7926de2f501e091718ff72429b808220
+export OPENAPI_GENERATOR_COMMIT=ba367e60e85392a7fd58cdb374c35c3aaaece872
```
mstruebing, line 33: There is already a way newer commit, should we bother updating this in this PR again?

mstruebing (commented):
Hey @davidgamero some small comments but in general I'm happy with this PR :)

settings
davidgamero: oh do you mean bump all the way to the newest? i was just trying to advance enough to get the media type changes, but we can hop further ahead i can update it

mstruebing: Yes. At least if this doesn't introduce other problems.
Otherwise, We should merge it in the current state in my opinion.

brendandburns:
@mstruebing I'll let you merge this when ready.

davidgamero:
i was able to make some progress isolating the gc memory issue- maybe we make a separate issue to track it

by executing `nyc mocha` with the `--node-command inspect` to enable debugging i was able to connect and get a GC snapshot and allocation timeline

the majority of the growing heap usage seems to be coming from SourceMapConsumer in `node_modules/source-map/lib/source-map-consumer.js` at the bottom of a pretty long call stack that includes @ babel/core 

i found [this](https://github.com/meteor/meteor/issues/9568) potentially relevant link that has encouraging similarities.

by changing nyc.instrument=false in package.json on line 51, the memory never exceeded 600MB, while re-enabling that line eventually balloons up to 12GB which was the most i was willing to test. 

i will continue investigating upgrading our source-map indirect dependency so we can hopefully get this resolved

settings
davidgamero: agreed i've updated it to the newest hash now

mstruebing:
/lgtm
/approve

Thank you very much for this effort 🙏 
I agree we should track and fix the memory limit separately.

PR #1396, opened 2023-11-08
cpFromPod: close writerStream synchronously

fpm-peter:
This replicates https://github.com/kubernetes-client/javascript/pull/880 since https://github.com/kubernetes-client/javascript/pull/943 didn't solve the issue.

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -27,34 +27,37 @@ export class Cp {
         tgtPath: string,
         cwd?: string,
     ): Promise<void> {
-        const tmpFileName = await generateTmpFileName();
-        const command = ['tar', 'zcf', '-'];
-        if (cwd) {
-            command.push('-C', cwd);
-        }
-        command.push(srcPath);
-        const writerStream = fs.createWriteStream(tmpFileName);
-        const errStream = new WritableStreamBuffer();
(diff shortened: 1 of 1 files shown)


fpm-peter:
/hold
there is another problem. The promise doesn't resolve correctly. Will update once I've had more time to investigate

brendandburns:
Looks like CI/CD tests are failing.

fpm-peter:
I can see why the last person didn't finish the PR. With the fake websocket never actually connecting to anything, the promise cannot resolve. Just comparing the called command with the expected called command proves nothing besides that exec is working as intended which is already validated in the exec_test. The exec-websocket-return is ignored in cpFromPod since this also proves nothing but that a connection got established - which is the problem of the bug in the first place.
I didn't find anything in other tests but is there an established way to test with an actual pod?

brendandburns:
@fpm-peter unfortunately we don't have e2e tests with a real kubernetes cluster. You could add them, similar to the ones in the Java client here: https://github.com/kubernetes-client/java/blob/master/.github/workflows/maven.yml#L67

But that is a pretty big change to force you to do to merge this PR.

Perhaps the right answer is to remove this test if it isn't validating anything?

PR #1436, opened 2023-11-21
fix: add support for tls-server-name in KubeConfig

stefreak:
The feature that was added in 076531effb9d072d5544f7f36fb6a820ddd037a4 was missing in the 1.x release branch – we ported the existing feature from there.


brendandburns:
/lgtm
/approve

Thanks for the PR (and especially for adding unit tests!)

mstruebing:
@stefreak If you run `rpm run format` it should fix the code style issues and the pipeline should succeed.

Issue #1398, opened 2023-11-08
1.0.0-rc3 Error "None of the given media types are supported"


tugtugtug:
**Describe the bug**
When any patch functions are called, such as patchNamespacedStatefulSet, the client throws "None of the given media types are supported"

** Client Version **
e.g. `1.0.0-rc3`

** Server Version **
e.g. `1.25.1`

**To Reproduce**
- make a call to patchNamespacedStatefulSet with any parameters

**Expected behavior**
The patch functions should work. And in addition to be working, I'm also not sure how the API will figure out which mediaType to prefer given the content of the body. The old api allows overriding the headers via the options, now this is no longer available.

** Example Code**
```
let kubeConfig = new k8s.KubeConfig();
kubeConfig.loadFromDefault();
const k8sAppsApi = kubeConfig.makeApiClient(k8s.AppsV1Api);
// how to make sure this is done with header
// 'Content-Type': 'application/merge-patch+json'??
await k8sAppsApi.patchNamespacedStatefulSet({
			name: 'test',
			namespace: 'default',
			body: {
				spec: {
					replicas: replicas,
				},
			}
		})
```
**Environment (please complete the following information):**
OS: [Ubuntu 22.04]
NodeJS Version [18]
Cloud runtime [NA]

**Additional context**
Is there a general status the testing coverage of the 1.0.0 branch?
Given the release candidate tag, I had higher expectation of the quality of the code, but it seems quite hard to move forward with all these bugs.

mstruebing:
I think this is connected to #1341 and a known bug right now.

mstruebing:
I think this is connected to #1341 and should be supported if we merge this and cut another RC.

mstruebing:
@tugtugtug I've published a new release `1.0.0-rc4`: https://www.npmjs.com/package/@kubernetes/client-node/v/1.0.0-rc4
Could you confirm that this issue is solved in this release?

tugtugtug:
Verified fixed!

tugtugtug:
Hi @mstruebing,
  I thought things were working as no errors were reported. But apparently the example that changes the content-type does not work.
e.g.
```
    patchNamespacedStatefulSetWithHttpInfo(name, namespace, body, pretty, dryRun, fieldManager, fieldValidation, force, _options) {
        const requestContextPromise = this.requestFactory.patchNamespacedStatefulSet(name, namespace, body, pretty, dryRun, fieldManager, fieldValidation, force, _options);
[...]

PR #1444, opened 2023-11-22
ci: add release workflow for 1.x

mstruebing:
I wanted to publish a new release for https://github.com/kubernetes-client/javascript/issues/1398 but it seems like the workflow is only available on the master branch thus it means it can not get executed in the release-1.x branch. Porting over the file and use latest node LTS in the setup-node action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,51 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        type: string
+        required: true
+        description: The version of this release. Must be a semantic version of the form X.Y.Z.
+      dry_run:
+        type: boolean
+        required: true
+        default: false
+        description: Dry run, will not push branches or upload the artifacts.
+      skip_tag:
+        type: boolean
+        required: true
+        default: false
+        description: If true, don't tag this release, just push it.
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    environment: production
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+      releaseVersion:
+        type: string
+        required: true
+        description: The version of this release. Must be a semantic version of the form X.Y.Z.
+      dry_run:
+        type: boolean
+        required: true
+        default: false
+        description: Dry run, will not push branches or upload the artifacts.
+      pre_release:
```
brendandburns, line 15: For now, this needs to always be `true` or we will accidentally replace the "production" release on npm with the 1.x release. We can decide when we want to do that, but for now at least default to true, and possibly force it to true.

brendandburns:
One comment releated to the `pre-release` flag, but otherwise lgtm.

.github/workflows/release.yml
mstruebing: Good point. I think the condition I've added will always publish with the `--next` flag and never without the `--next` flag.

brendandburns: I'd be a lot more comfortable if we just removed this for now. Is there a reason to leave it in?

mstruebing: Good point, I've tried to make the changes as small as possible but you are very right, removing the risk completely is way better.
I've changed it to remove the input and all references/actions we do not want to tigger.

PR #1445, opened 2023-11-22
fix: Fix TypeError of topNodes

koooge:
Hi there,
This is a fix of topNodes(). It seems to have stopped working due to TypeError since 0.19.x.

```
TypeError: Cannot read properties of undefined (reading 'Running')
    at path/to/node_modules/@kubernetes/client-node/dist/top.js:61:92
    at Array.filter (<anonymous>)\n
    at Object.topNodes (path/to/node_modules/@kubernetes/client-node/dist/top.js:59:21)
...
```

### DIFF
--- a/src/top.ts
+++ b/src/top.ts
@@ -1,4 +1,4 @@
-import { CoreV1Api, V1Node, V1Pod, V1PodList, V1PodStatus } from './gen/api';
+import { CoreV1Api, V1Node, V1Pod, V1PodList } from './gen/api';
 import { Metrics, PodMetric } from './metrics';
 import {
     add,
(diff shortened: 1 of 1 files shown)


mstruebing:
Hey. @koooge thank you very much for your PR.
Are you sure this is the fix? When looking at the code and the error message I guess it's more likely the reason that the `V1PodStatus` class doesn't contain a `PhaseEnum` anymore. Can you validate?

With your change you should be able to remove the `@ts-ignore` comment above and the error should pop up.
I think the part after/including the `||` can just be removed.

koooge:
Hi @mstruebing , I think you are right! So this is almost the same as reverting https://github.com/kubernetes-client/javascript/commit/3520057e7ce2754254f2c34897be76a4cde58ea4

mstruebing:
Yea thank you 🙈 

/lgtm
/approve

Issue #1397, opened 2023-11-08
1.0.0rc3 missing AutoscalingV2Api


tugtugtug:
**Describe the bug**
I'm trying to migrate to the 1.0.0rc3 version of the client to avoid CVEs reported in the request lib depended by the current 0.20.0 version. But during the migration, I found several API versions are missing, e.g. the AutoscalingV2Api.

** Client Version **
e.g. `1.0.0-rc3`

** Server Version **
e.g. `1.25.1`

**To Reproduce**
1. makeApiClient(k8s.AutoscalingV2Api);
2. this is not compiling anymore as k8s.AutoscalingV2Api doesn't exist in the 1.0.0 version.

**Expected behavior**
I expect the 1.0.0 version to have matching API version support.

** Example Code**
```
let kubeConfig = new k8s.KubeConfig();
kubeConfig.loadFromDefault();
kubeConfig.makeApiClient(k8s.AutoscalingV2Api);
```

**Environment (please complete the following information):**
 - OS: [Ubuntu 22.04]
 - NodeJS Version [18]
 - Cloud runtime [NA]

**Additional context**
If no quick fix can be found, a workaround is welcome so we can move on with the migration.

mstruebing:
@tugtugtug This should be connected to us using an older Kubernetes version in the `release-1.x` branch.
I will update this probably sometime this week and cut another RC.

mstruebing:
@tugtugtug I've published a new release `1.0.0-rc4`: https://www.npmjs.com/package/@kubernetes/client-node/v/1.0.0-rc4
Could you confirm that this issue is solved in this release?

tugtugtug:
Verified resolved!

PR #1461, opened 2023-12-01
[release-1.x] regenerate client for 1.28

mstruebing:
We still used 1.22 in here.

I've updated the k8s branch to use the latest 1.28 tag (1.28.4) and regenerated the client.
I will do a separate PR to port over the action.

For reference, an issue which is affected by this: https://github.com/kubernetes-client/javascript/issues/1397


brendandburns:
/lgtm
/approve

We should make sure we do this more often :)

Issue #1420, opened 2023-11-14
1.0.0-rc3 Missing function / classes for configuration


eyalcha:
The following doesn't exists 

```
const { createConfiguration, ServerConfiguration } = require('../dist');
const { PromiseMiddlewareWrapper } = require('../dist/gen/middleware');
```

https://github.com/kubernetes-client/javascript/blob/release-1.x/examples/patch-example.js

SarthakNikhal:
Hey. Which functionality is broken because of this issue? Can I work on it?

mstruebing:
@eyalcha I believe this should be available if we release another RC.
Can you validate if the issue is still the same on the `release-1.x` branch?

mstruebing:
@eyalcha I've published a new release `1.0.0-rc4`: https://www.npmjs.com/package/@kubernetes/client-node/v/1.0.0-rc4
Could you confirm that this issue is solved in this release?

PR #1480, opened 2023-12-15
chore: switch from nyc to c8

### LINKED ISSUE #1421: release-1.x branch unit test memory consumption increase
this issue is to track the growth in memory usage when running `npm test`, which has lead to increasing the node memory as a stop gap

> i was able to make some progress isolating the gc memory issue- maybe we make a separate issue to track it
> 
> by executing nyc mocha with the --node-command inspect to enable debugging i was able to connect a memory profiler and get a GC snapshot and allocation timeline
> 
> the majority of the growing heap usage seems to be coming from SourceMapConsumer in node_modules/source-map/lib/source-map-consumer.js at the bottom of a pretty long call stack that includes @ babel/core
> 
> i found https://github.com/meteor/meteor/issues/9568 potentially relevant link that has encouraging similarities.
> 
> by changing nyc.instrument=false in package.json on line 51, the memory never exceeded 600MB, while re-enabling that line eventually balloons up to 12GB which was the most i was willing to test.
> 
> i will continue investigating upgrading our source-map indirect dependency so we can hopefully get this resolved

mstruebing:
I wanted to cut another RC to potentially close
* #1420
* #1398
* #1397

but the release is erroring at the test because of the memory limit
so I decided to use c8 instead of nyc because when trying to remove nyc
I concluded that the memory error comes from it and its last release was
4 years ago. It works as a drop-in replacement.

https://www.npmjs.com/package/c8

There are slight differences in the coverage reporting, what they catch as covered and uncovered, but I believe that's okay for us. 

* Closes #1421 

Before:
Screenshot 2023-12-15 at 12 41 21

After:
Screenshot 2023-12-15 at 12 41 08

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,5 @@ jobs:
       - run: node version-check.js
       - run: npm ci
       - run: npm test
-        env:
-          NODE_OPTIONS: --max-old-space-size=8192
       - run: npm run lint
       - run: npm audit --audit-level=critical
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
                 "@types/chai-as-promised": "^7.1.5",
                 "@types/mocha": "^10.0.1",
                 "@types/mock-fs": "^4.13.1",
+                "c8": "^8.0.1",
                 "chai": "^4.2.0",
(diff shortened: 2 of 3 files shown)


brendandburns:
/lgtm
/approve

Thanks!

PR #1488, opened 2023-12-27
fix: do not overwrite current context

carlossg:
with a null/undefined one

When having multiple configs in KUBECONFIG only one of them may have the current context set
but it will be overwritten by an empty value if preserveContext is not explicitly set
which is not the case when using kubectl binary

cherry-pick of #1487 into release-1.x


mstruebing:
/lgtm
/approve

Thank you

PR #1549, opened 2024-01-30
feat: custom path for service account token (supporting TokenVolume projection)

### LINKED ISSUE #1528: How to use this client with TokenVolume Projection
Hey,
I tried looking at the code but I couldn't figure how to do it. I am trying to use the more secure TokenVolume Projection to load short lived token instead of the long lived one, but I can't figure how to do it.

Can you please help me out with some examples?

omerlh:
fix #1528

I am not sure about the CA - I need to see if it will be there if using the projected volume 🤔

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -211,6 +211,12 @@ export class KubeConfig {
         const clusterName = 'inCluster';
         const userName = 'inClusterUser';
         const contextName = 'inClusterContext';
+        const tokenFile = process.env.TOKEN_FILE_PATH
+            ? process.env.TOKEN_FILE_PATH
+            : `${pathPrefix}${Config.SERVICEACCOUNT_TOKEN_PATH}`;
+        const caFile = process.env.KUBERNETES_CA_FILE_PATH
+            ? process.env.KUBERNETES_CA_FILE_PATH
+            : `${pathPrefix}${Config.SERVICEACCOUNT_CA_PATH}`;
 
         let scheme = 'https';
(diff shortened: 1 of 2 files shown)


omerlh:
While investigating the issue with CA and token projection I found this issue kubernetes/kubernetes#53315, I asked for help on Kubernetes slack

omerlh:
When using TokenProjection need also to mount the custom config map `kube-root-ca.crt` with the API server CA certificate so I add the option to specify path to the CA as well

brendandburns:
/lgtm
/approve

Thanks for the PR!

Once this merges, can you cherrypick it into the `release-1.x` branch also?

brendandburns:
ugh, looks like we need to fix the linter somewhat.

omerlh:
Yeah, the linter error does not look related to my changes so I am not sure what to do, I think need to upgrade linter rules maybe?

mstruebing:
Hey @omerlh I think the linter errors are not really failing, prettier is what is failing.
If I run `npm run format` and after that `npm run lint` I get an exit code of zero.
Without running `npm run format` it's 1. So if you run `npm run format` and commit and push these changes the linter should be satisfied.

I'm not sure about our linter setup though, we should probably switch to eslint or something else which is supported as tslint is deprecated since quite a while (2019)

omerlh:
Sure, I just did - let's see if this will help!

mstruebing:
/lgtm
/approve

Thank you for your contribution 🎉

PR #1568, opened 2024-02-11
fix: use load from cluster when default service mount is disabled

omerlh:
fixing a bug in #1549


brendandburns:
/lgtm
/approve

Thanks for the PR!

omerlh:
Thanks! When will you release a new version? I want to test it out

PR #1608, opened 2024-03-08
feat: add reexports for AbortError and FetchError from node-fetch to aid instanceof checks

xkabylgSICKAG:
`instanceof` checks must use the same instance, so if a user installed a different version of `node-fetch` in their project the `instanceof` checks would fail. By reexporting we ensure that a user can get the exact same instance that this client would return.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
             "dependencies": {
                 "@types/js-yaml": "^4.0.1",
                 "@types/node": "^20.3.1",
-                "@types/node-fetch": "^2.6.3",
+                "@types/node-fetch": "^2.6.9",
                 "@types/stream-buffers": "^3.0.3",
                 "@types/tar": "^6.1.1",
                 "@types/underscore": "^1.8.9",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -296,25 +296,12 @@
             }
         },
         "node_modules/@types/node-fetch": {
-            "version": "2.6.3",
-            "resolved": "https://registry.npmjs.org/@types/node-fetch/-/node-fetch-2.6.3.tgz",
-            "integrity": "sha512-ETTL1mOEdq/sxUtgtOhKjyB2Irra4cjxksvcMUR5Zr4n+PxVhsCD9WS46oPbHL3et9Zde7CNRr+WUNlcHvsX+w==",
+            "version": "2.6.11",
+            "resolved": "https://deagxartifactory.sickcn.net:443/artifactory/api/npm/npm/@types/node-fetch/-/node-fetch-2.6.11.tgz",
```
brendandburns, line 300: This should be the standard NPM registry, not some random URL registry.

brendandburns:
Thanks for the PR, one comment, which is that you have appeared to pick up your local npm mirror/registry in package-lock, it needs to be the main one.

package-lock.json
xkabylgSICKAG: Fixed! :)

xkabylgSICKAG:
Fixed the lockfile and rebased!

PR #1620, opened 2024-03-19
update openid-client

Babak-abd:
Updating openid to get jose updated (CVE-2024-28176)

├─┬ openid-client@5.5.0
     ├── jose@4.14.6


mstruebing:
/lgtm
/approve

Thank you, in fact we are using version ranges whereas the updated `openid-client` version should be in that range but it doesn't hurt either so I'm approving. 🚀

PR #1665, opened 2024-04-27
[MASTER] fix: use correct types for websocket

### LINKED ISSUE #1662: exec type is wrong
**Describe the bug**

`exec` method on `Exec` class returns `Promise<WebSocket>`.

As the [source code](https://github.com/kubernetes-client/javascript/blob/b8b0ec522bf42086f7c2e1b8648b478b3f584fa8/src/exec.ts#L1), `WebSocket` is imported as `import WebSocket = require('isomorphic-ws');`. Unfortunately, such `WebSocket` is the module value, not the type, the correct type should be `WebSocket.WebSocket`. So the result `WebSocket` type actually fallback to `globalThis.WebSocket` (from DOM or node type definitions as user's tsconfig) , which is not compatible with `ws` on node env.

mstruebing:
Same as #1664 but for `master` as well as I'm not entirely sure when we will promote `release-1.x`.

* Closes #1662

### DIFF
--- a/src/attach.ts
+++ b/src/attach.ts
@@ -23,7 +23,7 @@ export class Attach {
         stderr: stream.Writable | any,
         stdin: stream.Readable | any,
         tty: boolean,
-    ): Promise<WebSocket> {
+    ): Promise<WebSocket.WebSocket> {
         const query = {
             container: containerName,
             stderr: stderr != null,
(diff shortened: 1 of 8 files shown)


mstruebing:
How did I self approve this?

PR #1694, opened 2024-05-25
make namespace in Object header optional

schrodit:
the `KubernetesObjectHeader` type is currently wrong as a namespace is never required.
Clusterwide resources do not need a namespace at all and namespace resources will be defaulted to the `default` namespace (although something we might want to think about removing in a SDK.)


mstruebing:
/lgtm
/approve

Thanks 🚀

PR #1695, opened 2024-05-26
Properly parse metadata of custom Kubernetes objects

schrodit:
Part of https://github.com/kubernetes-client/javascript/issues/852

### DIFF
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -1,0 +1,115 @@
+import { ObjectSerializer } from "./api";
+import { V1ObjectMeta } from "./gen/model/v1ObjectMeta";
+
+export type AttributeType = {
+    name: string;
+    baseName: string;
+    type: string;
+};
+
+export class KubernetesObject {
+    /**
+    * APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+    */
+    'apiVersion'?: string;
+    /**
+    * Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+    */
+    'kind'?: string;
+    'metadata'?: V1ObjectMeta;
+
+    static attributeTypeMap: AttributeType[] = [
+        {
+            "name": "apiVersion",
+            "baseName": "apiVersion",
+            "type": "string"
+        },
+        {
+            "name": "kind",
+            "baseName": "kind",
+            "type": "string"
+        },
+        {
+            "name": "metadata",
+            "baseName": "metadata",
+            "type": "V1ObjectMeta"
+        },
+    ];
+}
+
+const isKubernetesObject = (data: unknown): boolean =>
+    !!data && typeof data === "object" && 'apiVersion' in data && 'kind' in data;
+
+/**
+ * Wraps the ObjectSerializer to support custom resources and generic Kubernetes objects.
+ */
+class KubernetesObjectSerializer {
+
+    private static _instance: KubernetesObjectSerializer;
+
+    public static get instance() {
+        if (this._instance) {
+            return this._instance;
+        }
+        this._instance = new KubernetesObjectSerializer();
+        return this._instance;
+    }
+
+    private constructor() {}
+
+    public serialize(data: any, type: string) {
+        const obj = ObjectSerializer.serialize(data, type);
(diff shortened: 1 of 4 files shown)


src/serializer.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (!isKubernetesObject(data)) {
+            return obj;
+        }
+
+        const instance: Record<string, any> = {};
+        const attributeTypes = KubernetesObject.attributeTypeMap;
+        for (let index = 0; index < attributeTypes.length; index++) {
+            let attributeType = attributeTypes[index];
+            instance[attributeType.name] = ObjectSerializer.serialize(data[attributeType.baseName], attributeType.type);
+        }
```
mstruebing, line 75: Would it be easier to a for ... in loop instead? (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

```js
for (const attributeType in KubernetesObject.attributeTypeMap) {
   // ...
}
```

Seems a bit less error prone and easier readable in my opinion.

Otherwise, as you declare `attributeType` in the loop itself, it could be a `const` instead of a `let` I think.

src/serializer.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (!isKubernetesObject(data)) {
+            return obj;
+        }
+
+        const instance = new KubernetesObject();
+        const attributeTypes = KubernetesObject.attributeTypeMap;
+        for (let index = 0; index < attributeTypes.length; index++) {
+            let attributeType = attributeTypes[index];
+            instance[attributeType.name] = ObjectSerializer.deserialize(data[attributeType.baseName], attributeType.type);
+        }
```
mstruebing, line 103: Similar here, or is there a reason for not using `for ... in`?

src/serializer.ts
```diff
@@ -0,0 +1,115 @@
+import { ObjectSerializer } from "./api";
+import { V1ObjectMeta } from "./gen/model/v1ObjectMeta";
+
+export type AttributeType = {
+    name: string;
+    baseName: string;
+    type: string;
+};
+
+export class KubernetesObject {
```
mstruebing, line 10: This is never used anywhere else, do we want to `export` it here for any consumer of this library to use or was this by mistake?

src/serializer.ts
```diff
@@ -0,0 +1,115 @@
+import { ObjectSerializer } from "./api";
+import { V1ObjectMeta } from "./gen/model/v1ObjectMeta";
+
+export type AttributeType = {
```
mstruebing, line 4: This is never used anywhere else, do we want to export it here for any consumer of this library to use or was this by mistake?

mstruebing (commented):
Great work - only a couple of nitpicks.

I think we should port this over to `release-1.x` as well so that we will have it in the next major release as well.

src/serializer.ts
schrodit: I think you rather mean the `for .. of` loop. I changed it to that one.

src/serializer.ts
schrodit: no reason. And I agree that the `for in` / `for of` loops are more readable.
Also changed it here

src/serializer.ts
schrodit: It will be used when I add support for custom types.

But yes currently its not used and I will export it when it actually used

schrodit:
> Great work - only a couple of nitpicks.
> 
> I think we should port this over to `release-1.x` as well so that we will have it in the next major release as well.

Whats the process here?
Do I just open another PR for the `release-1.x` branch?

src/serializer.ts
mstruebing: Yes sorry, `for .. in` is for objects whereas `KubernetesObject.attributeTypeMap` is an array, so you are right!
Thanks for changing it ❤️

mstruebing:
Yes exactly, otherwise these changes will be lost once we release the new client without the request library as a replacement for the current master. You can probably reuse most of the code you've created in this PR.

Your changes looking good from my point of view - will just wait a bit for @brendandburns to have a look as well probably.

brendandburns:
Looks good to me also.

/lgtm

Once this merges, you can cherry-pick this against the release-1.x branch, but it may require reworking.

brendandburns:
Looks like this is failing linting. Please fix to match our style guide.

schrodit:
> Looks good to me also.
> 
> /lgtm
> 
> Once this merges, you can cherry-pick this against the release-1.x branch, but it may require reworking.

ok will do that once merged

mstruebing:
```
[warn] src/object.ts
[warn] src/serializer_test.ts
[warn] src/serializer.ts
[warn] Code style issues found in 3 files. Run Prettier to fix.
```

You should run: `npm run format` and commit the formatting changes.

Issue #1595, opened 2024-03-04
create from object feature


zfrhv:
hello, i have been looking at the examples and couldn't find any example of creating from object function.  
like in the kubernetes python they have [create from dict](https://github.com/kubernetes-client/python/blob/df43d155d9252ab980bf5454e4b9574c5d5c7348/kubernetes/utils/create_from_yaml.py#L173) im looking for something similar for js.

all examples describe that in order to create an object you have to use function for that object.  
for example for creating namespace you have to run `k8sApi.createNamespace(namespace)`, but i want something like `k8sApi.createObject(namespace)`

brendandburns:
There's not a really great example of this, you need an equivalent of the generic client in Java, but it doesn't exist.

The closest example is this "raw" example:

https://github.com/kubernetes-client/javascript/blob/master/examples/raw-example.js

Which you could use, but you'll need to be able to figure out the right HTTP path from the object before hand.

zfrhv:
i see
thank you

why not to implement the this feature then? it seems pretty easy to do.
just take the object api version and kind, and easily you have the uri needed, like kubectl

i think it will very simplify the usage of this tool

brendandburns:
We'd gladly take a PR implementing this if you want to send it.

zfrhv:
alright, just so happens that i [already did it](https://github.com/zfrhv/hnc-ui/blob/main/image/k8s.js) ; )

i will create a seperate class called something like `k8sApi` instead of `CoreV1Api` sounds good?

it will take me few weeks tho, a little busy these days

brendandburns:
Sounds great! Perhaps `GenericApi` ?

zfrhv:
neat

PR #1705, opened 2024-06-03
add examples to generic api (inside object.ts)

zfrhv:
solves #1595

let me know if it looks good.
then I will add it to docs and maybe also use genericApi at get/apply/delete...

### DIFF
--- a/examples/typescript/apply/apply-from-file-example.ts
+++ b/examples/typescript/apply/apply-from-file-example.ts
@@ -1,0 +1,53 @@
+import * as k8s from '@kubernetes/client-node';
+import * as yaml from 'js-yaml';
+import { promises as fs } from 'fs';
+
+/**
+ * Replicate the functionality of `kubectl apply`.  That is, create the resources defined in the `specFile` if they do
+ * not exist, patch them if they do exist.
+ *
+ * @param specPath File system path to a YAML Kubernetes spec.
+ * @return Array of resources created
+ */
+export async function apply(specPath: string): Promise<k8s.KubernetesObject[]> {
+    const kc = new k8s.KubeConfig();
+    kc.loadFromDefault();
+
+    const client = k8s.KubernetesObjectApi.makeApiClient(kc);
+
+    const specString = await fs.readFile(specPath, 'utf8');
+    const specs: k8s.KubernetesObject[] = yaml.loadAll(specString);
+    const validSpecs = specs.filter((s) => s && s.kind && s.metadata);
+    const created: k8s.KubernetesObject[] = [];
+    for (const spec of validSpecs) {
+        // this is to convince the old version of TypeScript that metadata exists even though we already filtered specs
+        // without metadata out
+        spec.metadata = spec.metadata || {};
+        spec.metadata.annotations = spec.metadata.annotations || {};
+        delete spec.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration'];
+        spec.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration'] = JSON.stringify(spec);
+        try {
+            // try to get the resource, if it does not exist an error will be thrown and we will end up in the catch
+            // block.
+            await client.read(spec);
+            // we got the resource, so it exists, so patch it
+            //
+            // Note that this could fail if the spec refers to a custom resource. For custom resources you may need
+            // to specify a different patch merge strategy in the content-type header.
+            //
+            // See: https://github.com/kubernetes/kubernetes/issues/97423
+            const response = await client.patch(spec);
+            created.push(response.body);
+        } catch (err) {
+            // if the resource doesnt exist then create it
+            if (err instanceof k8s.HttpError && err.statusCode === 404) {
+              const response = await client.create(spec);
+              created.push(response.body);
+            } else {
+              throw err
+            }
+        }
+    }
+
+    return created;
+}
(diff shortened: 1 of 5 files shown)


src/generic_api.ts
```diff
@@ -0,0 +1,51 @@
+import querystring = require('querystring');
+import pluralize = require('pluralize');
+
+export interface K8sObject {
```
brendandburns, line 4: Perhaps `KubernetesObject` instead for clarity.

src/generic_api.ts
```diff
@@ -0,0 +1,51 @@
...
+    path += "/apis/" + object.apiVersion
+  }
+
+  // add namespace if object namespaced
+  if (object.metadata?.namespace) {
+    path += "/namespaces/" + object.metadata.namespace
+  }
+
+  // add object kind
+  path += "/" + pluralize(object.kind.toLowerCase())
```
brendandburns, line 37: fwiw, the plural  information is available via the Discovery API on the Kubernetes cluster. This probably works for nearly everything, so it's an ok approach for now, but perhaps add:

`// TODO: Use discovery api to get the correct plural`

brendandburns:
This looks pretty good, added some small comments and it looks like you need to rebase to pick up some dependabot changes that got merged.

zfrhv:
I was going to create `genericApi` class for get/apply/delete/create.. functions but i stumbled into [object.ts](https://github.com/kubernetes-client/javascript/blob/master/src/object.ts), it looks like its the generic api that I was looking for.
although looking at [example usage](https://github.com/kubernetes-client/javascript/blob/master/examples/typescript/apply/apply-example.ts) it seems a bit complicated to use.
still this class is exactly what I was about to make.

should I just add more example and try to use `watch` with `object.ts` instead of creating `genericApi`?

src/generic_api.ts
```diff
@@ -0,0 +1,52 @@
...
+    apiVersion: string;
+    kind: string;
+    metadata?: {
+        name?: string;
+        namespace?: string;
+        labels?: object;
+    };
+}
+
+export function CreatePath(object: KubernetesObject, include_name: boolean = true) {
```
mstruebing, line 14: As this is a function, should we make it (lower)camelCase?

mstruebing (commented):
Only one comment about the casing of the `CreatePath` function, otherwise great PR from my point of view. 🚀 

Would be great if you could cherry-pick your commit into the `release-1.x` branch after this got merged to not loose these changes if we release the new major version.
You should be able to reuse most of your code but perhaps get into trouble with `package.json`/`package-lock.json` files.

brendandburns:
@zfrhv Thanks! I had forgotten about that.  Yes, I think adding `watch` to that and also maybe simplifying it if you want to? That would be great!

zfrhv:
I tried to combine `object.ts` with `watch.ts` but it seems too complex for me.
I can create seperate watch function inside `object.ts` but thats just copy paste from `watch.ts` and it will look messy.
So I tried to call the watch function from `KubernetesObjectApi` inside `object.ts` but cant figure out how to pass the Kubernetes config, it seems like its locally stored as a token and im not sure how to use it.
maybe @schrodit can help to create `watch` inside `object.ts`

About the extra examples I added: i didnt tested them but i think they should work.
All my k8s clusters are airgapped and testing something from the internet is pain.
I wish someone else could test it for me <3

schrodit:
> I tried to combine `object.ts` with `watch.ts` but it seems too complex for me.
> 
> I can create seperate watch function inside `object.ts` but thats just copy paste from `watch.ts` and it will look messy.
> 
> So I tried to call the watch function from `KubernetesObjectApi` inside `object.ts` but cant figure out how to pass the Kubernetes config, it seems like its locally stored as a token and im not sure how to use it.
> 
> maybe @schrodit can help to create `watch` inside `object.ts`
> 
> 
> 
> About the extra examples I added: i didnt tested them but i think they should work.
> 
> All my k8s clusters are airgapped and testing something from the internet is pain.
> 
> I wish someone else could test it for me <3
> 
> 

I think I already did that in an internal wrapper.

Let me take a look and I open a separate PR.

examples/typescript/apply/apply-from-file-example.ts
```diff
@@ -0,0 +1,49 @@
...
+            await client.read(spec);
+            // we got the resource, so it exists, so patch it
+            //
+            // Note that this could fail if the spec refers to a custom resource. For custom resources you may need
+            // to specify a different patch merge strategy in the content-type header.
+            //
+            // See: https://github.com/kubernetes/kubernetes/issues/97423
+            const response = await client.patch(spec);
+            created.push(response.body);
+        } catch (e) {
```
schrodit, line 41: You should check here if the error code is 404 and only then try to create the resource. Otherwise we suppress the original error.

zfrhv:
I fixed the examples.
Im not good enought to understand how to add watch function inside `object.js` so I will leave it to the pros

mstruebing:
@zfrhv started the CI process and it seems like you have linting errors.
You can see the results either in the github action logs or locally by running `npm run lint`

brendandburns:
You can also run `npm run format` to fix at least some of them.

mstruebing:
/lgtm
/approve

Thanks for these example, I think they are very useful for new users. 🚀
As mentioned above, if you ever find the time, it would be great to port these to the `release-1.x` branch as well.

Issue #1693, opened 2024-05-22
Release of v1.0.0


rossanthony:
I'm curious if anyone is actively working on further enhancements to the [release-1.x](https://github.com/kubernetes-client/javascript/tree/release-1.x) branch? I'm keen to help get it over the finish line, if there's enough pending work to divide and conquer? 

I noticed there are a couple of things mentioned:

- Complexity of middleware use: https://github.com/kubernetes-client/javascript/issues/754#issuecomment-2004562303
- The errors being returned appear to have a body of a string or undefined, and not an object as they were in 0.2.0 - https://github.com/kubernetes-client/javascript/issues/754#issuecomment-2074904937

There's also a snyk vulnerability with the version of the tar package being using in 1.0.0-rc4 and I noticed there are some @types in the dependencies which should be moved to devDependencies: https://github.com/kubernetes-client/javascript/blob/release-1.x/package.json#L57-L63

@brendanburns let me know if there's anything I can do to help? It's hard to know what might be in flight already, so wanted to check first before starting anything and potentially stepping on anyones toes.

brendandburns:
We're happy to have help here. I know that @mstruebing was working on some things, but nothing too active. Please feel free to send PRs for any of those issues.

Lately the main blocker has been regenerating the client for Kubernetes 1.30

rossanthony:
> Lately the main blocker has been regenerating the client for Kubernetes 1.30

Oh interesting, so there are some breaking changes with the kube API in 1.30?

brendandburns:
No, for some reason when I regenerate, the code isn't changing, the generation logic is stuck on 1.28.4, it's something in the script somewhere but I haven't had time to figure it out.

davidgamero:
@rossanthony 

For the complexity of middleware use, the current approach relies on reconstructing internals `ServerConfiguration` as a result of the inability to merge headers at request time into generated API calls as the[ configuration is replaced with no option to merge](https://github.com/OpenAPITools/openapi-generator/blob/ed2aad6756db1dfc780f709f0a6ee0feeb00f31a/modules/openapi-generator/src/main/resources/typescript/api/api.mustache#L46)

in order to improve our ergonomics of the client, we could create something like a `WithHeader(kc, key, value): Configuration` utility wrapper to clean up the usage for now which could improve the Patch use case:

```
const kc = new k8s.KubeConfig();
...
await k8sApi.patchNamespacedPod(
  {
    name: podName,
    namespace: namespace,
[...]

PR #1708, opened 2024-06-04
Regenerate for Kubernetes 1.29

brendandburns:
cc @mstruebing 

I ran this manually and it worked. Not sure why the automation doesn't.


brendandburns:
Related #1693

PR #1754, opened 2024-06-24
add examples to generic api (inside object.ts)

zfrhv:
same as #1705

### DIFF
--- a/examples/kubectl/equivalents/apply.js
+++ b/examples/kubectl/equivalents/apply.js
@@ -1,5 +1,5 @@
 // in a real program use require('@kubernetes/client-node')
-import * as k8s from '../../../dist';
+import * as k8s from '../../../dist/index';
 
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
--- a/examples/kubectl/equivalents/create.js
+++ b/examples/kubectl/equivalents/create.js
@@ -1,0 +1,15 @@
+// in a real program use require('@kubernetes/client-node')
+import * as k8s from '../../../dist/index';
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const client = k8s.KubernetesObjectApi.makeApiClient(kc);
+
+const namespace = {
+    metadata: {
+        name: 'test'
+    }
+}
+
+client.create(namespace)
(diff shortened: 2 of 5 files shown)


examples/kubectl/equivalents/apply.js
```diff
@@ -0,0 +1,21 @@
+const k8s = require('@kubernetes/client-node');
```
mstruebing, line 1: Please use a relative, file import instead to make sure we always use the code from the specific commit and not the released one:

```
// in a real program use require('@kubernetes/client-node')
import * as k8s from '../../../dist/index';
```

Sorry, I've missed that in the other PR but I think on `master` we sill have the old pattern, which is ok as we want to switch to `release-1.x` as soon as possible even though it's taking a while.

mstruebing (commented):
Did you test these with the new code?
Not sure if we changed anything in these APIs right now on top of my head.
If not, I can test this soonish but doesn't have the time right now.

zfrhv:
hi, I didnt tested these examples not with the new code and not with old.
I dont have environments anymore to test.

sorry for the misunderstanding

examples/kubectl/equivalents/apply.js
zfrhv: sure

zfrhv: wait sorry, why should we use relative import?
its examples how the users should use it, and they will just `npm install @kubernetes/client-node` and then `require('@kubernetes/client-node')` in their code

davidgamero: when importing from `@kubernetes/client-node`, by default the installed version from the `node_modules` directory will be imported, which will be whatever the last release was when running `npm i @kubernetes/client-node`

the relative import will use the current version of the code in the repo (at the checked out commit)

we can test the examples by executing them, and testing the current commit can catch issues as soon as code is changed, while testing imports from `@kubernetes/client-node` will only detect issues when a release is made

it is possible to get around this by replacing the import at test execution time or putting the tests into a separate module that uses a [local path import](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths), but for now the comment is used

zfrhv: thank you for explaining

mstruebing:
/lgtm
/approve

Thanks 🚀

PR #1801, opened 2024-07-23
server side apply support

davidgamero:
add `application/apply-patch+yaml` Patch Strategy to allow [server-side apply ](https://kubernetes.io/docs/reference/using-api/server-side-apply/#api-implementation)

from @eplightning 's feedback on release

also husky cleaned up  deprecated part of one of it's scripts](https://github.com/typicode/husky/issues/1468)


brendandburns:
/lgtm
/approve

thanks!

Issue #1809, opened 2024-07-25
Auth exec error message is not clear


0x5457:
**Describe the bug**

When the auth exec command configured in kubeconfig does not exist or cannot be executed for other reasons. The following error is obtained

```sh
TypeError: Cannot read properties of null (reading 'toString')
    at ExecAuth.getCredential (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/exec_auth.js:86:39)
    at ExecAuth.applyAuthentication (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/exec_auth.js:23:33)
    at KubeConfig.applyAuthorizationHeader (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/config.js:447:33)
    at KubeConfig.applyOptions (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/config.js:455:20)
    at KubeConfig.applyToHTTPSOptions (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/config.js:109:20)
    at KubeConfig.applyToFetchOptions (/code/test/node_modules/.pnpm/@kubernetes+client-node@1.0.0-rc6/node_modules/@kubernetes/client-node/dist/config.js:85:20)
```
** Client Version **
`1.0.0-rc6`

** Server Version **
`1.0.0-rc5`

**Environment (please complete the following information):**
 - OS: MacOs
 - NodeJS Version: v20.4.0

brendandburns:
Closing since #1810 has merged.

PR #1810, opened 2024-07-25
fix: if spawnSync has an error should throw it

0x5457:
close: #1809


mstruebing:
/lgtm
/approve

Thanks 🚀

PR #1831, opened 2024-08-10
add generic and typed watch to KubernetesObjectApi

schrodit:
Follow up of https://github.com/kubernetes-client/javascript/pull/1705#issuecomment-2148569436

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -546,7 +546,7 @@ export class KubernetesObjectApi extends ApisApi {
      *
      * @returns WatchResult object that can be used to abort the watch.
      */
-    public async watch<T>({
+    public async watch<T extends KubernetesObject | KubernetesObject>({
         resource,
         options = {},
         callback,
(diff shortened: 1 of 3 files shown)


src/object.ts
```diff
@@ -44,6 +45,60 @@ enum KubernetesPatchStrategies {
...
+    object: T;
+};
+
+export type WatchCallback<T extends KubernetesObject | KubernetesObject> = (
+    phase: KubernetesEventType,
+    apiObj: T,
+    watchObj?: WatchObject<T>,
+) => void;
+
+export type WatchOptions = {
```
schrodit, line 79: I think those are the only 2 options that make sense for watches

src/object.ts
```diff
@@ -474,6 +532,64 @@ export class KubernetesObjectApi extends ApisApi {
...
+        done,
+    }: {
+        resource: {
+            apiVersion: string;
+            kind: string;
+            namespace?: string;
+        };
+        options?: WatchOptions;
+        callback: WatchCallback<T>;
+        done: (err: unknown) => void;
```
schrodit, line 562: might make sense to also have done as optional callback

brendandburns:
Looks like CI/CD is failing for some reason.

schrodit:
> This looks good to me. Any chance that you can add unit tests? If not I'll do it as a follow up.

added some tests

src/util.ts
```diff
@@ -162,3 +162,35 @@ export async function generateTmpFileName(): Promise<string> {
...
 }
+
+/**
+ * If you don't want to or can't await a promise right await and need to store
+ * it in a variable to await it later, wrap it with this function.
+ * Otherwise, if the promise rejects before you await it, it will cause an
+ * unhandled promise rejection which can cause process termination (in node).
+ */
+export const awaitLater = <T, P extends Promise<T>>(p: P): P => {
+    p.catch(() => {});
```
mstruebing, line 173:
```suggestion
    p.catch(() => {}); // tslint:disable-line:no-empty
```

schrodit:
@mstruebing can you take another look?

Do we have to port it to the release-1 branch?

brendandburns:
/lgtm
/approve

Thanks for the PR. It would be appreciated (but not required) to port this to the 1.x branch also.

Soon the 0.x branch will be deprecated and will be replaced by the 1.x branch.

Issue #1767, opened 2024-07-02
cpToPod passes an async function into exec's statusCallback, but the promise is never awaited


Moggers:
cpToPod [passes an async function as the statusCallback](https://github.com/kubernetes-client/javascript/blob/master/src/cp.ts#L90-L93); but [exec does not await the function](https://github.com/kubernetes-client/javascript/blob/master/src/exec.ts#L56-L58)
Once upon a time [it was declared as an asynchronous function](https://github.com/kubernetes-client/javascript/commit/25197421fc7752c510af292e1367d0351274af48#diff-748946b70991cc31404a6e031013af45671e6b08472b4be2c875b0a1a3d92d23L36) but even in that case, the [function wasn't awaited](https://github.com/kubernetes-client/javascript/commit/25197421fc7752c510af292e1367d0351274af48#diff-748946b70991cc31404a6e031013af45671e6b08472b4be2c875b0a1a3d92d23R61).
This appears to result in the throw within cptopod (in cases where stderr is populated or the websocket connection fails) coercing into a promise which never gets handled; resulting ultimately in an uncaught promise rejection.

brendandburns:
We'd be happy to take a PR with a test demonstrating the problem, along with the fix.

As it stands right now, I'm not sure this is that easy to reproduce without a unit test (and I'm not certain that this is actually what is causing your problem)

xelinorg:
By just reading the code I cam see that **_public async cpToPod_**  on https://github.com/kubernetes-client/javascript/blob/60b5427b4dfe2a26a206af95e54ea65a354162d7/src/cp.ts#L68C4-L68C25 should return a  Promise<void> but does not return anything.

You can not return _**this.execInstance.exec**_ on https://github.com/kubernetes-client/javascript/blob/60b5427b4dfe2a26a206af95e54ea65a354162d7/src/cp.ts#L81 because this returns a Promise<WebSocket.WebSocket>.

You could await at _**this.execInstance.exec**_  to get the websocket and then after examining the status of the websocket conditionally return Promise.resolve() or Promise.reject(). Or if there is not need to examine anything just return Promise.resolve().

I hope this makes sense.

cjihrig:
Should this issue be closed since https://github.com/kubernetes-client/javascript/pull/1880 was merged?

mstruebing:
Actually yes, thank you.
The keyword in the pull request was wrong so it did not automatically close.

PR #1880, opened 2024-09-20
Fix cp promise returns by converting the exec callbacks into promises

joeferner:
Addresses issue https://github.com/kubernetes-client/javascript/issues/1767

Wraps the exec in a promise properly such that awaiting the promise will only return after the file is copied


brendandburns:
Thanks for the PR! Can you add a unit test demonstrating the problem as well so that we can be sure this fix works (and that it stays fixed)

joeferner:
enhanced unit tests for both `cpFromPod` and `cpToPod` to test both the promises and that the data send/received was correct. Also deleted the temporary file created when running `cpToPod`

src/cp_test.ts
```diff
@@ -1,28 +1,55 @@
...
 import { Exec } from './exec';
 import { Cp } from './cp';
-import { WebSocketHandler, WebSocketInterface } from './web-socket-handler';
+import { BinaryHandler, WebSocketHandler, WebSocketInterface } from './web-socket-handler';
+import { V1Status } from './api';
+import { randomUUID } from 'crypto';
+import { sleep } from './util';
 
 describe('Cp', () => {
+    let tmpDir: string | undefined;
```
mstruebing, line 19: If you initialize `tmpDir` here with an empty string you don't need to check in every test for it, you could even initialize it with a proper dir:

```suggestion
    let tmpDir: string = `${tmpdir()}/${randomUUID()}`;
```

mstruebing (commented):
Only one very minor comment, otherwise this looks like a solid PR to me.

PR #1888, opened 2024-09-26
chore: update ws

mstruebing:
Update `ws` to version 8.18 because of a security vulnerability and npm dependencies to latest versions according to our version constraints.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "stream-buffers": "^3.0.2",
         "tar": "^7.0.0",
         "tslib": "^2.4.1",
-        "ws": "^8.11.0"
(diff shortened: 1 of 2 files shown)


brendandburns:
This has a ton of updates to package-lock. Are these all related to the `ws` update?

I would have thought that dependabot would be keeping everything up to date.

mstruebing:
They are not all related to `ws` - I ran `npm update` as well. 
If you would feel better with only the `ws` update i will update the PRs accordingly.

mstruebing:
@brendandburns I've updated both PRs to only update `ws` withour running `npm update`.

I would also have thought that dependabot will take care but it since quite a while since this appeared I've decided to do it manually.

PR #1898, opened 2024-09-27
deps: replace `byline` with `node:readline`

cjihrig:
`byline` was being used to read a stream line by line. The core `node:readline` module appears to provide the functionality needed by this package, so remove an external dependency.


src/watch.ts
```diff
@@ -1,5 +1,4 @@
-import byline = require('byline');
-import { RequestOptions } from 'https';
```
cjihrig, line 2: This line is unrelated to this PR, but the code was unused.

Issue #1907, opened 2024-10-04
Inconsistencies in examples


cjihrig:
**Describe the bug**
There are a few inconsistencies/bugs in the examples in the repo:
- There has been some drift between the examples on the `master` branch and those on the `release-1.x` branch.
- Some of the examples use `@kubernetes/client-node`, while others use the build from `dist/`.
- Some use `require()`, while others use `import`. It probably makes sense to migrate to `import`. However, the examples that use `import` should probably be renamed to use a `.mjs` file extension (otherwise they don't run).
- Not really an issue, but the Node core imports can also be updated to use the `node:` scheme.
- Some examples use the `Promise` API directly, while some use `async`/`await`. This is not an issue, but it might be worth moving to `async`/`await` just to keep the examples simpler.

**Client Version**
latest on master and release-1.x branches.

**Server Version**
n/a

**To Reproduce**
View and try to run the various examples.

**Expected behavior**
The examples are consistent and executable.

**Example Code**
n/a

**Environment (please complete the following information):**
n/a

**Additional context**
n/a

mstruebing:
> Some of the examples use @kubernetes/client-node, while others use the build from dist/.

The examples should use the files from the `dist` directory so that you can directly execute them with the current code.
Otherwise it would try to use the published npm package.

> Some use require(), while others use import. It probably makes sense to migrate to import. However, the examples that use import should probably be renamed to use a .mjs file extension (otherwise they don't run).

I do not really have an opinion here, but I feel like that `import` is more modern, therefore I would be okay with updating them.

> Not really an issue, but the Node core imports can also be updated to use the node: scheme.

Could you explain what the advantage would be? Is there any difference at all? I would guess that if there is a package out there (or in your monorepo/whatever) a package with the same would take precedence whereas with `node:` this issue doesn't occur?

Some examples use the Promise API directly, while some use async/await. This is not an issue, but it might be worth moving to async/await just to keep the examples simpler.

I vote for using `async/await` at least in the examples from the `release-1.x` branch.

cjihrig:
> Could you explain what the advantage would be? Is there any difference at all? I would guess that if there is a package out there (or in your monorepo/whatever) a package with the same would take precedence whereas with node: this issue doesn't occur?

[...]

PR #1922, opened 2024-10-09
refactor Node core module imports

cjihrig:
This commit is a first pass at cleaning up the imports in the project. It moves to the 'node:' scheme for core modules, removes some unnecessary imports, and removes some unused imports.

Refs: https://github.com/kubernetes-client/javascript/issues/1907

### DIFF
--- a/src/attach.ts
+++ b/src/attach.ts
@@ -1,6 +1,6 @@
 import WebSocket = require('isomorphic-ws');
-import querystring = require('querystring');
-import stream = require('stream');
+import querystring = require('node:querystring');
+import stream = require('node:stream');
 
 import { KubeConfig } from './config';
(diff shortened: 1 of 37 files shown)


src/attach.ts
```diff
@@ -1,6 +1,6 @@
 import WebSocket = require('isomorphic-ws');
 import querystring = require('querystring');
```
mstruebing, line 2: Couldn't we use `import querystring = require('node:querystring');` here is well?

mstruebing (commented):
Thanks so much. 🚀 
There is just one small missing `node:` import I guess.

src/attach.ts
cjihrig: Updated. Thanks!

mstruebing:
/lgtm
/approve

Thanks 🚀

PR #1923, opened 2024-10-10
refactor imports in src

cjihrig:
This commit uses `import` instead of `require()` in all cases, except when importing JSON files. It also removes a number of '* as' imports that appear to be unnecessary.

A few notes:
- All of the dependencies changed here seem to have a default export except for `tar`, so you'll still see `* as tar` in one place in the code.
- I haven't touched the examples yet because there are still a few issues that need to be figured out.
  - To use `@kubernetes/client-node` or `dist`?
  - To move to ESM the files with a `.js` extension probably won't run directly. Is this OK, or should they be renamed to `.mjs`?
  - It would be nice to be able to refactor `import * as k8s from` to just `import k8s from`, but I think some work would be necessary to add a default export.

### DIFF
--- a/src/attach.ts
+++ b/src/attach.ts
@@ -1,6 +1,6 @@
-import WebSocket = require('isomorphic-ws');
-import querystring = require('node:querystring');
-import stream = require('node:stream');
+import WebSocket from 'isomorphic-ws';
+import querystring from 'node:querystring';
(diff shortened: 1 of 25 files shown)


brendandburns:
/lgtm
/approve

Thanks!

Issue #1926, opened 2024-10-13
Fix CVE in jsonpath-plus dep


timd73:
**Describe the bug**
The `jsonpath-plus` dependency contains a critical CVE:
[CVE-2024-21534](https://www.cve.org/CVERecord?id=CVE-2024-21534)

It is fixed (as least as far as "Snyk" scans are concerned) in `10.0.0`

The request is to update the dependency in `@kubernetes/client-node`

** Client Version **
`0.22.0` (and earlier)

** Server Version **
e.g. `1.19.1`

**Environment (please complete the following information):**
 - NodeJS client

cjihrig:
The dependency is updated on the master branch in https://github.com/kubernetes-client/javascript/pull/1916, and on the release-1.x branch in https://github.com/kubernetes-client/javascript/pull/1914.

It looks like a release for the 0.x line is underway based on https://github.com/kubernetes-client/javascript/pull/1920.

mstruebing:
@brendanburns I've tried to run the release but the workflow isn't successful.
I suspect an outdated authentication token or similar. 

The error message is:

```
npm error 404 Not Found - PUT https://registry.npmjs.org/@kubernetes%2fclient-node - Not found                                                              npm error 404  '@kubernetes/client-node@0.22.1' is not in this registry.          
```

I get this for the `master` as well as for the `release-1.x` branch.
Do you know what's wrong?

Here is the log for the `main` branch: https://github.com/kubernetes-client/javascript/actions/runs/11317557777/job/31471135586

brendandburns:
This is due to the npm token expiring, I will generate a new one and re-add it to the gh actions.

brendandburns:
@mstruebing this should now work.

rlsf:
can we get also a release of the 1.x branch?

timd73:
And what's the ETA on the release of the 0.x line?

mstruebing:
@brendandburns thanks, npm release worked fine.

I also released `1.0.0-rc7` as the `next` tag on npm:

* https://www.npmjs.com/package/@kubernetes/client-node/v/0.22.1
* https://www.npmjs.com/package/@kubernetes/client-node/v/1.0.0-rc7

PR #1935, opened 2024-10-15
ci: set github token permission to write in release workflow

mstruebing:
In response to #1926 - the release to npm did work but the git-tagging wasn't successful so I did that manually.
I suspect missing github token permissions, this __should__ fix the issue, but I did not test it to be honest.

If we see this working, I can create a PR for the `release-1.x` branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,8 @@
 name: Release
 
+permissions:
+  contents: write
+
 on:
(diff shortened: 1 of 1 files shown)


brendandburns:
/lgtm
/approve

I'm not 100% sure this will fix things because of https://github.com/kubernetes/org/issues/5126 but we'll see.

PR #1953, opened 2024-10-18
override whatwg-url dependency

cjihrig:
The generated client on the release-1.x branch relies on the node-fetch@2.x module, which depends on whatwg-url@5.0.0, which uses the Node core punycode module. Node's punycode module was runtime deprecated in v21, which causes this module to print a deprecation warning when used on newer versions of Node.

whatwg-url@9.0.0 stopped using punycode directly, but continued to use it indirectly via its own tr46 dependency. The problematic use of punycode was finally removed in whatwg-url@12.0.1.

node-fetch@2.x claims backwards compatibility to Node v4, but the fixed version of whatwg-url only claims compatibility back to Node 12. For this reason, the node-fetch project has stated that they will not address this issue. For reference, Node v4 went EOL in 2018, and Node v12 when EOL in early 2022. Node 18 is currently the oldest supported version, and Node v23 was released this week.

Ideally, the generator will move to a newer version of node-fetch, native fetch, or even the undici module that implements fetch in core. Until that happens, this module can override whatwg-url and silence the deprecation warning.

It's also worth noting that this punycode deprecation is not the same one as the punycode deprecation on the master branch.

This change has been tested locally on v18.0.0.

Refs: https://github.com/nodejs/node/pull/47202
Refs: https://github.com/node-fetch/node-fetch/pull/1793

(Sorry for the excessive commit message, but there is a lot of context to capture and maybe it will help someone looking through the git history one day.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2012,6 +2012,14 @@
                 "node": ">= 8"
             }
         },
+        "node_modules/punycode": {
+            "version": "2.3.1",
+            "resolved": "https://registry.npmjs.org/punycode/-/punycode-2.3.1.tgz",
+            "integrity": "sha512-vYt7UD1U9Wg6138shLtLOvdAu+8DsC/ilFtEVHcH+wydcSpNE20AfSOduf6MkRFahL5FY7X1oU7nKVZFtfq8Fg==",
+            "engines": {
+                "node": ">=6"
+            }
+        },
         "node_modules/punycode.js": {
             "version": "2.3.1",
             "resolved": "https://registry.npmjs.org/punycode.js/-/punycode.js-2.3.1.tgz",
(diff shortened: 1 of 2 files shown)


brendandburns:
/lgtm
/approve

I hope that Dependabot continues to update this dependency...

cjihrig:
> I hope that Dependabot continues to update this dependency

I'm not 100% sure how Dependabot and overrides might work together. However, prior to this change, `whatwg-url` was at 5.0.0, while the latest on npm is 14.0.0. node-fetch won't be updating it on their end per the linked issue.

cjihrig:
Unfortunately, the punycode warning is present in the v1.0.0 release. It looks like [overrides are only respected in a project's root `package.json`](https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md#only-root-package-may-contain-overrides), so the change in this PR is not considered when end users install this package.

PR #1965, opened 2024-10-23
refactor example code

cjihrig:
This commit addresses most of the issues raised in #1907:

- Migrate from require() to import.
- Migrate from Promises API to async/await.
- Use @kubernetes/client-node instead of dist/
- Add examples specific tslint to allow console.

This commit does not sync the examples from the master branch to the release-1.x branch.


examples/typescript/tsconfig.json
```diff
@@ -1,7 +1,12 @@
 {
     "extends": "../../tsconfig.json",
+    "compilerOptions": {
```
cjihrig, line 3: These changes allow top level `await` and the use of `@kubernetes/client-node` in the TypeScript examples.

brendandburns:
/lgtm
/approve

Thanks for the cleanups!

PR #1972, opened 2024-10-24
Switch from ts-node to tsx for tests

brendandburns:
This should clean up long standing problems with tests + esm module loading in certain cases.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,6 +1,5 @@
 {
-  // Specify "require" for CommonJS
-  "require": "ts-node/register",
+  "require": "tsx",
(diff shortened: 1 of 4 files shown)


cjihrig:
It looks like `ts-node` is also referenced [here](https://github.com/kubernetes-client/javascript/blob/master/test/mocha.opts#L1).

brendandburns:
@cjihrig I removed that file, I don't think it is used. (tests are passing without it)

mstruebing:
/lgtm
/approve

Removing not used files and code is always good 🙌 
Thanks 🚀

PR #1974, opened 2024-10-24
Don't concatenate shell arguments.

brendandburns:
Fixes [https://github.com/kubernetes-client/javascript/security/code-scanning/17](https://github.com/kubernetes-client/javascript/security/code-scanning/17)

To fix the problem, we should avoid using `proc.execSync` with a dynamically constructed command string. Instead, we can use `child_process.execFileSync`, which takes the command and its arguments as separate parameters, thus avoiding shell interpretation of the arguments. This change will ensure that the arguments are passed safely to the command without the risk of command injection.

**Steps to fix:**
1. Replace the use of `proc.execSync` with `proc.execFileSync`.
2. Split the command and its arguments into separate parameters for `execFileSync`.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/gcp_auth.ts
+++ b/src/gcp_auth.ts
@@ -66,7 +66,14 @@ export class GoogleCloudPlatformAuth implements Authenticator {
         if (!cmd) {
             throw new Error('Token is expired!');
         }
-        const args = config['cmd-args'] ? config['cmd-args'].split(' ') : [];
+        const args = (config['cmd-args'] ? config['cmd-args'].split(' ') : []).map(
+            (arg: string): string => {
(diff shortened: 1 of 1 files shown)


mstruebing (commented):
Seems like the tests are failing for a valid reason.

I've pushed a commit which fixes one of the failing tests.
I did not look to deep into it, but the "fix" seems hacky and weird to me.
Just want to save you some time if you have a look later :)

brendandburns:
@mstruebing I added a fix which I think is more correct, removing leading/trailing quote marks for each arg.

mstruebing:
/lgtm
/approve

Thanks 🚀

Issue #1960, opened 2024-10-21
Fix CVE in jsonpath-plus (again)


soniqua:
**Describe the bug**
The jsonpath-plus dependency contains a critical CVE, even after upgrading to `10.0.0`: [CVE-2024-21534](https://www.cve.org/CVERecord?id=CVE-2024-21534)

The library has been fixed as of version [10.0.7](https://github.com/JSONPath-Plus/JSONPath/commit/b70aa713553caf838a63bac923195a5bc541fd72) or higher.

There's an [open dependabot PR](https://github.com/kubernetes-client/javascript/pull/1957) for resolution.

** Client Version **
`0.22.1`

**Environment (please complete the following information):**
 - NodeJS Client

**Additional context**
- Previous fix: https://github.com/kubernetes-client/javascript/issues/1926

timd73:
@mstruebing 
Any idea on an ETA for a `0.22.x` patch release?

brendandburns:
No current eta, but probably within the week. fwiw, this library only uses this library for kubeconfig file loading. If you have malicious JSONPath in your kubeconfig, you have far worse problems than this RCE.

Also, unless your kubeconfig contains a jsonpath value you're not impacted by this CVE.

timd73:
@brendandburns  thanks for the update/ETA, and for the added info.

FWIW, I am not concerned about the actual vulnerability, as I'd gathered that it's not truly exploitable. Rather it's about making the vulnerability scanners happy (remind me, do we work for them, or they for us?).

But good to know about how it impacts this library - thanks again!

brendandburns:
@timd73 just pushed 0.22.2 to npm with the revised dependency.

DeveloperAlan:
Hey I've noticed in our vulnerability scanners that this issues still persists in the `0.22.2` version of the package.
![Screenshot 2024-11-06 at 10 15 14 am](https://github.com/user-attachments/assets/a5ae06dd-b4b6-41a8-a6e9-13802c5e272c)

Think this package would also need to update the `package.json` to `^10.1.0` since the scanner looks at the package.json file for vulnerabilities

Thanks!

mstruebing:
That actually makes sense as it seems like we also don't publish our `package-lock.json` file.
I don't think for a library it makes sense to publish a lock file, but in contrast that means we need to update our version ranges.

PR #1994, opened 2024-10-30
Rev to new patch version.

brendandburns:
Prep for a patch release to fix: #1960 

We should really move to a branch and cherry-pick model vs. releasing from HEAD but oh well :)


mstruebing:
/lgtm
/approve

we should!

PR #1995, opened 2024-10-31
Improve watch performance.

### LINKED ISSUE #1961: ListWatch initial list performance is slow
**Describe the bug**
When a list watch is created (or any other time `listWatch.resourceVersion === ''`), the ListWatch will attempt to bootstrap its list of objects. Insertion of items from the API Server's List response into `ListWatch.objects` is n^2, and can be very slow for very large quantities of objects (on the order of 100K).

In particular:
1. `ListWatch.doneHandler` invokes `ListWatch.addOrUpdateItems` with all new items. [code](https://github.com/kubernetes-client/javascript/blob/release-1.0.0-rc6.1/src/cache.ts#L144)
2. `ListWatch.addOrUpdateItems` iterates over all items returned from the API Server and calls `addOrUpdateObject` for each item. [code](https://github.com/kubernetes-client/javascript/blob/release-1.0.0-rc6.1/src/cache.ts#L166)
3. `addOrUpdateObject` calls `findKubernetesObject` which calls `objects.findIndex` which again goes over all items. [code](https://github.com/kubernetes-client/javascript/blob/release-1.0.0-rc6.1/src/cache.ts#L249)

** Client Version **
`1.0.0-rc6`

** Server Version **
`1.29.8` (Although I don't think this is relevant)

**To Reproduce**
I added a shim into cache.js in the @kubernetes/client-node package to keep track of how many times a object in cache was accessed via `findKubernetesObject`.
```javascript
function findKubernetesObject(objects, obj) {
    return objects.findIndex((elt) => {
        ++global.accesses;   // This is the only addition
        return isSameObject(elt, obj);
    });
}
```

_test.mjs_
```javascript

const apiUrl = '/apis/v1/namespaces/default/configmaps';

global.accesses = 0;

const kc = new k8s.KubeConfig(); kc.loadFromDefault();

const watch = new k8s.Watch(kc);
const client = kc.makeApiClient(k8s.CoreV1Api);

const lw = new k8s.ListWatch(
    apiUrl,
    watch,
    async () => {
        const x = await client.listNamespacedConfigMap({namespace: 'default'});
        console.log('fetched items', x.items.length);
        return x;
    },
    false
);

const start = performance.now();
await lw.start()
console.log('elapsed', performance.now() - start);
console.log('object accesses', accesses)
```

Test context
```shell
$ minikube start
...
$ seq 50000 | xargs -P 100 -I'{}' kubectl create cm cm-'{}'
...
$ node test.mjs
fetched items 50004
elapsed 12140.640042
object accesses 2500350012
```

**Expected behavior**
It seems reasonable that insertion is O(n) if possible.

** Example Code**
N/A

 - OS: Linux
 - NodeJS Version: 20
 - Cloud runtime: EKS + Minikube

brendandburns:
FIxes #1961 

This is a port of #780 to the 1.x release branch.

cc @mstruebing

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -293,7 +293,6 @@ function isSameVersion<T extends KubernetesObject>(o1: any, o2: T): boolean {
     );
 }
 
-
 // Public for testing.
 export function deleteObject<T extends KubernetesObject>(
     objects: CacheMap<T>,
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -20,8 +20,11 @@ export interface ObjectCache<T> {
     list(namespace?: string): ReadonlyArray<T>;
 }
 
+// exported for testing
+export type CacheMap<T extends KubernetesObject> = Map<string, Map<string, T>>;
+
 export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, Informer<T> {
-    private objects: T[] = [];
+    private objects: CacheMap<T> = new Map();
     private resourceVersion: string;
     private readonly indexCache: { [key: string]: T[] } = {};
     private readonly callbackCache: { [key: string]: (ObjectCallback<T> | ErrorCallback)[] } = {};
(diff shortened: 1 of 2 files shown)


mstruebing:
I've ran prettier for you, this should make CI pass 🤞

brendandburns:
Thanks for fixing the formatting! Somewhere along the line, our git hooks stopped working (at least in my environment) we should look into that :)

Issue #2023, opened 2024-11-12
Build and Deploy Docs is broken


clayrosenthal:
**Describe the bug**
The [build and deploy docs action](https://github.com/kubernetes-client/javascript/actions/workflows/deploy-docs.yml) appears to have been broken for about a year.

** Client Version **
0.22.2

** Server Version **
N/A

**To Reproduce**
1. Run the github action
2. Action fails

**Expected behavior**
Docs to be generated and published.

** Example Code**
[Your workflow](https://github.com/kubernetes-client/javascript/blob/master/.github/workflows/deploy-docs.yml)

**Environment (please complete the following information):**
 - OS: ubuntu 
 - NodeJS Version 20
 - Cloud runtime Github Action runner

**Additional context**
I want to read the docs and was bummed they aren't updated anymore.

Appears in the [action repo](https://github.com/JamesIves/github-pages-deploy-action/issues/1285) there is a fix by adding write permissions:
```
permissions:
  contents: write
```

brendandburns:
fwiw, these docs aren't that good, since they are generated from the generated APIs. You're better off reading https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/

Closing, since the PR merged.

passcod:
I am not in fact "better off" when I'm trying to find API information specific to these Typescript bindings and not the k8s API in general, and the linked PR merged with comments specifically saying "it is still not working" due to the CLA thing blocking the build.

At this point the docs are two years out of date. If you're not going to fix them, maybe remove the link from the readme?

PR #2024, opened 2024-11-12
Add write permissions to deploy docs

clayrosenthal:
Fixing #2023, based on the [repo](https://github.com/JamesIves/github-pages-deploy-action/issues/1285) for the action.

> GitHub changed new repository default settings and made [workflow permissions read-only by default](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/). The fix was to add write permissions to the workflow:
```
permissions:
  contents: write
```

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - master
+permissions:
+  contents: write
 jobs:
   build-and-deploy-docs:
     runs-on: ubuntu-latest


mstruebing:
/lgtm
/approve

Thanks 🚀

mstruebing:
It is still not working because of easy CLA.
I'm not sure if there is any way around this without a private access token.

clayrosenthal:
I'm not too familiar with easy CLA, but can you change a setting to allow actions to bypass the requirement? Or make the CI bot considered having signed the CLA?

PR #2025, opened 2024-11-13
bump jsonpath-plus

liviarett:


mstruebing:
/lgtm
/approve

Thanks 🚀
We should keep in mind to bump the range manually when updating dependencies because of CVEs.

PR #2031, opened 2024-11-18
chore: bump jsonpath-plus allowed version range

mstruebing:
we are getting more and more reports about this as we allow a version where this CVE exists.
This should silent these reports :)


timd73:
@mstruebing  and will there be a new release with this?

davidgamero (approved):
lgtm

PR #2035, opened 2024-11-19
Add Health class to call health check endpoints on current context

feloy:
Add an Health class with 2 public methods `livez` and `readyz` to check the health check endpoints of the current context.

If these endpoints are not served (status 404), the deprecated `healthz` is checked. If this endpoint `healthz` is also not served (status 404), the cluster is considered healthy, else its result is used.

### DIFF
--- a/src/health.ts
+++ b/src/health.ts
@@ -1,5 +1,4 @@
 import fetch from 'node-fetch';
-import { AbortSignal } from 'node-fetch/externals';
 import { KubeConfig } from './config';
 import { RequestOptions } from 'node:https';
 
@@ -30,8 +29,9 @@ export class Health {
 
         const requestURL = new URL(cluster.server + path);
         const requestInit = await this.config.applyToFetchOptions(opts);
-        const controller = new AbortController();
-        requestInit.signal = controller.signal as AbortSignal;
+        if (opts.signal) {
+            requestInit.signal = opts.signal;
+        }
         requestInit.method = 'GET';
 
         try {
--- a/src/health.ts
+++ b/src/health.ts
@@ -1,6 +1,5 @@
 import fetch from 'node-fetch';
 import { AbortSignal } from 'node-fetch/externals';
-import { URL } from 'url';
 import { KubeConfig } from './config';
 import { RequestOptions } from 'node:https';
 
--- a/src/health.ts
+++ b/src/health.ts
@@ -1,0 +1,57 @@
+import fetch from 'node-fetch';
+import { AbortSignal } from 'node-fetch/externals';
+import { URL } from 'url';
+import { KubeConfig } from './config';
+import { RequestOptions } from 'https';
+
+export class Health {
+    public config: KubeConfig;
+
+    public constructor(config: KubeConfig) {
+        this.config = config;
+    }
+
+    public async readyz(opts: RequestOptions): Promise<boolean> {
+        return this.check('/readyz', opts);
+    }
+
+    public async livez(opts: RequestOptions): Promise<boolean> {
+        return this.check('/livez', opts);
+    }
+
+    private async healthz(opts: RequestOptions): Promise<boolean> {
+        return this.check('/healthz', opts);
+    }
+
+    private async check(path: string, opts: RequestOptions): Promise<boolean> {
+        const cluster = this.config.getCurrentCluster();
+        if (!cluster) {
+            throw new Error('No currently active cluster');
+        }
+
+        const requestURL = new URL(cluster.server + path);
+        const requestInit = await this.config.applyToFetchOptions(opts);
+        const controller = new AbortController();
+        requestInit.signal = controller.signal as AbortSignal;
+        requestInit.method = 'GET';
+
+        return await fetch(requestURL.toString(), requestInit)
+            .then((response) => {
+                const status = response.status;
+                if (status === 200) {
+                    return true;
+                } else if (status === 404) {
+                    if (path === '/healthz') {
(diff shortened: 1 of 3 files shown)


src/health.ts
```diff
@@ -0,0 +1,57 @@
+import fetch from 'node-fetch';
+import { AbortSignal } from 'node-fetch/externals';
+import { URL } from 'url';
+import { KubeConfig } from './config';
+import { RequestOptions } from 'https';
```
cjihrig, line 5: Can you please use `node:https` here.

src/health.ts
```diff
@@ -0,0 +1,57 @@
...
+            throw new Error('No currently active cluster');
+        }
+
+        const requestURL = new URL(cluster.server + path);
+        const requestInit = await this.config.applyToFetchOptions(opts);
+        const controller = new AbortController();
+        requestInit.signal = controller.signal as AbortSignal;
+        requestInit.method = 'GET';
+
+        return await fetch(requestURL.toString(), requestInit)
```
cjihrig, line 38: Is there any benefit to mixing `async`/`await` style with the `Promises` API? If not, can you use `async`/`await` for simplicity/readability?

feloy: I got inspiration from `log.ts` which is using this pattern, but no, I cannot see any benefit. I changed to `async`/`await`.

feloy:
Thanks @cjihrig for your review. I've made the changes, and added a unit test on an uncovered case

src/health.ts
```diff
@@ -0,0 +1,56 @@
+import fetch from 'node-fetch';
+import { AbortSignal } from 'node-fetch/externals';
+import { URL } from 'url';
```
cjihrig, line 3: I missed this the first time around, but [`URL`](https://nodejs.org/api/globals.html#url) is a global that doesn't need to be explicitly imported.

cjihrig (commented):
LGTM, thanks!

src/health.ts
```diff
@@ -0,0 +1,55 @@
...
+        const controller = new AbortController();
+        requestInit.signal = controller.signal as AbortSignal;
+        requestInit.method = 'GET';
+
+        try {
+            const response = await fetch(requestURL.toString(), requestInit);
+            const status = response.status;
+            if (status === 200) {
+                return true;
+            } else if (status === 404) {
```
brendandburns, line 42: remove `else` since there is `return` in the preceding case.

src/health.ts
```diff
@@ -0,0 +1,55 @@
...
+                if (path === '/healthz') {
+                    // /livez/readyz return 404 and healthz also returns 404, let's consider it is live
+                    return true;
+                }
+                return this.healthz(opts);
+            } else {
```
brendandburns, line 48: remove `else` since all paths above end in `return`

src/health.ts
```diff
@@ -0,0 +1,55 @@
...
+                return false;
+            }
+        } catch {
+            return false;
```
brendandburns, line 52: I think that we should `reject` the promise here instead of returning false.

brendandburns:
Thanks for the PR. Minor comments, then LGTM

(and special thanks for adding tests!)

src/health.ts
```diff
@@ -0,0 +1,55 @@
...
+            if (status === 404) {
+                if (path === '/healthz') {
+                    // /livez/readyz return 404 and healthz also returns 404, let's consider it is live
+                    return true;
+                }
+                return this.healthz(opts);
+            }
+            return false;
+        } catch {
+            throw new Error('Error occurred in health request');
```
cjihrig, line 52: You can attach the original error like this, while still throwing a custom error.

```suggestion
        } catch (err) {
            throw new Error('Error occurred in health request', { cause: err });
```

feloy: Changing the tsconfig with the following would fix the problem, but I'm not sure of the potential side-effects:

```
  "compilerOptions": {
    "lib": ["es2022"],
```

cjihrig: Ah, ok. There must be a TypeScript setting that needs to be tweaked in the project since [`cause` is a newer addition to the spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause). I will resolve my comment.

feloy:
I also made a change to how the abortcontroller signal was managed. Previously, the abort controller passed to fetch was not accessible. I changed to use the one (if any) passed as parameter, so the caller can abort the request

feloy:
setting as draft, the time to make some test, to find the best way for the user to handle abort

feloy:
I made one more change to throw the original error when it is an AbortError, so the caller can detect that the call has been aborted

It is ready for review again

Sorry for all these afterward changes

brendandburns:
Looks good to me. I'd like to get an LGTM from @cjihrig also since he reviewed earlier and then we can merge.

cjihrig (approved):

feloy:
/assign mstruebing


> [APPROVALNOTIFIER] This PR is **NOT APPROVED**
> 
> This pull-request has been approved by: _[cjihrig](https://github.com/kubernetes-client/javascript/pull/2035#pullrequestreview-2453348549)_, _[feloy](https://github.com/kubernetes-client/javascript/pull/2035#)_ **Once this PR has been reviewed and has the lgtm label**, please assign [mstruebing](https://github.com/mstruebing) for approval. For more information see [the Kubernetes Code Review Process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process).
> 

/assign mstruebing

Issue #1407, opened 2023-11-09
cpFromPod / cpToPod is missing --retries flag


fpm-peter:
**Describe the bug**
`cpFromPod` / `cpToPod` silently fail on larger files, resulting in incomplete files.

** Client Version **
`0.20.0`

** Server Version **
`1.25.7+rke2r1`

**To Reproduce**
upload a larger file, at least 200MB. The kubernetes issue has gone into detail about network settings a.s.o.

**Expected behavior**
`cpFromPod` / `cpToPod` should try to upload the whole file or error out on premature disconnect.

**Environment (please complete the following information):**
 - OS: Linux
 - NodeJS Version 20

**Additional context**
Has been identified and 'fixed' in `kubectl cp` https://github.com/kubernetes/kubernetes/issues/60140 with the `--retries` flag.

brendandburns:
the logic in `kubectl` is pretty complicated. Someone will need to port that logic into the copy logic in this client.

Knickkennedy:
I noticed that in [kubectl cp](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L387) that it looks like the tar.Read just uses a byte stream to capture where it left off and continue. Could that logic be replicated here? Basically just create an io pipe and keep track of where we are in the bytes? Also sorry if this is out of left field, I'm new to contributing to Kubernetes in any capacity.

PR #2037, opened 2024-11-19
Add support for the v5 protocol

brendandburns:
Support for the v5 streaming protocol, described here:

https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/4006-transition-spdy-to-websockets#proposal-new-remotecommand-sub-protocol-version---v5channelk8sio

May help fix this: https://github.com/kubernetes-client/javascript/issues/1407

### DIFF
--- a/src/web-socket-handler.ts
+++ b/src/web-socket-handler.ts
@@ -42,13 +42,11 @@ export class WebSocketHandler implements WebSocketInterface {
     }
 
     public static closeStream(streamNum: number, streams: StreamInterface): void {
-        console.log('Closing stream: ' + streamNum);
         switch (streamNum) {
             case WebSocketHandler.StdinStream:
                 streams.stdin.pause();
                 break;
             case WebSocketHandler.StdoutStream:
-                console.log('closing stdout');
                 streams.stdout.end();
                 break;
             case WebSocketHandler.StderrStream:
@@ -249,7 +247,6 @@ export class WebSocketHandler implements WebSocketInterface {
                 } else if (data instanceof Buffer) {
                     const streamNum = data.readUint8(0);
                     if (streamNum === WebSocketHandler.CloseStream) {
-                        console.log('Closing stream!');
                         WebSocketHandler.closeStream(data.readInt8(1), this.streams);
                     }
                     if (binaryHandler && !binaryHandler(streamNum, data.subarray(1))) {
--- a/src/web-socket-handler.ts
+++ b/src/web-socket-handler.ts
@@ -4,7 +4,13 @@ import stream = require('stream');
 import { V1Status } from './api';
 import { KubeConfig } from './config';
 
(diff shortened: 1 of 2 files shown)


src/web-socket-handler.ts
```diff
@@ -17,12 +23,39 @@ export interface WebSocketInterface {
...
     public static readonly StatusStream: number = 3;
     public static readonly ResizeStream: number = 4;
+    public static readonly CloseStream: number = 255;
+
+    public static supportsClose(protocol: string): boolean {
+        return protocol === 'v5.channel.k8s.io';
+    }
+
+    public static closeStream(streamNum: number, streams: StreamInterface): void {
+        console.log('Closing stream: ' + streamNum);
```
cjihrig, line 45: There are three `console.log()`s left over.

src/web-socket-handler_test.ts
```diff
@@ -2,6 +2,7 @@ import { promisify } from 'util';
 import { expect } from 'chai';
 import WebSocket = require('isomorphic-ws');
 import { ReadableStreamBuffer, WritableStreamBuffer } from 'stream-buffers';
+import stream = require('stream');
```
cjihrig, line 5:
```suggestion
import stream from 'node:stream';
```

src/web-socket-handler_test.ts
```diff
@@ -303,6 +304,107 @@ describe('WebSocket', () => {
...
+                sent = data;
+            },
+            close: () => {},
+        } as WebSocket;
+        const stdinStream = new ReadableStreamBuffer();
+        WebSocketHandler.handleStandardInput(ws, stdinStream);
+        stdinStream.emit('end');
+        expect(sent).to.not.be.null;
+        expect(sent!.readUint8(0)).to.equal(255); // CLOSE signal
+        expect(sent!.readUInt8(1)).to.equal(0); // Stdin stream is #0
```
cjihrig, line 404: Is there a reason one of these uses `readUint8()` and the other uses `readUInt8()` (note difference in casing of the `i`).

src/web-socket-handler.ts
brendandburns: ugh, so embarrassing to leave test printfs in the PR :)

fixed.

src/web-socket-handler_test.ts
brendandburns: Converted all to `readUint8` The Internet seems to be ambiguous on which is the right casing. Both appear to work...

src/web-socket-handler_test.ts
brendandburns: Converted all to readUint8 The Internet seems to be ambiguous on which is the right casing. Both appear to work...

mstruebing:
/lgtm
/approve

Learned a bit about streaming while reviewing this PR 🎉

cjihrig:
Does this make sense for the v1.x branch as well?

PR #2038, opened 2024-11-20
Resolve copy promises

sjawhar:
The promise returned by `cpToPod` never resolves. Followed the example from here to fix: https://github.com/kubernetes-client/javascript/issues/982#issuecomment-1432485735

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -35,6 +35,7 @@ export class Cp {
         command.push(srcPath);
         const writerStream = fs.createWriteStream(tmpFileName);
         const errStream = new WritableStreamBuffer();
+        let resolved = false;
         return new Promise<void>((resolve, reject) => {
             this.execInstance
                 .exec(
@@ -60,7 +61,10 @@ export class Cp {
                                 file: tmpFileName,
                                 cwd: tgtPath,
                             });
-                            resolve();
+                            if (!resolved) {
+                                resolved = true;
+                                resolve();
+                            }
                         } catch (e) {
                             reject(e);
                         }
@@ -68,7 +72,10 @@ export class Cp {
                 )
                 .then((conn) => {
                     conn.on('close', () => {
-                        resolve();
+                        if (!resolved) {
+                            resolved = true;
+                            resolve();
+                        }
                     });
                 })
                 .catch(reject);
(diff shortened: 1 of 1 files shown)


src/cp.ts
```diff
@@ -66,6 +66,11 @@ export class Cp {
                         }
                     },
                 )
+                .then((conn) => {
+                    conn.onclose = (event) => {
+                        resolve();
```
brendandburns, line 71: There is a resolve above at line 63. I don't think that this resolve is required. In fact I'm not sure it's what we want, because this will resolve when the websocket is returned. We want to resolve when the copy completes.

sjawhar: That callback only gets called if `status != null`: https://github.com/kubernetes-client/javascript/blob/8db906a63eff5bf296f3fac4572f83ea3db51283/src/exec.ts#L55

For copies (or maybe for all `exec` calls that use `stdin`), `status == null`, so the callback never gets called, so the promise never resolves. `await`ing calls to `cpToPod` hang forever.

sjawhar: If the copy isn't complete when the websocket is closed, how can the caller ever now that it has completed?

sjawhar:
> Hey @sjawhar thanks for your pr but I think the tests are broken, could you have a look? You can run them locally via `npm test`

Yeah, sorry about that, I had quickly made the edits in the browser to make sure I didn't forget. Now I've actually, like, tested it and stuff 😅

cjihrig:
Is there a test that can be added for this?

sjawhar:
> Is there a test that can be added for this?

I've spend some time on it but didn't figure out the different mocks well enough to understand why the tests were passing before instead of timing out as expected. The lines in the test that send a `V1Status` seem not to match actual behavior. Removing them causes the cp promise to never resolve, so some other change is also needed.

If it helps at all, I've read through the code and set breakpoints to interactively debug and this the issue largely comes down to this:
https://github.com/kubernetes-client/javascript/blob/88da3bd5dbc0c1ab60a61eff6dd4e411efba198c/src/web-socket-handler.ts#L55-L73

It looks to me like when stdin is provided, the websocket is closed without waiting for a `StatusStream` response, so the status callback never gets called. Maybe that'll help you fix the tests, or maybe the right fix is actually upstream of the copy function (maybe #2037 ?).

src/cp.ts
brendandburns: Ah, I missed that you are doing this on the 'close' event. My mistake. But we should probably make sure that we don't resolve the promise twice.

sjawhar: OK, added a check for that in a way that shouldn't cause things to break if changes elsewhere cause this logic to change.

himanshusharma27:
hi please assign this request to me

himanshusharma27:
okh thanks

PR #2041, opened 2024-11-20
update deprecated husky command

cjihrig:
The husky install command is deprecated and prints a message saying so. This commit updates the command according to https://github.com/typicode/husky/issues/1415.


mstruebing:
/lgtm
/approve

Thanks 🚀

PR #2046, opened 2024-11-22
makes exec_auth spawn the command asynchronously

### LINKED ISSUE #2044: exec_auth spawns the command synchronously
**Describe the bug**

exec_auth spawns the command synchronously. 

When the exec command sends information to stdout/stderr to inform the user what to do to authenticate, the program is blocked waiting indefinitely, because the spawn is done synchronously.

 

** Client Version **
1.0.0-rc7

** Server Version **

Azure 1.29.9

**To Reproduce**

- Start an Azure cluster with 'Microsoft Entra ID authentication with Kubernetes RBAC' authn & authz
- get the credentials
- do not execute any request to the cluster, so the token is not cached (or you can remove it: `rm ~/.kube/cache/kubelogin/yourID.json`
- make a request to the API
- the program is blocked executing the `kubelogin` app, which is blocked, waiting for the user to authenticate

**Expected behavior**

The call should be done async
bonus: the information provided by kubelogin should be displayed to the user

** Example Code**

```
import { CoreV1Api, KubeConfig } from "@kubernetes/client-node";

const kc = new KubeConfig();
kc.loadFromDefault();

const k8sApi = kc.makeApiClient(CoreV1Api);
const result = k8sApi.listNamespace();
```

 - OS: MacOS
 - NodeJS Version: v20.15.1
 - Cloud runtime: Azure

**Additional context**

feloy:
This PR makes exec_auth calls `spawn` instead of `spawnSync`.

This is cery useful for a GUI to continue working during the plugin is executed.

Fixes #2044

### DIFF
--- a/src/exec_auth.ts
+++ b/src/exec_auth.ts
@@ -100,11 +100,14 @@ export class ExecAuth implements Authenticator {
             opts = { ...opts, env };
         }
 
-        return new Promise((accept, reject) => {
+        return new Promise((resolve, reject) => {
             let stdoutData: string = '';
             let stderrData: string = '';
+            let savedError: Error | undefined = undefined;
 
             const subprocess = this.execFn(exec.command, exec.args, opts);
+            subprocess.stdout.setEncoding('utf8');
+            subprocess.stderr.setEncoding('utf8');
 
             subprocess.stdout.on('data', (data: Buffer | string) => {
                 stdoutData += data.toString('utf8');
@@ -115,16 +118,26 @@ export class ExecAuth implements Authenticator {
             });
 
             subprocess.on('error', (error) => {
+                savedError = error;
                 throw error;
             });
 
             subprocess.on('close', (code) => {
-                if (code !== 0) {
-                    throw new Error(stderrData);
+                if (savedError) {
+                    reject(savedError);
+                    return;
                 }
-                const obj = JSON.parse(stdoutData) as Credential;
-                this.tokenCache[user.name] = obj;
-                accept(obj);
+                if (code !== 0) {
+                    reject(new Error(stderrData));
+                    return;
+                }
+                try {
+                    const obj = JSON.parse(stdoutData) as Credential;
+                    this.tokenCache[user.name] = obj;
+                    resolve(obj);
+                } catch (error) {
+                    reject(error);
+                }
             });
         });
     }
--- a/src/exec_auth.ts
+++ b/src/exec_auth.ts
@@ -109,17 +109,16 @@ export class ExecAuth implements Authenticator {
             subprocess.stdout.setEncoding('utf8');
             subprocess.stderr.setEncoding('utf8');
 
-            subprocess.stdout.on('data', (data: Buffer | string) => {
-                stdoutData += data.toString('utf8');
+            subprocess.stdout.on('data', (data: string) => {
+                stdoutData += data;
             });
 
-            subprocess.stderr.on('data', (data) => {
-                stderrData += data.toString('utf8');
+            subprocess.stderr.on('data', (data: string) => {
+                stderrData += data;
             });
 
             subprocess.on('error', (error) => {
                 savedError = error;
-                throw error;
             });
 
             subprocess.on('close', (code) => {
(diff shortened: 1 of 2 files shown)


src/exec_auth.ts
```diff
@@ -99,15 +99,33 @@ export class ExecAuth implements Authenticator {
...
-        throw new Error(result.stderr.toString('utf8'));
+
+        return new Promise((accept, reject) => {
+            let stdoutData: string = '';
+            let stderrData: string = '';
+
+            const subprocess = this.execFn(exec.command, exec.args, opts);
+
+            subprocess.stdout.on('data', (data: Buffer | string) => {
+                stdoutData += data.toString('utf8');
```
cjihrig, line 110: I think the `data.toString('utf8')` line here and on line 114 may be subtly incorrect.

With this logic, chunks are emitted as `Buffer`s, which are then converted to utf8 strings. However, utf8 characters can be multiple bytes, and the `Buffer` based logic does not know anything about that. It's possible that a utf8 character gets split across multiple chunks, which results in the wrong final string.

I would suggest updating this code in one of two ways:
- Do all of the chunk collecting with `Buffer`s and then convert everything to utf8 at once in the `'close'` event.
- Use `subprocess.std{out,err}.setEncoding('utf8')` on line 108 so that Node emits `'data'` events with proper utf8 chunks.

src/exec_auth.ts
```diff
@@ -99,15 +99,33 @@ export class ExecAuth implements Authenticator {
...
-            throw result.error;
-        }
-        if (result.status === 0) {
-            const obj = JSON.parse(result.stdout.toString('utf8')) as Credential;
-            this.tokenCache[user.name] = obj;
-            return obj;
-        }
-        throw new Error(result.stderr.toString('utf8'));
+
+        return new Promise((accept, reject) => {
```
cjihrig, line 103:
```suggestion
        return new Promise((resolve, reject) => {
```

src/exec_auth.ts
```diff
@@ -99,15 +99,33 @@ export class ExecAuth implements Authenticator {
...
+            subprocess.stdout.on('data', (data: Buffer | string) => {
+                stdoutData += data.toString('utf8');
+            });
+
+            subprocess.stderr.on('data', (data) => {
+                stderrData += data.toString('utf8');
+            });
+
+            subprocess.on('error', (error) => {
+                throw error;
```
cjihrig, line 118: Should this call `reject(error)` instead of throwing? It might also be worth putting `error` in a variable instead and then handling the `resolve()`/`reject()` logic in the `'close'` event so that you don't have to handle the case where the promise is both resolved and rejected.

src/exec_auth.ts
```diff
@@ -99,15 +99,33 @@ export class ExecAuth implements Authenticator {
...
+            });
+
+            subprocess.on('close', (code) => {
+                if (code !== 0) {
+                    throw new Error(stderrData);
```
cjihrig, line 123: Instead of throwing, maybe `reject(new Error(stderrData))`.

src/exec_auth.ts
```diff
@@ -99,15 +99,33 @@ export class ExecAuth implements Authenticator {
...
+                }
+                const obj = JSON.parse(stdoutData) as Credential;
```
cjihrig, line 125: Since this can throw, consider wrapping it in a `try...catch` and rejecting on error.

src/exec_auth.ts
feloy: ok, thanks, I'll change with the second option

src/exec_auth.ts
feloy: thanks, I'll change to delay the reject to the `close` event

feloy:
Many thanks for your review. I've made the changes and pushed a new commit

src/exec_auth.ts
```diff
@@ -99,15 +99,46 @@ export class ExecAuth implements Authenticator {
...
+                stdoutData += data.toString('utf8');
+            });
+
+            subprocess.stderr.on('data', (data) => {
+                stderrData += data.toString('utf8');
+            });
+
+            subprocess.on('error', (error) => {
+                savedError = error;
+                throw error;
```
cjihrig, line 122: I'm not sure that it still makes sense to throw here.

feloy: my mistake, I forgot to remove it

cjihrig:
LGTM, thanks!

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #2052, opened 2024-11-25
src: remove AbortSignal import

cjihrig:
`AbortSignal` is available globally in all supported versions of Node. These imports were causing problems when attempting to upgrade TypeScript settings because apparently `'node-fetch/externals'` exists in the types package, but not in the actual node-fetch module.


mstruebing:
/lgtm
/approve

Thanks 🚀

PR #2053, opened 2024-11-25
src: replace Promise chains with async/await

cjihrig:
This helps with readability of the code. It also removed some Promises that were not not needed.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -136,36 +136,35 @@ export class Log {
         requestInit.signal = controller.signal as AbortSignal;
         requestInit.method = 'GET';
 
-        await fetch(requestURL.toString(), requestInit)
-            .then((response) => {
-                const status = response.status;
-                if (status === 200) {
-                    // TODO: the follow search param still has the stream close prematurely based on my testing
-                    response.body.pipe(stream);
-                } else if (status === 500) {
-                    const v1status = response.body as V1Status;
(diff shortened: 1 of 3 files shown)


src/log.ts
```diff
@@ -136,36 +136,35 @@ export class Log {
...
+            } else {
+                throw new ApiException<undefined>(
+                    status,
+                    'Error occurred in log request',
+                    undefined,
+                    normalizeResponseHeaders(response),
+                );
+            }
+        } catch (err: any) {
+            throw new ApiException<undefined>(err, 'Error occurred in log request', undefined, err);
```
cjihrig, line 166: Note - this line is ported from the previous `catch()` call, but I don't think passing `err` as the first parameter is actually correct. `catch()` typed `err` as `any`. When I changed to a `try...catch`, that typing was not there and TypeScript complained that the first argument should be a numeric code (probably similar to how `src/metrics.ts` does it).

brendandburns: Yeah, I'm on board with switching this to match metric.ts

brendandburns:
This basically looks good to me, one response to your query, but happy to merge as-is if you prefer,

cjihrig:
I've updated the `catch` block to work like the one in the metrics file.

mstruebing:
/lgtm
/approve

Thanks 🚀

PR #2091, opened 2024-12-05
Proxy url

krmodelski:
Add http/https and socks proxy support


brendandburns:
This looks good to me, thanks for adding tests.

Once this merges, you may want to cherry-pick it into the `release-1.x` since that is where we are going to focus development going forward.

mstruebing:
Code looks good to me as well, I did not test it though.
Therefore I'm hesitant to approve right now.
I would need a couple of days to test, if anyone else has time it would be much appreciated. 🙏

brendandburns:
@mstruebing I'm personally ok to merge w/o e2e testing. I'm confident that this won't break anyone who doesn't have `proxy-url` in their kubeconfig, and if it breaks people with `proxy-url` in their config, we can take those issues as they come, since it doesn't work at all today, it seems low-risk.

wdyt?

mstruebing:
Sounds good, considering that this is a new feature it shouldn't break anything you are right.

/lgtm
/approve

PR #2105, opened 2024-12-12
Generate code for 1.32

brendandburns:
cc @cjihrig


src/gen/apis/AdmissionregistrationApi.ts
```diff
@@ -1,16 +1,16 @@
 // TODO: better import syntax?
-import {BaseAPIRequestFactory, RequiredError, COLLECTION_FORMATS} from './baseapi.js';
-import {Configuration} from '../configuration.js';
-import {RequestContext, HttpMethod, ResponseContext, HttpFile, HttpInfo} from '../http/http.js';
+import {BaseAPIRequestFactory, RequiredError, COLLECTION_FORMATS} from './baseapi';
```
cjihrig, line 2: It looks like the update is removing the recently added file extensions, which ESM requires. Is https://github.com/kubernetes-client/gen/commit/b461333bb57fa2dc2152f939ed70bac3cef2c1f6 being used (maybe I did something wrong when I did the update last time)?

brendandburns: oh, my bad on the code review of #2062 we need to update the github action. I'll send a PR.

brendandburns: hrm, let me take a look. we may need to update the github action. I sent https://github.com/kubernetes-client/javascript/pull/2106

cjihrig: Maybe this line in the GitHub Action needs to be updated as well - https://github.com/kubernetes-client/javascript/blob/release-1.x/.github/workflows/generate-javascript.yml#L13.

mstruebing: That's just the default which can be overridden on every run, but would make sense to update as well.

brendandburns:
Closing this PR, I will re-open a different one with a fresh generation.

PR #2108, opened 2024-12-13
Regenerate for Kubernetes 1.32

brendandburns:
cc @mstruebing @cjihrig 

Ok, I spot checked a few of these and they have the `.js` extension, so I think that this is right. Please take a careful look since JS modules are not my forte.


cjihrig (approved):
I spot checked 30+ files and nothing jumps out as wrong. The CI seems happy now too. LGTM (and sorry for missing the GHA update before).

mstruebing:
I've rebuild the distribution and ran a couple of the js examples which seemed to work.
I'll approve and we will see if any problem arise 🎉 

/lgtm
/approve

PR #2118, opened 2024-12-19
Merge `master` into `release-0.x`

mstruebing:
When this is merged we can rename `release-1.x` to `master` or even `main` as per newer convention.

I first tried rebasing the branch which worked without problems, which was expected.
But we protected the branch so I could not force push, I could change the rules for a moment - but I think that's not the intention of branch protection rules.


brendandburns:
Let's move to `release-1.x` to `main` as part of this move.

mstruebing:
FYI: I've created a `main` branch out of the latest `release-1.x` branch and changed the settings of the repository to reflect that as the new default branch. 
Additionally I've set the same branch protection rules as we've had for the `master` branch.

PR #2122, opened 2024-12-19
test: support testing transpiled code

cjihrig:
This commit makes it possible to execute the tests after transpilation. This should provide additional confidence that the code shipped to npm works as intended.

I think we'll also want to integrate this into the CI. Should it be for every PR, or the release job only?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,3 +25,4 @@ jobs:
             - run: npm test
             - run: npm run lint
             - run: npm audit --audit-level=critical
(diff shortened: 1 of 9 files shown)


mstruebing (commented):
I think we should do this for every PR instead of only the release job.
That way we know for sure that the current code works and don't find out when we want to release.

cjihrig:
Added to the PR job. It seems to work.

brendandburns:
/lgtm
/approve

Thanks for doing this! We should run it on all PRs, otherwise it will be really painful whenever we try to do a release and things have been broken for a while.

PR #2125, opened 2024-12-24
fix: KubernetesObjectSerializer::deserialize() should support list api

thestupd:
- Added a new KubernetesObjectList class to handle list APIs with appropriate schema attributes (apiVersion, kind, metadata, and items).
- Updated the deserialization logic to instantiate and populate KubernetesObjectList instances when appropriate.
- Added test cases to validate the deserialization of Kubernetes object lists, ensuring correctness and compatibility with unknown object schemas.

### DIFF
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -1,5 +1,6 @@
-import { ObjectSerializer, V1ListMeta } from './api';
+import { ObjectSerializer } from './api';
 import { V1ObjectMeta } from './gen/model/v1ObjectMeta';
+import { V1ListMeta } from './gen/model/v1ListMeta';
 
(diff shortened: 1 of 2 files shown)


empty21:
The custom Serializer update in version `0.22.0` for `KubernetesObject`, replacing `ObjectSerializer`, caused a breaking issue in my application. This is because the `KubernetesObjectList` case was not handled properly.  

Example:  
```typescript
const responseList = await objectApi.list("foo", "bar", "default");
const firstItem = responseList.body.items[0];
// Modify some data
await objectApi.replace(firstItem); // This fails because the items in the list are not deserialized correctly.
```

Issue #2128, opened 2024-12-27
Type definition for `KubernetesObjectApi.read()` is incorrect


jportner:
**Describe the bug**

When reading a [Kubernetes object](https://kubernetes.io/docs/concepts/overview/working-with-objects/), the request must include the `apiVersion`, `kind`, and `metadata` fields for the object. If the object is found, the _response_ will include the `spec` and `status`.

It appears the `KubernetesObjectApi.read()` type definition was accidentally changed in the 1.0.0 release, so the request and response object are the same.

Old type definition in 0.x:

https://github.com/kubernetes-client/javascript/blob/8d4a84c61f916b3809c20734711c3bb085599731/src/object.ts#L19-L27

https://github.com/kubernetes-client/javascript/blob/8d4a84c61f916b3809c20734711c3bb085599731/src/object.ts#L344-L350

The request payload is `KubernetesObjectHeader<T>` and the response is `T`.

New type definition in 1.0.0:

https://github.com/kubernetes-client/javascript/blob/27ba0e1651dbab04a1c85b6d3d9a1c1cfb979b26/src/object.ts#L280-L286

The request payload is `T` and the response is `T`.

**Client Version**

N/A

**Server Version**

N/A

**To Reproduce**

[...]

PR #2129, opened 2024-12-27
Fix type definition for KubernetesObjectApi.read()

jportner:
Resolves https://github.com/kubernetes-client/javascript/issues/2128

This PR restores the old type definition that existed in 0.x.

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -19,6 +19,16 @@ import { PatchStrategy } from './patch.js';
 /** Kubernetes API verbs. */
 type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'list' | 'replace';
 
+type KubernetesObjectHeader<T extends KubernetesObject | KubernetesObject> = Pick<
+    T,
+    'apiVersion' | 'kind'
+> & {
+    metadata: {
+        name: string;
+        namespace?: string;
+    };
+};
+
 interface GroupVersion {
     group: string;
     version: string;
@@ -278,7 +288,7 @@ export class KubernetesObjectApi {
      * @return Promise containing the request response and [[KubernetesObject]].
      */
     public async read<T extends KubernetesObject | KubernetesObject>(
-        spec: T,
+        spec: KubernetesObjectHeader<T>,
         pretty?: string,
         exact?: boolean,
         exportt?: boolean,
--- a/src/object.ts
+++ b/src/object.ts
@@ -19,16 +19,6 @@ import { PatchStrategy } from './patch.js';
 /** Kubernetes API verbs. */
 type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'list' | 'replace';
 
-type KubernetesObjectHeader<T extends KubernetesObject | KubernetesObject> = Pick<
-    T,
-    'apiVersion' | 'kind'
-> & {
-    metadata: {
-        name: string;
-        namespace?: string;
-    };
-};
-
 interface GroupVersion {
     group: string;
     version: string;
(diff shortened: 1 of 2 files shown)


cjihrig:
Changes here LGTM. It looks like after these changes, main contains all of the changes from #854 except some minor differences which I believe are a result of the different underlying clients.

/lgtm

brendandburns:
Hrm, our CI/CD doesn't appear to be triggering, probably because we need to update the github action to include the `main` branch.

cjihrig:
@jportner sorry, but would you mind rebasing this to pick up https://github.com/kubernetes-client/javascript/pull/2132.

cjihrig:
FYI @brendandburns this is what I see. I don't have permission to start the CI:
Screen Shot 2024-12-29 at 1 52 58 AM

@jportner thanks for rebasing. Still LGTM

/lgtm

brendandburns:
@cjihrig ok, I think you need to add yourself here:

https://github.com/kubernetes/org/blob/main/config/kubernetes-client/org.yaml#L141

Also, let's remove @davidgamero from admin since he hasn't been super involved lately (we can leave him as approver/reviewer)

brendandburns:
/lgtm

Is this unit testable? If so, we should add unit tests.

saurabh2590 (approved):

jportner:
> /lgtm
> 
> Is this unit testable? If so, we should add unit tests.

Well I can't think of a way to test the looser type definition directly.
But I did make a second change to the existing unit test for this function in 4c1e6c2dda94d1f48407031546b376383d2ed9db, which demonstrates that the problem is fixed by the first change.

If you temporarily revert the first change in 85007ed780a2acd30b57d6176d66b275d912e4d6, you can see that there's now a type error in the tests. I tried to push a temporary revert to demonstrate that but it looks like the GitHub workflow wouldn't run again until I got another LGTM label 😅 

Anyway, with the first change temporarily reverted (HEAD: 0170e82c0d555eb5d49f9f63f56b5196a11af18b) if you try `npm run build-with-tests` you'll see the type error:

```sh
npm run build-with-tests

> @kubernetes/client-node@1.0.0 build-with-tests
> tsc --project tsconfig-with-tests.json && cp 'src/test/echo space.js' dist/test

src/object_test.ts:1762:66 - error TS2345: Argument of type '{ apiVersion: string; kind: string; metadata: { name: string; namespace: string; }; }' is not assignable to parameter of type 'CustomTestResource'.
  Property 'spec' is missing in type '{ apiVersion: string; kind: string; metadata: { name: string; namespace: string; }; }' but required in type 'CustomTestResource'.

1762             const custom = await client.read<CustomTestResource>({
                                                                      ~
1763                 apiVersion: 'example.com/v1',
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ... 
1768                 },
     ~~~~~~~~~~~~~~~~~~
1769             });
     ~~~~~~~~~~~~~

  src/object_test.ts:1728:17
    1728                 spec: {
                         ~~~~
    'spec' is declared here.


Found 1 error in src/object_test.ts:1762
```

Are you satisfied with that?

mstruebing:
/lgtm

> I tried to push a temporary revert to demonstrate that but it looks like the GitHub workflow wouldn't run again until I got another LGTM label

Someone from within the org needs to approve CI runs for outside collaborators.

@brendandburns is there any specific reason this is active for CI runs which do testing, linting, etc.?
I think there are ways to prevent workflow run changes from within a branch to be used in a workflow run (it will use the workflow definition from `main` if configured correctly).
Would be easier for outsiders to contribute and reduces the feedback cycle loop.

brendandburns:
@mstruebing there are security concerns w/ running random PRs from unknown users automatically. People can do things like craft malicious PRs that mine Bitcoin, etc.

I think we want a maintainer to look at each PR before we approve it for running tests on our compute.

brendandburns:
That test looks good enough to me, I mainly want something that will catch if we regress this change somehow. I think this fits the bill. Thanks!

/lgtm
/approve

Issue #632, opened 2021-04-08
Dependencie net-keepalive does not support win32


eduardocp:
**Describe the bug**
Dependencie `net-keepalive` does not support win32

**Client Version**
0.14.2

**To Reproduce**
In Windows:
yarn add @kubernetes/client-node

**Expected behavior**
Package successfully add

**Current behavior**
Error on fetching packages
![image](https://user-images.githubusercontent.com/1816859/114062596-11f49c00-986e-11eb-98b6-79cb63c62abe.png)

bacongobbler:
Hey thanks for reaching out.

Good catch with this one. I wonder if it is possible to selectively disable the tcp keepalive probe header on non-unix systems, as that's how Go provides this functionality.

From https://golang.org/pkg/net/ (emphasis bold):

> // KeepAlive specifies the interval between keep-alive
    // probes for an active network connection.
    // If zero, keep-alive probes are sent with a default value
    // (currently 15 seconds), **if supported by the protocol and operating
    // system.** Network protocols or operating systems that do
    // not support keep-alives ignore this field.
    // If negative, keep-alive probes are disabled.

bacongobbler:
Alternatively, it does appear the net-keepalive maintainers are welcome to PRs to provide windows support. So another option would be to perform a no-op pointing to Go's pkg/net implementation as reference.

dominykas:
I'm seeing an error on Linux as well:

```
07:21:22 AssertionError [ERR_ASSERTION]: Unable to get socket fd
07:21:22     at _getSocketFD (/mnt/project/node_modules/net-keepalive/lib/index.js:31:3)
07:21:22     at Object.setKeepAliveInterval (/mnt/project/node_modules/net-keepalive/lib/index.js:73:14)
07:21:22     at Request.<anonymous> (/mnt/project/node_modules/@kubernetes/client-node/dist/watch.js:82:27)
```

cebald:
Hello, I am also seeing this issue with net-keepalive not supporting win64. I was just able to install v0.14.0. Is there a later version that I can use in the meantime where net-keepalive should still work? Thanks for your help.

eduardocp:
> Hello, I am also seeing this issue with net-keepalive not supporting win64. I was just able to install v0.14.0. Is there a later version that I can use in the meantime where net-keepalive should still work? Thanks for your help.
> 
> EDIT: v0.14.1 installed fine as well. v0.14.2 seems to be the first problem 

[...]

Issue #559, opened 2020-11-24
Unable to detect if a watch is active


maxweisel:
This is the same bug that occurred in the Go client library here: https://github.com/kubernetes/kubernetes/issues/65012

Essentially if you create a watch stream and the connection is severed, the stream is unaware and believes it is open. It will stay open forever and just stop reporting events instead of trying to reconnect. This occurs anytime our k8s control plane is updated. Ideally it would detect the broken connection and reconnect to another replica or at least mark the connection closed.

It seems the solution is to occasionally http2 ping the connection to ensure that it's still open, and close if the ping is not received.

I've dug into the library a little and it seems watch.js uses [request](https://www.npmjs.com/package/request) which is now deprecated. The watch class does allow specifying my own request implementation in the constructor, which I imagine I could use to add the ping feature, but I imagine this is a bug that will impact everyone who uses the watch event, so I figured I'd file an issue.

Happy to contribute a patch that includes this functionality, I'd love some guidance on what the preferred direction should be. Is there a suitable replacement for request that I could implement while I'm working on this patch?

Max

brendandburns:
Thanks, please see the discussion of alternatives to `request` here:

https://github.com/kubernetes-client/javascript/issues/414

The main problem with switching libraries is that all of this code is generated by the OpenAPI Generator project:
https://github.com/OpenAPITools/openapi-generator

We could change generators, but that would also result in significant breaking API changes in the library. I think that the best option is to fork the request generator in OpenAPI Generator (`typescript-node`) and switch it to a different HTTP library while maintaining the same generated API signatures.

This is a pretty significant undertaking though, and for now no one has taken it on. If there is interest, I'm happy to advise on what needs to be done. I don't have the cycles for it personally, but I can act as a guide.

Regarding hanging watches, I think the best answer is to set the socket timeout pretty low (e.g. 1 minute) which _should_ cause the TCP connection to reset anyway.

flyboarder:
@brendandburns I would be willing to attempt migrating to a new generator.

entropitor:
@maxweisel Thank you for discovering this! We have been plagued by this and I wasn't sure what the problem was

@jhagestedt It seems https://github.com/kubernetes-client/javascript/commit/1b313ce09ad2f50ba1292e07e87386b2fa03e3df might be the culprit. Why is the timeout set to 0? 

(Although I'm using 0.11.1 and I also seem to have the problem)

entropitor:
[...]

PR #2131, opened 2024-12-28
Reintroduce timeout and keep-alive for watch requests to match client-go

### LINKED ISSUE #2127: Watch keepalive not working anymore with 1.0.0
**Describe the bug**
With the version 1.0.0, when a `watch` request is made and the connection drop, no more events are reported, but there is also no error that indicate that the connection was dropped. (I waited 10 minutes to be sure)

With version 0.22.2, under ~33s, I got the error `read ETIMEDOUT`, allowing me to retry the connection and restart the watch request.

This issue was already raised on #559 and fixed with PR #630. This PR add a config to configure keepalive packets in `watch.ts`, but I don't see this config on the version 1.0.0.

**Client Version**
`1.0.0` (but working on *0.22.2*)

**Server Version**
`1.30.3`

**To Reproduce**
In my case, I start the `watch-example.ts` and then cut the VPN connection to my cluster. As it's related to network perturbation, it's not easy to suggest an easy reproducible way to do it. I guess you can also use your firewall to drop packets in the middle of a watch request.

**Expected behavior**
I expect an error to be thrown when the Kubernetes api becomes unavailable while a watch request is running like it was the case for previous versions.

** Example Code**
I used the `watch-example.ts`

 - OS: Linux
 - NodeJS Version 20.16.0

Alabate:
This will send sends keep-alive probes to the server every 30 seconds. These features were present prior to the 1.0 but were inadvertently removed.

Fixes #2127

Previous relevant issues:
- Initial issue: #559
  - PR: #630
- Improvement: #632
  - PR: #635

### Implementation details
Setting `requestInit.timeout = 30000` is equivalent to `socket.setTimeout(30000)` as you can [see in the sources](https://github.com/nodejs/node/blob/main/lib/_http_agent.js#L527).

One of the solution suggested in #2127 was to use the keepAlive of the `http.Agent`. But [as documented here](https://nodejs.org/api/http.html#new-agentoptions), that's just a boolean that instruct the agent that we want sockets to be re-used. We want to send packets at a fixed rate to know when a connection is broken. This can be done with the [`socket.setKeepAlive()` method](https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay), but to use it, we need to access the socket object.

I managed to access the socket and call `setKeepAlive()`, but only after the `await fetch()`, when the response is already arriving. That's the only way I found to access the socket. The Agent create the socket, but [there is no event](https://nodejs.org/api/http.html#class-httpagent) to know when a socket created, you can just access them by list like I did. There is a[ `socket` event on the request object](https://nodejs.org/api/http.html#event-socket), but `node-fetch` abstract it away, and I couldn't find a clean way to access it ([See discussion](https://github.com/node-fetch/node-fetch/discussions/1720) that request this feature).

The only way I've found to get the socket before the request was by monkey-patching the `agent.createConnection()`, but I think this solution is worse.
```javascript
const _createConnection = agent.createConnection;
agent.createConnection = function (...args) {
  const socket = _createConnection.apply(agent, args);
  socket.setTimeout(3000);
  socket.setKeepAlive(true, 3000);
  return socket;
}
```

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -40,6 +40,7 @@ export class Watch {
         const controller = new AbortController();
         requestInit.signal = controller.signal as AbortSignal;
         requestInit.method = 'GET';
+        requestInit.timeout = 30000;
 
         let doneCalled: boolean = false;
         const doneCallOnce = (err: any) => {
@@ -52,6 +53,15 @@ export class Watch {
 
         try {
             const response = await fetch(watchURL, requestInit);
+
+            // Enable socket keep-alive to prevent connection stalls
+            if (requestInit.agent && typeof requestInit.agent === 'object') {
+                for (const socket of Object.values(requestInit.agent.sockets).flat()) {
+                    if (socket) {
+                        socket.setKeepAlive(true, 30000);
+                    }
+                }
+            }
 
             if (response.status === 200) {
                 response.body.on('error', doneCallOnce);


cjihrig:
> One of the solution suggested in https://github.com/kubernetes-client/javascript/issues/2127 was to use the keepAlive of the http.Agent. But [as documented here](https://nodejs.org/api/http.html#new-agentoptions), that's just a boolean that instruct the agent that we want sockets to be re-used. We want to send packets at a fixed rate to know when a connection is broken. This can be done with the [socket.setKeepAlive() method](https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay), but to use it, we need to access the socket object.

There is also a `keepAliveMsecs` option in the same docs you linked to. Would using that in conjunction with the `keepAlive` boolean do what you need and avoid looping over all of the agent's sockets? I believe that is how `socket.setKeepAlive()` is implemented.

Alabate:
I agree that it should, but when I tried it didn't work, so I though I did not understand the documentation. But in the end, I think it's more because of this issue: https://github.com/nodejs/node/issues/47137

To confirm it, I ran the following snippet that simply create an agent and a request and try to read the `setKeepAlive` status:

```javascript
import * as protocol from 'node:http';

const agent = new protocol.Agent({
    keepAlive: true,
    keepAliveMsecs: 3000,
});
console.log('========= Protocol ', agent.protocol, '==========');

const req = protocol.request(`${agent.protocol}//google.com`, { agent: agent }, (res) => {
    logKeepaliveSocketStatus(res.socket);

    req.socket.setKeepAlive(true, 4000);
    logKeepaliveSocketStatus(res.socket);
});
req.end();

function logKeepaliveSocketStatus(socket) {
    const kSetKeepAlive = Object.getOwnPropertySymbols(socket).find(
        (sym) => sym.description === 'kSetKeepAlive',
    );
    const kSetKeepAliveInitialDelay = Object.getOwnPropertySymbols(socket).find(
        (sym) => sym.description === 'kSetKeepAliveInitialDelay',
    );
    console.log('----- Socket status -----');
    console.log('kSetKeepAlive:', socket[kSetKeepAlive]);
    console.log('kSetKeepAliveInitialDelay:', socket[kSetKeepAliveInitialDelay]);
}
```

When I execute this script with the `node:http` module, I get the following output:
```
========= Protocol  http: ==========
----- Socket status -----
kSetKeepAlive: true
kSetKeepAliveInitialDelay: 3
----- Socket status -----
kSetKeepAlive: true
kSetKeepAliveInitialDelay: 4
```

However, if I replace `node:http` with `node:https`, I get:

```
========= Protocol  https: ==========
----- Socket status -----
kSetKeepAlive: false
kSetKeepAliveInitialDelay: 0
----- Socket status -----
kSetKeepAlive: true
kSetKeepAliveInitialDelay: 4
```

And in our case we use the https agent

brendandburns:
Please add unit tests also.

anandfresh (approved):

cjihrig:
As much as I don't like the workaround here, I haven't been able to find a better way with the current project setup. It's also surprising that no one has tried to fix https://github.com/nodejs/node/issues/47137. LGTM once there are tests.

anandfresh (approved):

rossanthony:
@Alabate I need this fix for something I'm working on, if you're too busy would you mind if I try to complete the unit tests for this fix so we can hopefully get this merged?

rossanthony:
@brendandburns I've opened another PR with @Alabate's fix + unit test coverage: https://github.com/kubernetes-client/javascript/pull/2367

cjihrig:
Superseded by #2367. Thanks for getting this started though.

PR #2144, opened 2025-01-08
websocket: do not close when protocol supports it

cjihrig:
This commit updates the WebSocket handler to not call `ws.close()` when stdin is closed but the ws protocol already supports closing (v5 protocol for example).


davidgamero (approved):
/approve
/lgtm

Issue #2116, opened 2024-12-18
Establish a plan for linting moving forward


cjihrig:
**Describe the bug**
This package uses TSLint, which is deprecated. The suggested upgrade path seems to have some issues and complexity. Since there are a number of options for linting, I figured it was worth opening an issue to decide on a direction first.

A few possibilities:
- Move to ESLint directly. This would have been my choice in previous versions, but the latest ESLint seems more complex and potentially involves adding a number of support modules.
- Move to an alternative tool. Examples would be things like [neostandard](https://www.npmjs.com/package/neostandard), [typescript-eslint](https://www.npmjs.com/package/typescript-eslint), [biome](https://github.com/biomejs/biome).

I think no matter which option we pick, we might end up tweaking the lint rules over time if we want to match the TSLint configuration as closely as possible (not sure if that is necessarily a goal though).

**Client Version**
latest

**Server Version**
not relevant

**To Reproduce**
Run `npm ci` on this repo

**Expected behavior**
`npm ci` finishes with no warnings or errors related to this package

**Example Code**
not relevant

**Environment (please complete the following information):**
local development

**Additional context**
Quoting my comment from #2050 regarding migrating away from TSLint:
> FWIW, I briefly started looking into this. It's not as simple as I had hoped it would be. TSLint provides an automatic migration to ESLint that preserves many of the current linting settings. However, it seems to migrate to a now outdated version of ESLint, so the generated config has some problems. In turn, ESLint also provides an automatic update migration. Unfortunately, ESLint seems to have gotten a lot more complex in its latest version and requires installing a bunch of additional modules.

brendandburns:
I'm ok with adjusting our style to match a different linter. I would want to see the differences in a few example files first and then we could converge on a different solution.

Is there one linter (other than eslint) that appears to be more commonly used than the others?

cjihrig:
> Is there one linter (other than eslint) that appears to be more commonly used than the others?

I think most of the top options are layers built on top of ESLint. Based on download numbers and what I've heard of people using, I would probably say typescript-eslint.

There is also Biome, which is not based on ESLint. It is a full toolchain, including a formatter that advertises 97% compatibility with Prettier, and a linter that supports more than 270 ESLint rules. I have never used it, but it could mean fewer dependencies, and likely better performance since it is Rust based.

cjihrig:
Fixed by https://github.com/kubernetes-client/javascript/pull/2162

PR #2162, opened 2025-01-14
deps: replace tslint with typescript-eslint

cjihrig:
tslint has been deprecated. This commit replaces it with typescript-eslint.

This is a proposed fix for https://github.com/kubernetes-client/javascript/issues/2116


src/util.ts
```diff
@@ -8,7 +8,7 @@ export async function podsForNode(api: CoreV1Api, nodeName: string): Promise<V1P
 
 export function findSuffix(quantity: string): string {
     let ix = quantity.length - 1;
-    while (ix >= 0 && !/[\.0-9]/.test(quantity.charAt(ix))) {
+    while (ix >= 0 && !/[.0-9]/.test(quantity.charAt(ix))) {
```
brendandburns, line 11: I think this is a breaking change we want the escaped `\.` to just match <dot> vs <dot> in regexp which matches all characters?

brendandburns:
Looks good. One comment on a regexp, but maybe I'm missing something?

Thanks for doing this!

src/util.ts
cjihrig: That change was related to this error:

>   11:26  error  Unnecessary escape character: \.  no-useless-escape

Here are the docs for [no-useless-escape](https://eslint.org/docs/latest/rules/no-useless-escape). I can put it back and disable the lint rule for that line if you'd prefer.

brendandburns: ok, thanks for checking.

PR #2166, opened 2025-01-16
deps: remove whatwg-url override

cjihrig:
This commit removes the final override from the package.json file. The whatwg-url override was added to silence Node core deprecation warning DEP0040. While this silences the warning in the CI, it does not help users of this module.

I have also opened https://github.com/nodejs/node/pull/56632. If landed, that will make the warning go away for users of this library.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3081,6 +3081,28 @@
                 }
             }
         },
+        "node_modules/node-fetch/node_modules/tr46": {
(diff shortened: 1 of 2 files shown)


cjihrig:
Node 23.7.0 was released today. The punycode warning has been silenced 🎉  on v23. Now just need to wait for the fix to be backported to the v22 release line.

Issue #2168, opened 2025-01-20
kc.loadFromCluster() doesn't setup `getCurrentCluster()` correctly.


koooge:
Hi there,
[patch-example.js](https://github.com/kubernetes-client/javascript/blob/1.0.0/examples/patch-example.js) doesn't work with `kc.loadFromCluster()` instead of `kc.loadFromDefault()`. The app is running in a pod, and it causes `Error: Cluster is undefined` https://github.com/kubernetes-client/javascript/issues/2117#issuecomment-2602394199  

I'm not sure if it's a bug or if there is another usage to fix it. At least, loadFromCluster works in v0 for me.

brendandburns:
it looks like `currentCluster` is undefined in the `loadFromCluster` function.  You could try substituting `getClusters()[0]` instead of `getCurrentCluster()` and see if that works.

But we should probably check and fix to make it so that `getCurrentCluster()` works.

cjihrig:
I think this is a minimal reproduction:

```js
import * as k8s from '@kubernetes/client-node';
const kc = new k8s.KubeConfig();
kc.loadFromCluster();
const currentContext = kc.getCurrentContext();
const currentCluster = kc.getCluster(currentContext);
console.log(currentCluster); // Logs null
```

When `kc.loadFromCluster()` is called, `kc.clusters` is set to:

```js
{
    name: 'inCluster',
    caFile: `${pathPrefix}${SERVICEACCOUNT_CA_PATH}`,
    server: `${scheme}://${serverHost}:${port}`,
    skipTLSVerify: false,
},
```

The call to `kc.getCurrentContext()` always returns the value `'inClusterContext'`.

So, you're always executing `kc.getCluster('inClusterContext')`. Internally, this searches the list of clusters for one whose `name` is `'inClusterContext'`. A cluster with the same name as the context does not exist.

cjihrig:
> But we should probably check and fix to make it so that getCurrentCluster() works.

It looks like that does work.

cjihrig:
@koooge can you confirm if changing the code from `kc.getCluster(currentContext);` to `kc.getCurrentCluster();` gets you past that error. It seems to me like the example should be using that code.

brendandburns:
@cjihrig thanks for looking closer at this. You are correct, there is a bug in the code from @koooge

It should be:

`kc.getCluster(kc.getContextObject(kc.getCurrentContext()).cluster)`

but really `kc.getCurrentCluster()` is easier.

[...]

PR #2191, opened 2025-01-31
Improve loadFromCluster testing

brendandburns:


cjihrig:
This LGTM. I believe the issue with #2168 is that the problematic code came from our own examples, but more tests are always good.

/lgtm
/approve

PR #2195, opened 2025-02-01
src: update for node-fetch v3 compatibility

cjihrig:
While the code generator still needs to be updated to use node-fetch v3, these changes make the library compatible with it. Since node-fetch v3 is more up to date, I think it makes sense to try to be compatible with it.

Tested with v3 by applying these changes on top of #2146. This seems to work with node-fetch v2 and v3.


brendandburns:
/lgtm
/approve

Thanks for figuring this out!

PR #2216, opened 2025-02-10
docs: correct directory and set entryPointStrategy

### LINKED ISSUE #2209: Cant generate docs via `npm run docs` - "The glob ./src/gen/api did not match any files"
Since the online documentation is out out of date (https://github.com/kubernetes-client/javascript/issues/2190), i want to view it locally.
Cloned the repo cd into it, ran `npm install`, `npm run generate` and `npm run docs`:
```
> @kubernetes/client-node@1.0.0 docs
> typedoc src/gen/api

[warning] The glob ./src/gen/api did not match any files
[error] Unable to find any entry points. See previous warnings
[error] Found 1 errors and 1 warnings
```

mstruebing:
The directory where to generate the docs from was wrong, I'm not sure
when this changed but probably during the move from release-0.x to
release-1.x.

Also the `entryPointStrategy` needed to be set so that we are able to
include these files to generate the docs even if our ts-config doesn't
include them.
This was the default behavior a couple of releases of typedoc ago but
was changed.

Closes #2209

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
         "test-transpiled": "mocha --no-config dist",
         "prepare": "npm run build && husky",
         "prepack": "npm run build",
-        "docs": "typedoc src/gen/api"
+        "docs": "typedoc src/gen/apis"
     },
     "c8": {
         "include": [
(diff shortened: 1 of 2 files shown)


cjihrig:
FYI - the bot does not like the commit message.

mstruebing:
@cjihrig thanks for the notification.
I've changed the commit message to remove the `Closes #2209` part.

PR #2217, opened 2025-02-10
Fix list watch error handling

CGNonofr:
when listFn returns an error, the ListWatch is broken and no ERROR event is sent

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -142,28 +142,33 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
             return;
         }
         this.callbackCache[CONNECT].forEach((elt: ErrorCallback) => elt(undefined));
-        if (!this.resourceVersion) {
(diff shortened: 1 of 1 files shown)


mstruebing:
@CGNonofr thank you very much for your contribution 🚀 

It seems like the email address you did the commit with is not linked to your GitHub account. This is a requirement for EasyCLA to pass which is required to merge this PR.

Also it would be very nice if you could add a test to make sure this we keep this behavior in the future (and that it works as expected of course).

CGNonofr:
As per my last PR, I don't think I'll want to sign the CLA

I authorize you to close that PR and include the fix in the way you want

I don't have the time right now to investigate how to implement such test, maybe in the next weeks

brendandburns:
Created this issue to track:
https://github.com/kubernetes-client/javascript/issues/2222

Closing this PR.

PR #2224, opened 2025-02-12
Fix a bug where an error on request wouldn't propogate.

### LINKED ISSUE #2222: List/Watch fails without error if the list function failse
**Describe the bug**
When listFn returns an error, the ListWatch is broken and no ERROR event is sent

**Client Version**
All

**Server Version**
n/a

**To Reproduce**
Fail the initial list function in a list/watch call.

**Expected behavior**
An error should be emitted to the done handler

See #2217

brendandburns:
Fixes #2222

### DIFF
--- a/src/cache_test.ts
+++ b/src/cache_test.ts
@@ -1,4 +1,4 @@
-import { deepStrictEqual, notStrictEqual, strictEqual, throws } from 'node:assert';
+import { deepStrictEqual, fail, equal, notStrictEqual, strictEqual, throws } from 'node:assert';
 import mock from 'ts-mockito';
 
 import { V1Namespace, V1NamespaceList, V1ObjectMeta, V1Pod, V1PodList, V1ListMeta } from './api.js';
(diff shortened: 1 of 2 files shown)


src/cache_test.ts
```diff
@@ -1460,4 +1460,29 @@ describe('delete items', () => {
...
+        };
+        const lw = new ListWatch('/some/path', fake, listFn);
+        let gotErr: Error | null = null;
+        const errCalled = new Promise<void>((resolve, reject) => {
+            lw.on('error', (err) => {
+                gotErr = err;
+                resolve();
+            });
+        });
+        try {
```
cjihrig, line 1480: I don't think we need the `try...catch` since the unexpected error will fail the test.

src/cache_test.ts
```diff
@@ -1460,4 +1460,29 @@ describe('delete items', () => {
...
+            await lw.start();
+            await errCalled;
+            equal(gotErr, requestErr);
```
cjihrig, line 1483: Since we want a `===` check here:
```suggestion
            strictEqual(gotErr, requestErr);
```

cjihrig (commented):
Two small nits on the test.

src/cache_test.ts
brendandburns: done (and removed the import at the top)

brendandburns:

Thanks!

cjihrig (approved):
/lgtm
/approve

PR #2230, opened 2025-02-14
Add a devcontainer, since the default for codespaces is node 16 :0

brendandburns:


.github/dependabot.yml
```diff
@@ -5,9 +5,10 @@ updates:
       schedule:
           interval: 'daily'
           time: '02:00'
+      target-branch: 'main'
     - package-ecosystem: 'npm'
       directory: '/'
-      target-branch: 'master'
+      target-branch: 'main'
```
cjihrig, line 11: Will this stop updating the v0.x branch? If it does intentionally, maybe we should just end-of-life the v0.x branch officially? I think it also means we would have two checks for the main branch.

brendandburns: I'm ok with keeping this as long as it was on purpose. I wasn't sure if this was intentional or a left-over.

cjihrig: AFAIK it's necessary for having dependabot continue updating the master branch, but if anyone knows better, I'm happy to be wrong.

brendandburns:
switched `main` back to `master`

cjihrig (approved):
/lgtm
/approve

.github/dependabot.yml
mstruebing: Based on what I can see from the dependabot PRs is that it is bumping on `master` rather than `release-1.x`.
Should that be changed to `release-1.x` or do I miss something?

Issue #2014, opened 2024-11-04
Upgrade to 0.22.2 fails due to use of new major version of client-node.


TonyJDavies:
**Describe the bug**
When upgraded to latest version `0.22.2` of this library getting the following error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/app/node_modules/openid-client/build/index.js from /home/app/node_modules/@kubernetes/client-node/dist/oidc_auth.js not supported.
Instead change the require of index.js in /home/app/node_modules/@kubernetes/client-node/dist/oidc_auth.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/home/app/node_modules/@kubernetes/client-node/dist/oidc_auth.js:5:35)
    at /home/app/node_modules/@kubernetes/client-node/dist/oidc_auth_delayed.js:44:70
    at async DelayedOpenIDConnectAuth.applyAuthentication (/home/app/node_modules/@kubernetes/client-node/dist/oidc_auth_delayed.js:44:22)
    at async KubeConfig.apply(--- cut ---)
    at async KubeConfig.applyOptions (/home/app/node_modules/@kubernetes/client-node/dist/config.js:388:9)
    at async KubeConfig.applyToRequest (/home/app/node_modules/@kubernetes/client-node/dist/config.js:110:9)
```

This appears to be due to using the new version of `client-node`.
** Client Version **
`0.22.2`

** Server Version **
e.g. `1.19.1`

**To Reproduce**

**Expected behavior**

** Example Code**
Code snippet for what you are doing

**Environment (please complete the following information):**
 - OS: [e.g. Windows, Linux]
 - NodeJS Version [eg. 10]
 - Cloud runtime [e.g. Azure Functions, Lambda]

**Additional context**

brendandburns:
I'll try to repro this. I would have thought that the unit tests for this code would have failed though.

Do you have a snippet of code that reproduces this?

Also, what version of Node are you using?

cjihrig:
It looks like #1969 bumped the openid-client dependency from 5.7.0 to 6.1.3. Looking at the [v6.0.0 release notes](https://github.com/panva/openid-client/releases/tag/v6.0.0), that dependency switched to being an ESM module. Since the TypeScript code is transpiled to use `require()`, I'm guessing that explains the error.

cjihrig:
> I would have thought that the unit tests for this code would have failed though.

[...]

PR #2240, opened 2025-02-18
Offer both ESM & CommonJS versions of the Javascript client

lucavb:
## Summary

This pull request introduces key improvements to the @kubernetes/client-node package, focusing on enhancing module compatibility, maintainability, and test code clarity. The changes have been distributed across three commits to facilitate easier review.
## Changes

* Build Enhancement:
        Added support for the CommonJS module format in the bundled package. This change includes the creation of separate tsconfig.cjs.json and tsconfig.esm.json files to manage different module outputs.

*   Chore:
        Sorted keys in tsconfig.json for improved readability and maintainability.

*    Test Improvement:
        Removed a needless return statement from src/integration_test.ts as the return type of deepEqual is void. This change clarifies the test logic and aligns with best practices for writing tests.

## Reason for Change

The primary motivation for adding CommonJS support is to optimize the performance of applications that still rely on jest. Without this change, these applications would need to transpile the entire package, leading to slower build times and reduced efficiency. Additionally, cleaning up the test code enhances code readability and maintainability.
Impact

These changes should improve the flexibility and performance of the package distribution, making it easier to integrate with a wider range of environments and frameworks, while also ensuring that the test code is clean and easy to understand.

## Review Notes

The changes have been distributed across three separate commits, making it easier for reviewers to focus on specific aspects of the update and provide targeted feedback.

### DIFF
--- a/src/integration_test.ts
+++ b/src/integration_test.ts
@@ -40,7 +40,7 @@ describe('FullRequest', () => {
 
             const list = await k8sApi.listNamespacedPod({ namespace: 'default' });
 
-            return deepEqual(list, result);
+            deepEqual(list, result);
         });
     });
 });
--- a/package.json
+++ b/package.json
@@ -8,17 +8,29 @@
         "url": "https://github.com/kubernetes-client/javascript.git"
     },
     "files": [
-        "dist/**/*.ts",
-        "dist/**/*.js",
-        "dist/**/*.map"
+        "dist/{mjs,cjs}/*.js",
+        "dist/{mjs,cjs}/*.map",
+        "dist/{mjs,cjs}/*.ts",
+        "dist/{mjs,cjs}/gen/**/*.js",
+        "dist/{mjs,cjs}/gen/**/*.map",
+        "dist/{mjs,cjs}/gen/**/*.ts",
+        "dist/{mjs,cjs}/gen/*.js",
+        "dist/{mjs,cjs}/gen/*.map",
+        "dist/{mjs,cjs}/gen/*.ts"
     ],
-    "main": "dist/index.js",
-    "types": "dist/index.d.ts",
+    "main": "dist/mjs/index.js",
+    "types": "dist/mjs/index.d.ts",
+    "exports": {
+        ".": {
+            "require": "./dist/cjs/index.js",
+            "import": "./dist/esm/index.js"
+        }
+    },
     "scripts": {
         "format": "prettier --log-level error --write \"./src/**/*.ts\"",
         "lint": "eslint \".\" && prettier --check \"./src/**/*.ts\"",
         "clean": "rm -Rf node_modules/ dist/",
-        "build": "tsc",
+        "build": "tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json",
         "build-with-tests": "tsc --project tsconfig-with-tests.json && cp 'src/test/echo space.js' dist/test",
         "generate": "./generate-client.sh",
         "watch": "tsc --watch",
--- a/tsconfig.cjs.json
+++ b/tsconfig.cjs.json
@@ -1,0 +1,10 @@
+{
+    "compilerOptions": {
+        "module": "CommonJS",
+        "moduleResolution": "node",
+        "outDir": "dist/cjs"
+    },
+    "exclude": ["node_modules", "src/*_test.ts", "src/test", "dist"],
+    "extends": "./tsconfig.json",
+    "include": ["*.ts", "src/**/*"]
+}
--- a/tsconfig.esm.json
+++ b/tsconfig.esm.json
@@ -1,0 +1,8 @@
+{
+    "compilerOptions": {
+        "outDir": "dist/mjs"
+    },
+    "exclude": ["node_modules", "src/*_test.ts", "src/test", "dist"],
+    "extends": "./tsconfig.json",
+    "include": ["*.ts", "src/**/*"]
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,5 +20,16 @@
         // "declarationMap": true
     },
     "exclude": ["node_modules", "src/*_test.ts", "src/test", "dist"],
-    "include": ["*.ts", "src/**/*"]
+    "include": ["*.ts", "src/**/*"],
+    "reference": [
+        {
+            "path": "tsconfig.cjs.json"
+        },
+        {
+            "path": "tsconfig.mjs.json"
+        },
+        {
+            "path": "tsconfig-with-tests.json"
+        }
+    ]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,21 +1,21 @@
 {
     "compilerOptions": {
-        "module": "nodenext",
-        "noImplicitAny": false,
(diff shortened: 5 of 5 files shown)


lucavb:
If someone could share with me which forbidden keyword I've used in the main commit, then I'll gladly update it :)

cjihrig:
I don't think this will address https://github.com/kubernetes-client/javascript/issues/2014. That issue was originally opened against the v0.x release line, but applies to the v1.x release line as well. The problem is that this library depends on `openid-client`, which has moved to ESM only. Won't that still be a problem for you?

cjihrig:
Just my personal opinion - this module should not be published as CJS if it is known to be an incomplete solution because we will still get bug reports about it.

By the way, have you looked into `require(esm)` or dynamic import? `require(esm)` is on Node 22 and 23, and has been merged in Node 20 (so it should be in the next v20 release).

src/integration_test.ts
```diff
@@ -40,7 +40,7 @@ describe('FullRequest', () => {
 
             const list = await k8sApi.listNamespacedPod({ namespace: 'default' });
 
-            return deepEqual(list, result);
+            deepEqual(list, result);
```
rossanthony, line 43: Was the return here unnecessary? I'm guessing that deepEqual will throw and fail the test if these two object don't match?

lucavb: Yes, that `return` was superfluous. The return type is simply `void` and so it kinda makes no difference.

lucavb:
> Just my personal opinion - this module should not be published as CJS if it is known to be an incomplete solution because we will still get bug reports about it.

I get that and I don't specifically like it either. I wish CJS would just go away and we'd all be happy. But I also feel like that is something we'll discuss in maybe 10 years time and unfortunately not today. And vitest similarly was also not a drop in replacement for us :/

With regards to the bug reports about it. I can kind of see that but then again you've also already received one issue about this package being ESM only. 

> By the way, have you looked into `require(esm)` or dynamic import? `require(esm)` is on Node 22 and 23, and has been merged in Node 20 (so it should be in the next v20 release).

I've just tested it (`--experimental-require-module`) and jest still complains. Which to be fair makes sense to me. They hook into the loading process and try to preprocess the file. So the problem is not that node in general can not deal with it but rather that Jest can't. Again really not happy about it.

cjihrig:
> So the problem is not that node in general can not deal with it but rather that Jest can't.

Jests problems are very long standing and well known at this point. In my opinion, time would be better spent migrating away from Jest rather than shipping partially working CJS implementations to workaround Jest.

> But I also feel like that is something we'll discuss in maybe 10 years time and unfortunately not today.

I think we are significantly closer than that, you're just using a test runner that actively fights against it 😄. `require(esm)` will be in all supported Node.js release lines in two months and dynamic import is already available.

src/integration_test.ts
cjihrig: Can you move this to a separate PR.

lucavb:
Not exactly sure what else I can in turn offer here in terms of opinion. In my opinion offering CommonJS side by side does not really offer any downsides and would make consumers happy. Seemingly there is also another project that is currently not able to update to version 1.0.0 [here](https://github.com/vscode-kubernetes-tools/vscode-kubernetes-tools/pull/1456). 

So I guess my question is whether this addition has any chance of getting merged or not.

mstruebing:
> this module should not be published as CJS if it is known to be an incomplete solution

I think that is the main point here and I agree with @cjihrig. 
Would still be interesting what @brendanburns opinion about it is.

lucavb:
I wanted to add another data point from my side. In the latest version of Node 22, which we are currently using, `--experimental-require-module` is enabled by default, yet it still doesn't work as expected. Therefore, I'm not very optimistic that Node will resolve this issue for us anytime soon. Additionally, as mentioned earlier, migrating to Vitest is not a straightforward process for us.

I completely understand your decision to not continue supporting CommonJS and recognize that Jest is the main issue here. Given these challenges, if this package remains ESM-only, we might not be able to migrate to version 1.0.0 in the near future. The transpiling to CJS is proving to be too slow for our needs.

cjihrig:
> The transpiling to CJS is proving to be too slow for our needs.

At the risk of getting off topic here - why not vendor/fork and perform the transpilation once? This module doesn't get released all that frequently.

brendandburns:
I agree with @cjihrig that we don't want to publish a commonjs module here. the older branches should provide sufficient utility while jest or node figures out how they want to fix this.

Apologies that this isn't a great solution for your case, but I don't want to confuse the support matrix by adding another option.

robross0606:
I am _so_ confused by this project.  If this library only supports ESM, then why don't any of the release notes say this?  That's a massive breaking change and should be noted front and center.  [The README](https://github.com/kubernetes-client/javascript/blob/main/README.md) still shows every example with `require()` which is CommonJS and does not mention anywhere that this module is now ESM-only.  Since this is typescript-based, and it is fairly common to transpile to both CommonJS and ESM, I assumed this library handled both.  And since it wasn't mentioned anywhere prominent that the library is ESM-only, I just wasted _hours_ working on an upgrade/migration only to hit an issue when I finally tried unit tests.

Could we at least update the README here to include this information?  I'd think the examples in the README should be updated too, but at least a bold note about ESM vs CommonJS would be nice.

cjihrig:
You're right, the README should be updated - this was missed while updating all of the examples. I have gone ahead and added bold text to the v1.0.0 release notes.

lucavb:
Thank you for considering the proposal. While I'm a bit disappointed by the decision not to support CommonJS, I respect your authority and the reasoning behind it. I appreciate the effort you put into maintaining and improving the project. I'll go ahead and close the PR now.

brendandburns:
@robross0606 I totally understand the challenges of the transition. We're facing similar things in the Java client with the phasing out of java8.

It's also unfortunate that the request project decided to shut itself down despite lots of libraries (like the older version of this one) depending on it.

The challenge that we are faced with is that as various things we depend on switch to ESM, the set of functionality that we can import becomes smaller and smaller. I realize that by making the transition now we're effectively forcing the decision upstream. When we switched away from request, it made sense to try to combine the API change with the module change so that people only had a hard migration once.

I understand that for some users this is pretty painful.

That said, the beauty of open source is that you don't have to take a different dependency, as @cjihrig pointed out, you can fork this repository, apply this PR, and even push it to NPM if you want to, I expect that it shouldn't be too much work to do this every release and rebase as we add bug fixes etc.

It seems to me at least that ESM is the way of the future, so everyone is headed there eventually, so I don't see any long-term viable path to maintaining both libraries.

Thanks

robross0606:
At the very least, I think it is unreasonable to not backport known audit vulnerability patches when you know you've put users in this situation.  It is entirely possible to eradicate the use of `request` without moving to ESM.  You did not help anyone by making that decision for them.  You made things harder.

While the decision you've made is "the future", it also leaves people in the present with an impossible decision.  Entire testing frameworks like `jest` where we've invested hundreds of tests and mocks does not work well with ESM.  In fact, support is still labelled as "experimental".  Therefore, you've forcing people into the future before it is stable or ready.

Issue #2249, opened 2025-02-22
Missing Source Map Files in `@kubernetes/client-node@1.0.0` Cause ~100 Vite Warnings


MichaelJCole:
#### Description
The `@kubernetes/client-node` package (v1.0.0) includes source map comments (e.g., `//# sourceMappingURL=StorageV1alpha1Api.js.map`) in its generated JS files under `dist/gen/apis/`, but the `.map` files are not included in the npm package. This causes repeated "Failed to load source map" warnings in Vite (used by Nuxt 3) during development, cluttering logs and confusing developers. The warnings are cosmetic but degrade the dev experience due to a packaging issue.

#### Steps to Reproduce
1. Create a Nuxt 3 project:
   ```
   npx nuxi init k8s-test
   cd k8s-test
   npm install
   npm install @kubernetes/client-node@1.0.0
   ```
2. Add to `server/utils/ingress.ts`:
   ```typescript
   import * as k8s from '@kubernetes/client-node';
   export async function testK8s() {
     const kc = new k8s.KubeConfig();
     kc.loadFromCluster();
     const k8sApi = kc.makeApiClient(k8s.NetworkingV1Api);
     console.log('K8s API initialized');
   }
   ```
3. Add to `plugins/test.server.ts`:
   ```typescript
   import { defineNuxtPlugin } from '#app';
   import { testK8s } from '~/server/utils/ingress';
   export default defineNuxtPlugin(() => { testK8s(); });
   ```
4. Run `npm run dev` and check logs.

#### Expected Behavior
No source map warnings, either by including `.map` files or removing the comments.

#### Actual Behavior
~100 Warnings like:
```
[7:46:46 PM]  WARN  Failed to load source map for /node_modules/@kubernetes/client-node/dist/gen/apis/StorageV1alpha1Api.js.
Error: ENOENT: no such file or directory, open '/node_modules/@kubernetes/client-node/dist/gen/apis/StorageV1alpha1Api.js.map'
```

#### Examples
- Affected files: `StorageV1alpha1Api.js`, `CoreV1Api.js`, `util.js`, etc.
- Warning: See above log snippet.

#### Why It’s a Problem
Clutters dev logs, confuses developers, adds minor overhead.

#### Suggested Fix
- Include `.map` files in the package, or
- Remove `//# sourceMappingURL=...` comments during build (e.g., `--sourceMap false`).

#### Workaround
Patch locally with `patch-package`:
```
npm install --save-dev patch-package
[...]

PR #2260, opened 2025-02-23
include source maps in npm pack

cjihrig:
This commit includes the generated source maps in the artifact published to npm. This commit also simplifies the patterns used to match files. Some files, such as README are always included, so it has been removed here.

Refs: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#files
Fixes: https://github.com/kubernetes-client/javascript/issues/2249

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,13 +8,9 @@
         "url": "https://github.com/kubernetes-client/javascript.git"
     },
     "files": [
-        "dist/*.ts",
-        "dist/*.js",
-        "dist/gen/*.ts",
-        "dist/gen/*.js",
-        "dist/gen/**/*.ts",
-        "dist/gen/**/*.js",
-        "README.md"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,13 +8,9 @@
         "url": "https://github.com/kubernetes-client/javascript.git"
     },
     "files": [
-        "dist/*.ts",
```
cjihrig, line 11: @brendandburns I guess we could use these patterns as a mechanism to only include certain source maps (referring to your comment here https://github.com/kubernetes-client/javascript/issues/2249#issuecomment-2676442930), but as I said on the other issue, I'm not sure that's worth doing.

brendandburns:
/lgtm
/approve

Thanks for cleaning this up. I don't feel strongly about excluding the generated map files, if it's easy to create them.

PR #2265, opened 2025-02-24
bump jsonpath-plus minimum version to 10.3.0

kevinthenet:
[Resolves vulnerability reported by Snyk](https://security.snyk.io/vuln/SNYK-JS-JSONPATHPLUS-8719585).

For any new installations this should pick up `10.3.0`, but this change ensures there's no risk of installing any problematic versions.

- [x] `npm test` passes
- [x] `npm run generate` yields no errors

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
                 "hpagent": "^1.2.0",
                 "isomorphic-ws": "^5.0.0",
                 "js-yaml": "^4.1.0",
-                "jsonpath-plus": "^10.2.0",
+                "jsonpath-plus": "^10.3.0",
                 "node-fetch": "^2.6.9",
                 "openid-client": "^6.1.3",
                 "rfc4648": "^1.3.0",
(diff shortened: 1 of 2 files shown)


brendandburns:
/lgtm
/approve

I wonder why @dependabot didn't send a PR for this.

kevinthenet:
@brendandburns 
> I wonder why @dependabot didn't send a PR for this.

I recall seeing a PR/commit of the sort before v1.x, but I think the file regressed or was undone by another PR. In any case, it would be nice if we could cherry-pick this commit to the `0.22.3` tag so folks that haven't upgraded to 1.x have the vulnerability saved as well, though I'm not sure if you're still supporting `0.22.x`.

Happy to pitch in if needed.

cjihrig:
It was bumped by dependabot in https://github.com/kubernetes-client/javascript/pull/2236.

cjihrig:
There is a [v1.1.0 pending](https://github.com/kubernetes-client/javascript/commit/552e1dd5a49e52cbbb758684fb5b5250524fadfe), although I don't know the exact timeline for a release.

PR #2272, opened 2025-02-25
Improve coverage by adding unit tests for log.ts

brendandburns:
Largely generated by github copilot (ftw!)

Found a genuine bug where the fetch `response.body` wasn't being read as a stream.

Add an exclusion for some generated code that we probably don't want to test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
         ],
         "exclude": [
             "src/gen/*/**.ts",
+            "src/gen/**.ts",
             "src/index.ts",
             "src/*_test.ts",
             "src/test"
--- a/package.json
+++ b/package.json
@@ -33,8 +33,7 @@
             "src/**/*.ts"
         ],
         "exclude": [
-            "src/gen/*/**.ts",
-            "src/gen/**.ts",
+            "src/gen/**/*.ts",
             "src/index.ts",
             "src/*_test.ts",
             "src/test"
--- a/src/log_test.ts
+++ b/src/log_test.ts
@@ -87,7 +87,7 @@ describe('Log', () => {
                 },
             });
 
-            rejects(async () => {
+            await rejects(async () => {
                 await logWithoutCluster.log(namespace, podName, containerName, stream);
             }, /No currently active cluster/);
         });
@@ -107,7 +107,7 @@ describe('Log', () => {
                 .query({ container: 'mycontainer' })
                 .reply(501, { message: 'Error occurred in log request' });
 
-            rejects(async () => {
+            await rejects(async () => {
                 await log.log(namespace, podName, containerName, stream);
             }, /Error occurred in log request/);
         });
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
         ],
         "exclude": [
             "src/gen/*/**.ts",
+            "src/gen/**.ts",
```
cjihrig, line 37: I think this makes the line above it unnecessary.

src/log_test.ts
```diff
@@ -1,8 +1,177 @@
...
+            const namespace = 'default';
+            const podName = 'mypod';
+            const containerName = 'mycontainer';
+            const stream = new Writable({
+                write(chunk, encoding, callback) {
+                    callback();
+                },
+            });
+
+            rejects(async () => {
```
cjihrig, line 90: `rejects()` returns a Promise. We either need to `await` it or return it (I believe mocha supports that). Otherwise, if this assertion fails, it will lead to an unhandled rejection, which mocha ignores (I recently had to debug this exact case when a team was migrating from mocha to node:test).

src/log_test.ts
```diff
@@ -1,8 +1,177 @@
...
+                    callback();
+                },
+            });
+
+            nock('https://example.com')
+                .get('/api/v1/namespaces/default/pods/mypod/log')
+                .query({ container: 'mycontainer' })
+                .reply(501, { message: 'Error occurred in log request' });
+
+            rejects(async () => {
```
cjihrig, line 110: Same comment here.

src/log_test.ts
```diff
@@ -1,8 +1,177 @@
...
+                    callback();
+                },
+            });
+
+            nock('https://example.com')
+                .get('/api/v1/namespaces/default/pods/mypod/log')
+                .query({ container: 'mycontainer' })
+                .reply(500, { message: 'Error occurred in log request' });
+
+            rejects(async () => {
```
cjihrig, line 130: And here.

package.json
brendandburns: If I remove that line it prints coverage for all of the generated API files. I agree that semantically I feel like you are correct, but in practice it doesn't work right.

src/log_test.ts
brendandburns: done.

src/log_test.ts
brendandburns: done.

src/log_test.ts
brendandburns: done.

brendandburns:

package.json
cjihrig: Oh wait. I think it's because the existing pattern should be `src/gen/**/*.ts`. That fixes it for me locally.

mstruebing: As far as I understand globs, the `**.ts` would not be different than `*.s` and `**/*.ts` would pick up all ts files including subdirs. I've actually tested it right now with ls. So this should be enough:

```suggestion
        "exclude": [
            "src/gen/**/*.ts",
```

brendandburns: done.

brendandburns:
Thanks for the help with the glob! This is ready for (re)review

cjihrig (approved):
/lgtm
/approve

PR #2275, opened 2025-02-26
call time middleware support

davidgamero:
regenerate to support call-time middleware from upstream work in https://github.com/OpenAPITools/openapi-generator/pull/20430

this will enable the use of the `middleware` property of the `options` argument to append additional middleware to calls

in particular, we can support something like a WithHeader middleware that allows setting content type for json merge patch or even a wrapper that returns a full `options` type that can be passed

after unblocking the bugfix, I'll add the WithHeader middleware as an exported function returning a Middleware for ease of use

ex:
```typescript
// Options type
coreApi.putPod(
  { ... },
  WithHeader('Content-Type','application/json-patch+json'), // would require some chaining/ edits to WithHeader proposed implementation
)

// Middleware array type
coreApi.putPod(
  { ... },
  {
    middleware: WithHeaderMiddleware('Content-Type','application/json-patch+json'),
    middlewareMergeStrategy: 'append'
  }
)

//WithHeaderMiddleware implements Middleware[] Interface
function WithHeaderMiddleware(key: string, value: string): Middleware[]{
	return [{
		pre: (c: RequestContext)=> {
			return new Promise<RequestContext>((resolve)=>{
				c.headers[key] = value
				resolve(c)
			}
		)},
		post: (c: ResponseContext)=> {return new Promise<ResponseContext>((resolve)=>{ resolve(c) })},
	}]
}

// Returns RequestOptions that set a header
func WithHeader(key: string, value: string opt?: RequestOptions): RequestOptions{
  return {
    ...opt,
    middleware: WithHeaderMiddleware(key, value).concat(opt.middleware),
    middlewareMergeStrategy: 'append' // preserve chained middleware from opt
  }
}
```


davidgamero:
waiting on https://github.com/OpenAPITools/openapi-generator/pull/20745 to fix

brendandburns:
Looks like CI/CD is failing, seems like some error about node 16 or some such, do we need to update our Github action?

davidgamero:
The ci errors are related to file extensions in the generated files. 

I have a PR open upstream to fix it and can re generate once it merges

I'll try updating node on the actions if that fix doesn't resolve it

.github/workflows/test.yml
```diff
@@ -26,3 +26,14 @@ jobs:
             - run: npm run lint
             - run: npm audit --audit-level=critical
             - run: npm run build-with-tests && npm run test-transpiled
+            - name: Extract k8s version
+              id: extract_version
+              run: |
+                  version=$(grep 'export KUBERNETES_BRANCH=' settings | cut -d '"' -f 2)
+                  echo "setting kubernetes version to $KUBERNETES_V_STRING"
+                  echo "KUBERNETES_V_STRING=$version" >> "$GITHUB_ENV"
+            - name: start minikube
```
cjihrig, line 35: This appears to be failing the CI. Do we need these changes?

src/middleware.ts
```diff
@@ -0,0 +1,36 @@
+import type { RequestContext, ResponseContext, ConfigurationOptions, Middleware } from './gen/index.ts';
```
cjihrig, line 1: I think this probably gets transpiled away since it is all types, but for consistency, can we do this:
```suggestion
import type { RequestContext, ResponseContext, ConfigurationOptions, Middleware } from './gen/index.js';
```

src/middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+import { PromiseMiddleware } from './gen/middleware.js';
+
+//WithHeaderMiddleware returns Middleware[] that sets a header value
+export function WithHeaderMiddleware(key: string, value: string): Middleware[] {
```
cjihrig, line 5: The UpperCamelCase function names make this feel more like Golang than JavaScript.

src/middleware.ts
```diff
@@ -0,0 +1,36 @@
...
+    return [
+        {
+            pre: (c: RequestContext) => {
```
cjihrig, line 8: Would it work to make these `async` functions and get rid of the explicit `Promise` management?

cjihrig (commented):
It looks like there is a problem in the CI, and the bot does not recognize the most recent commit. Also, is it possible to add any unit tests?

davidgamero:
> It looks like there is a problem in the CI, and the bot does not recognize the most recent commit. Also, is it possible to add any unit tests?

yes i think the CI issue is this https://github.com/medyagh/setup-minikube/issues/565

how do we feel about adding something like minikube tests? i was trying to add an integration test that runs the patch example against a minikube cluster

i can add unit tests too, but they'd be mostly just testing the generated codes' features

cjihrig:
> how do we feel about adding something like minikube tests?

Seems like a good idea to me.

src/middleware.ts
davidgamero: i've updated these

.github/workflows/test.yml
davidgamero: they aren't critical, but i'd like to add some minikube integration tests. i can move that to a separate PR

davidgamero:
it appears the CLA doesn't respect force pushes, so i will have to open a new PR redoing this with all signed commits?

cjihrig:
What if you squash locally and force push this branch?

davidgamero:
you are right. that worked

brendandburns:
fwiw, we have KIND based e2e in the Java repo here:

https://github.com/kubernetes-client/java/blob/master/.github/workflows/maven.yml#L66

If you want to add equivalent e2e for this client that would be awesome.

cjihrig:
Maybe we should move the E2E tests to a separate PR and get this merged.

davidgamero:
> fwiw, we have KIND based e2e in the Java repo here:
> 
> https://github.com/kubernetes-client/java/blob/master/.github/workflows/maven.yml#L66
> 
> If you want to add equivalent e2e for this client that would be awesome.

i'll use kind then to keep it similar.

> Maybe we should move the E2E tests to a separate PR and get this merged.

agreed i'll remove the gh action changes

davidgamero:
i've added unit tests covering the two header-setting middleware utility functions for now, and will focus on e2e in a separate PR later

cjihrig (approved):
Nice work

/lgtm
/approve

PR #2288, opened 2025-03-04
Improve coverage and testability of config.ts

brendandburns:


src/config_test.ts
```diff
@@ -1630,5 +1634,60 @@ describe('KubeConfig', () => {
             strictEqual(inputData!.toString(), data);
             mockfs.restore();
         });
+        it('should try to load from WSL on Windows with wsl.exe not working', () => {
+            const kc = new KubeConfig();
+            const commands: { command: string; args: string[] }[] = [];
+            (kc as any).spawnSync = (cmd: string, args: string[]) => {
```
cjihrig, line 1640: No need to change this PR, but just something for future consideration - all supported Node release lines do have this type of mocking and spying built in.

cjihrig (commented):
LGTM minus the `console.log()`.

src/config_test.ts
```diff
@@ -322,6 +324,8 @@ describe('KubeConfig', () => {
             };
 
             assertRequestOptionsEqual(opts, expectedOptions);
+            console.log(requestContext.getAgent());
```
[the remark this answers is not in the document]
brendandburns: dang it! (done)

src/config_test.ts
brendandburns: Thanks for pointing this out, it didn't show up when I looked around for mocking `child_process` but now I found it and I refactored the test to use this instead.

brendandburns:

src/config_test.ts
cjihrig: Hm, I'm actually surprised that the CI is passing. Mocking an ES module like that should not work because they have immutable bindings. When I pull this PR down locally, these tests fail.

The approach here would work with the previous state of the PR though, where you were mocking the `KubeConfig` object's method. I think we should go back to that.

ESM mocking is quite the can of worms. There is support for that in newer versions of Node (and some ecosystem modules as well), but they build on top of Node's ESM loaders, which have been extremely unstable and are still experimental.

cjihrig: I dug into this a bit more. The mock is surprisingly working. The reason the tests are failing on my local machine is that `loadFromDefault()` is returning early [here](https://github.com/kubernetes-client/javascript/blob/3b44c174a71506d2a0e6aac36f56d5539887b1bf/src/config.ts#L412-L419), so the mocked `spawnSync()` calls never happen.

cjihrig: This gets the new tests passing for me locally:

```diff
diff --git a/src/config_test.ts b/src/config_test.ts
index 9bb31dd5..30f4d49f 100644
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -1613,6 +1613,17 @@ describe('KubeConfig', () => {
     });
 
     describe('BufferOrFile', () => {
+        let originalEnv;
+
+        before(() => {
+            originalEnv = process.env;
+            process.env = {};
+        });
+
+        after(() => {
+            process.env = originalEnv;
+        });
+
         it('should load from root if present', () => {
             const data = 'some data for file';
             const arg: any = {
```

src/config_test.ts
```diff
@@ -322,6 +326,7 @@ describe('KubeConfig', () => {
             };
 
             assertRequestOptionsEqual(opts, expectedOptions);
+            strictEqual((requestContext.getAgent()! as any).options.servername, 'kube.example2.com');
```
mstruebing, line 329: Isn't this rather an `https.Agent` than an `any?`
This would remove the unnecessary type casting but you probably still need the bang-operator or a different check.
Not sure if it's worth it, just wanted to mention that.

mstruebing:
/lgtm

I wasn't ware of this form of mocking either, looks very cool!
LGTM but I will let @cjihrig re-review as he is much deeper into the mocking and Node/ESM stuff.

src/config_test.ts
brendandburns: options isn't exposed as public (at least not in the typings), so Typescript doesn't like it. The cast as `any` is mostly to get the typescript compiler to leave it alone.

Issue #2160, opened 2025-01-14
Patch fails with `error decoding patch: json: cannot unmarshal object into Go value of type []handlers.jsonPatchOp`


yubinnng:
**Describe the bug**

I would like to patch a resource with a request body in [JSON Merge Patch, RFC 7386](https://tools.ietf.org/html/rfc7386) format. However, even if I specify the Content-Type `application/merge-patch+json` in a `PromiseMiddlewareWrapper`, add it into a `ServerConfiguration` and then use the configuration in the client method `k8sApi.patchNamespacedPod({xxx}, configuration)`, the error still occurs:

```
ApiException [Error]: HTTP-Code: 400
Message: Unknown API Status Code!
Body: "{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"error decoding patch: json: cannot unmarshal object into Go value of type []handlers.jsonPatchOp\",\"reason\":\"BadRequest\",\"code\":400}\n"
Headers: {"audit-id":"f2f1b40e-9d4a-49cf-af2b-989bed7de17e","cache-control":"no-cache, private","connection":"close","content-length":"211","content-type":"application/json","date":"Tue, 14 Jan 2025 02:17:01 GMT","x-kubernetes-pf-flowschema-uid":"93068417-329d-4e0f-9768-8a2b6794f954","x-kubernetes-pf-prioritylevel-uid":"f7dc9ee7-016c-48fa-afc3-77112540d77f"}
    at CoreV1ApiResponseProcessor.patchNamespacedPodWithHttpInfo (/Users/I584817/Downloads/js-examples/node_modules/@kubernetes/client-node/dist/gen/apis/CoreV1Api.js:15722:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///Users/XXX/Downloads/js-examples/merge-patch-example.js:46:5 {
  code: 400,
  body: '{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"error decoding patch: json: cannot unmarshal object into Go value of type []handlers.jsonPatchOp","reason":"BadRequest","code":400}\n',
  headers: {
    'audit-id': 'f2f1b40e-9d4a-49cf-af2b-989bed7de17e',
    'cache-control': 'no-cache, private',
    connection: 'close',
    'content-length': '211',
    'content-type': 'application/json',
[...]

PR #2296, opened 2025-03-05
Prepare for 1.0.1 release

brendandburns:
cc @cjihrig 

Related to #2160

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@kubernetes/client-node",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@kubernetes/client-node",
-            "version": "1.0.0",
+            "version": "1.0.1",
             "license": "Apache-2.0",
             "dependencies": {
                 "@types/js-yaml": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@kubernetes/client-node",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "description": "NodeJS client for kubernetes",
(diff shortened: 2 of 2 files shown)


cjihrig (approved):
/lgtm
/approve

cjihrig:
It just occurred to me, this should be v1.1.0 because we have new APIs in https://github.com/kubernetes-client/javascript/pull/2275.

cjihrig:
An "rc" might be a good idea due to the shear size of #2275, but I don't feel strongly one way or the other.

brendandburns:
@cjihrig can you clarify? I looked at that PR, I see new (fixed?) behavior, but no new APIs and no breaking changes.

I feel like this is sufficient to mark as a patch release, not a minor version, but I'm open to discussion.

brendandburns:
Given that they are additive, I wouldn't consider it a breaking change (which is typically my bar for minor vs patch)

brendandburns:
Personally, I don't feel a strong need for RC, since people can always revert back to 1.0.0 if they encounter problems.

Also most of the changes in that PR are identical changes, just in lots of different files.

Ultimately we should probably move to a branch and cherry-pick model instead of release from head, that would also give better explanations of what changed.

davidgamero:
> Given that they are additive, I wouldn't consider it a breaking change (which is typically my bar for minor vs patch)

I see- by that bar it definitely isn't a breaking change and is fully backwards compatible, so the rc isn't necessary

cjihrig:
> Given that they are additive, I wouldn't consider it a breaking change (which is typically my bar for minor vs patch)

OK. I was going by semver. But if this repo doesn't use semver, then 1.0.1 is fine.

/unhold

brendandburns:
@cjihrig I re-read the semver spec and you're right.

It's probably worth following semver, I'll send another PR.

PR #2311, opened 2025-03-10
Add the ability to inject authenticators.

### LINKED ISSUE #2263: Can the ability to customize Authenticators be opened up?
I've noticed that `KubeConfig` provides several static `Authenticator` implementations. However, I need to add additional authentication methods for my internal use cases. Could you make the `authenticators` property in `KubeConfig` configurable or provide other methods to extent authentication methods?

brendandburns:
Fixes #2263

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -65,6 +65,12 @@ export class KubeConfig implements SecurityAuthentication {
         new OpenIDConnectAuth(),
     ];
 
+    // Optionally add additional external authenticators, you must do this
+    // before you load a kubeconfig file that references them.
+    public static addAuthenticator(authenticator: Authenticator): void {
+        this.authenticators.push(authenticator);
+    }
+
     /**
      * The list of all known clusters
      */
--- a/src/config.ts
+++ b/src/config.ts
@@ -65,10 +65,13 @@ export class KubeConfig implements SecurityAuthentication {
         new OpenIDConnectAuth(),
     ];
 
+    // List of custom authenticators that can be added by the user
+    private custom_authenticators: Authenticator[] = [];
+
     // Optionally add additional external authenticators, you must do this
     // before you load a kubeconfig file that references them.
-    public static addAuthenticator(authenticator: Authenticator): void {
-        this.authenticators.push(authenticator);
+    public addAuthenticator(authenticator: Authenticator): void {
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -65,6 +65,12 @@ export class KubeConfig implements SecurityAuthentication {
         new OpenIDConnectAuth(),
     ];
 
+    // Optionally add additional external authenticators, you must do this
+    // before you load a kubeconfig file that references them.
```
cjihrig, line 69: Do you think it is worth adding a test that intentionally does not do this (if for no other reason than knowing if it accidentally changes)?

cjihrig (commented):
This LGTM. Just one optional comment. If you don't want to address it, I'm fine with merging as is.

src/config.ts
```diff
@@ -65,6 +65,12 @@ export class KubeConfig implements SecurityAuthentication {
...
+    public static addAuthenticator(authenticator: Authenticator): void {
+        this.authenticators.push(authenticator);
```
davidgamero, line 71: is the intended use here to add them to *all* existing KubeConfig objects since the `authenticators` property is static?

also, could we avoid the loading order issue by adding an optional parameter for an array of `authenticators` to the constructor instead? that way the authenticators must be added prior to loading a config

src/config.ts
brendandburns: I don't think this can ever change, because the kubeconfig file references the authenticator as part of the YAML file, so if we don't know about the authenticator because it hasn't been registered when we load the kubeconfig, there's no way that we can link the authenticator requested in the file to the kubeconfig.

So while we could add a test, I'm not sure it would ever break.

src/config.ts
brendandburns: I'm ok with this, but I also think it makes it a little more complicated. I wanted the authenticators to be able to self-register just by being `import`-ed which requires the static registration.  @cjihrig wdyt?

anandfresh (approved):

brendandburns:
Ok, I switched this to custom authenticators that are local to the `KubeConfig` instance.

Please take another look.

cjihrig:
LGTM. I'll give @davidgamero a chance to weigh in before merging.

cjihrig (approved):
/lgtm
/approve

PR #2316, opened 2025-03-15
update HeaderMiddleware types, export PatchStrategy with Integration Tests

davidgamero:
- Update `setHeaderMiddleware` and `setHeaderOptions` to ObservableMiddleware type to match our exported type s

- Export `PatchStrategy` from the index

- Add `integration-test` script

- Add kind cluster integration test that validates these changes

ref https://github.com/kubernetes-client/javascript/issues/2160

It would be great to execute the examples _as integration tests_ against the kind cluster in the future to ensure examples must remain functional

### DIFF
--- a/examples/typescript/patch/patch-example.ts
+++ b/examples/typescript/patch/patch-example.ts
@@ -1,4 +1,5 @@
-import { CoreV1Api, KubeConfig, setHeaderOptions, JsonPatch } from '@kubernetes/client-node';
+import { CoreV1Api, KubeConfig, setHeaderOptions } from '@kubernetes/client-node';
+import { PatchStrategy } from '@kubernetes/client-node/patch';
 
 const kc = new KubeConfig();
 kc.loadFromDefault();
@@ -28,7 +29,7 @@ try {
             namespace: 'default',
             body: patch,
         },
-        setHeaderOptions('Content-Type', JsonPatch),
+        setHeaderOptions('Content-Type', PatchStrategy.JsonPatch),
     );
 
     console.log('Patched.');
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,8 @@
                 "prettier": "^3.0.0",
                 "pretty-quick": "^4.0.0",
                 "ts-mockito": "^2.3.1",
-                "tsx": "^4.19.1",
+                "ts-node": "^10.9.2",
+                "tsx": "^4.19.3",
                 "typedoc": "^0.27.1",
                 "typescript": "~5.8.2",
                 "typescript-eslint": "^8.26.0"
@@ -56,6 +57,30 @@
             "dev": true,
             "engines": {
                 "node": ">=18"
+            }
+        },
+        "node_modules/@cspotcode/source-map-support": {
+            "version": "0.8.1",
+            "resolved": "https://registry.npmjs.org/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz",
+            "integrity": "sha512-IchNf6dN4tHoMFIn/7OE8LWZ19Y6q/67Bmf6vnGREv8RSbBVb9LPJxEcnwrcwX6ixSvaiGoomAUvu4YSxXrVgw==",
+            "dev": true,
+            "license": "MIT",
+            "dependencies": {
+                "@jridgewell/trace-mapping": "0.3.9"
+            },
+            "engines": {
+                "node": ">=12"
+            }
+        },
+        "node_modules/@cspotcode/source-map-support/node_modules/@jridgewell/trace-mapping": {
+            "version": "0.3.9",
+            "resolved": "https://registry.npmjs.org/@jridgewell/trace-mapping/-/trace-mapping-0.3.9.tgz",
+            "integrity": "sha512-3Belt6tdc8bPgAtbcmdtNJlirVoTmEb5e2gC94PnkwEW9jI6CAHUeoG85tjWP5WquqfavoMtMwiG4P926ZKKuQ==",
+            "dev": true,
+            "license": "MIT",
+            "dependencies": {
+                "@jridgewell/resolve-uri": "^3.0.3",
+                "@jridgewell/sourcemap-codec": "^1.4.10"
             }
         },
         "node_modules/@esbuild/aix-ppc64": {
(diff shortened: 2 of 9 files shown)


src/test/integration/patchNamespace.ts
```diff
@@ -0,0 +1,47 @@
...
+        {
+            name: testNSName,
+            body: patch,
+        },
+        setHeaderOptions('Content-Type', PatchStrategy.JsonPatch),
+    );
+
+    const patchedNS = await coreV1Client.readNamespace({ name: testNSName });
+    deepEqual(patchedNS.metadata?.labels?.foo, 'bar');
+    deepEqual(patchedNS.metadata?.labels?.initialLabel, undefined);
```
mstruebing, line 45: I guess it doesn't hurt but is there a reason whe need `deepEqual` here instead of just `equal` as we are just comparing primitive values?

src/test/integration/patchNamespace.ts
```diff
@@ -0,0 +1,47 @@
+import assert, { deepEqual } from 'node:assert';
+import { PatchStrategy, CoreV1Api, KubeConfig, setHeaderOptions, V1Namespace } from '../../index.js';
+
+export default async function patchNamespace() {
+    const kc = new KubeConfig();
+    kc.loadFromDefault();
+
+    const coreV1Client = kc.makeApiClient(CoreV1Api);
+    let newNS = new V1Namespace();
+    const testNSName = 'integration-test-patch-ns';
```
mstruebing, line 10: If we would add a variable at the end i.e. some random generated string or use `generateName` instead of name we could run the test multiple times against the same cluster without any cleanup work.

We may don't need that often, but it could be good for local debugging.

mstruebing (commented):
Great work!
Just two small comments/nitpicks.

examples/typescript/patch/patch-example.ts
```diff
@@ -1,4 +1,5 @@
-import { CoreV1Api, KubeConfig, setHeaderOptions, JsonPatch } from '@kubernetes/client-node';
+import { CoreV1Api, KubeConfig, setHeaderOptions } from '@kubernetes/client-node';
+import { PatchStrategy } from '@kubernetes/client-node/patch';
```
cjihrig, line 2: Can we do this without requiring a second import of the client package?

src/test/integration/index.ts
```diff
@@ -0,0 +1,5 @@
+import patchNamespace from './patchNamespace.js';
+
+console.log('Integration testing');
+
+patchNamespace();
```
cjihrig, line 5: Should we await this? I'm assuming we would add more tests here in the future.

package-lock.json
```diff
@@ -43,7 +43,8 @@
                 "prettier": "^3.0.0",
                 "pretty-quick": "^4.0.0",
                 "ts-mockito": "^2.3.1",
-                "tsx": "^4.19.1",
+                "ts-node": "^10.9.2",
```
cjihrig, line 46: What caused the changes to this file?

src/test/integration/patchNamespace.ts
```diff
@@ -0,0 +1,47 @@
...
+    newNS.metadata = {
+        name: testNSName,
+        labels: {
+            initialLabel: 'initialValue',
+        },
+    };
+    console.log(`Creating namespace ${testNSName}`);
+    await coreV1Client.createNamespace({ body: newNS });
+    newNS = await coreV1Client.readNamespace({ name: testNSName });
+    assert(newNS.metadata?.name === testNSName);
```
cjihrig, line 20:
```suggestion
    assert.strictEqual(newNS.metadata?.name, testNSName);
```

src/test/integration/patchNamespace.ts
```diff
@@ -0,0 +1,47 @@
...
+    deepEqual(newNS.metadata?.labels?.initialLabel, 'initialValue');
```
cjihrig, line 21:
```suggestion
    assert.strictEqual(newNS.metadata?.labels?.initialLabel, 'initialValue');
```

src/middleware_test.ts
```diff
@@ -1,33 +1,36 @@
...
 describe('Middleware', async () => {
     describe('setHeaderMiddleware', async () => {
         it('should set a header when it is blank', async () => {
             const reqContext = new RequestContext('http://nowhere.com', HttpMethod.GET);
             deepStrictEqual(reqContext.getHeaders(), {});
             const headerMiddleware = setHeaderMiddleware('test-key', 'test-value');
-            const postMiddlewareRequest = await headerMiddleware[0].pre(reqContext);
-            deepStrictEqual(postMiddlewareRequest.getHeaders(), { 'test-key': 'test-value' });
+            const postMiddlewareRequest = await headerMiddleware.pre(reqContext);
+            await postMiddlewareRequest.toPromise().then((request) => {
```
cjihrig, line 12: Can we avoid mixing `await` and `then()` like this in this file. From a quick glance, it's not immediately obvious what happens if these assertions fail. If it causes an unhandledRejection, I'm not sure if Mocha will report it properly.

cjihrig (commented):
Left a few minor comments, but this is great. Thanks.

src/test/integration/index.ts
davidgamero: yes, that was what i was thinking

src/test/integration/patchNamespace.ts
davidgamero: agreed

src/middleware_test.ts
davidgamero: yes thank you for catching that

examples/typescript/patch/patch-example.ts
davidgamero: yes. updated

package-lock.json
davidgamero: i reverted this file now

davidgamero: must have been something during testing

src/middleware_test.ts
cjihrig: There seems to be a couple more instances of this pattern. Other than that, this LGTM.

davidgamero:
@cjihrig updated with the last couple changes for await syntax feedback

Issue #2306, opened 2025-03-09
Migrate to internal node implementation of fetch


brendandburns:
We are going to remove the `node-fetch` dependency and move to node's internal implementation.

davidgamero:
Upstream PR on the generator has merged to use undici.
https://github.com/OpenAPITools/openapi-generator/pull/22408

It relies on the "undici" module instead of using node-provided undici as features like interceptors, middleware, and stream are not yet available in the node native version.
https://undici.nodejs.org/#/?id=undici-module

I tested locally regenerating the client. As expected, most of the types in Config that rely on https/node-fetch are subtly different than the undici interfaces and will require updates.

I havent tested the mocking yet. As previously discussed, nock will no longer work for our tests with undici.

davidgamero:
linking some PRs here to keep things organized

first regeneration with native undici fetch from upstream #2770 
phase 2 with full removal of node-fetch and nock packages #2813

davidgamero:
pre-release 2.0.0-rc.1 with native fetch is live

https://www.npmjs.com/package/@kubernetes/client-node/v/2.0.0-rc.1

feedback from early adopters would very appreciated if you give it a try

PR #2333, opened 2025-03-27
begin migration to native fetch()

cjihrig:
Refs: https://github.com/kubernetes-client/javascript/issues/2306

A few notes:

- This unblocks the nock update. In fact, I believe it necessitates the update because the older nock version didn't seem to support native fetch.
- There is future room for improvement related to the use of web streams. Native fetch uses web streams, while node-fetch uses Node streams. Just to get this migration going, I used a conversion function (search the diff for `Readable.fromWeb()`), but ideally we would try working with web streams directly to reduce some overhead.
- `node-fetch` is still used in `src/gen/http/isomorphic-fetch.ts`, which I think would need to be resolved in the code generator.
- In `src/util.ts`, there is a `normalizeResponseHeaders()` that apparently exists due to a node-fetch quirk. I did not look into whether or not that is still needed, but keeping it doesn't seem to break anything.
- `FetchError` does not appear to be a thing in native fetch.

We may or may not want to keep this as a draft PR because of this list ~~and pending the CI~~ - what do others think?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
                 "husky": "^9.0.6",
                 "mocha": "^11.0.1",
                 "mock-fs": "^5.2.0",
-                "nock": "^13.2.9",
+                "nock": "^14.0.2",
                 "prettier": "^3.0.0",
                 "pretty-quick": "^4.0.0",
                 "ts-mockito": "^2.3.1",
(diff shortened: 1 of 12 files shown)


mstruebing:
That's pretty cool already!

I don't think `Readable.fromWeb()` is that bad actually, sure a little bit of overhead but nothing serious.
I had a quick look into the code gen but I'm a bit overwhelmed right now.
Not sure if we can adjust the current gen code or if we would need a different/new one, as we are using a different fetching library.

cjihrig:
My guess is that we'd need a different code generator (or at least new capabilities in the existing one) since there are enough small, subtle differences between node-fetch and native fetch.

brendandburns:
Looks good to me, my only question is for the `package.json` there's a lot of additional dependencies that seem kind of random. Is that all really from the `nock` upgrade?

cjihrig:
> Is that all really from the nock upgrade?

As far as I can tell, yes. The joys of depending on things from npm. nock started depending on `@mswjs/interceptors` (this is the thing that broke our tests), which brings in its own dependencies, like jsdom. You can verify any of them like this:

```
$ npm ls @asamuzakjp/css-color
@kubernetes/client-node@1.1.1 /Users/cjihrig/programming/javascript
└─┬ nock@14.0.2
  └─┬ @mswjs/interceptors@0.38.0
    └─┬ jsdom@26.0.0
      └─┬ cssstyle@4.3.0
        └── @asamuzakjp/css-color@3.1.1
```

cjihrig:
You can also check out this dependabot PR from last night that tried to make the same update - https://github.com/kubernetes-client/javascript/pull/2337.

brendandburns:
/lgtm
/approve

Thanks.

PR #2343, opened 2025-04-01
Add 1.1.x to README

davidgamero:
To stay consistent with the client compatibility table, i added the 1.1.x minor version release to the README

Since we are following semver, we won't have 1 minor version per k8s version anymore


cjihrig (approved):
/lgtm
/approve

PR #2351, opened 2025-04-03
fix: apply http agent if needed

### LINKED ISSUE #2348: TypeError: Protocol "http:" not supported. Expected "https:"
So yeah I spent like 2 hours deep dive debugging into this only to discover a bug.

https://github.com/kubernetes-client/javascript/blob/3283cb12f146baf0ced6061d06697a3ecd25917a/src/config.ts#L527-L551

more specifically

https://github.com/kubernetes-client/javascript/blob/3283cb12f146baf0ced6061d06697a3ecd25917a/src/config.ts#L533-L536

if no `proxyUrl` is set, then the default will be an HTTPS agent, which does not work with `kubectl proxy` environments.

mstruebing:
This applies an http agent if needed.
I'm not entirely sure about the security implications of this, but I guess it should be not a problem though.

Fixes #2348

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -545,8 +545,10 @@ export class KubeConfig implements SecurityAuthentication {
             } else {
                 throw new Error('Unsupported proxy type');
             }
-        } else if (cluster?.server?.startsWith('http:')) {
+        } else if (cluster?.server?.startsWith('http:') && cluster.skipTLSVerify) {
             agent = new http.Agent(agentOptions);
+        } else if (cluster?.server?.startsWith('http:') && !cluster.skipTLSVerify) {
+            throw new Error('HTTP protocol is not allowed when skipTLSVerify is not set or false');
         } else {
             agent = new https.Agent(agentOptions);
         }
--- a/src/config.ts
+++ b/src/config.ts
@@ -545,7 +545,7 @@ export class KubeConfig implements SecurityAuthentication {
             } else {
                 throw new Error('Unsupported proxy type');
             }
-        } else if (cluster && cluster.server && cluster.server.startsWith('http:')) {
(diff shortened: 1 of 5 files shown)


src/config_test.ts
```diff
@@ -298,7 +299,7 @@ describe('KubeConfig', () => {
         });
     });
 
-    describe('applyHTTPSOptions', () => {
+    describe.only('applyHTTPSOptions', () => {
```
cjihrig, line 302: Can you remove the `only`.

src/config.ts
```diff
@@ -544,6 +545,8 @@ export class KubeConfig implements SecurityAuthentication {
             } else {
                 throw new Error('Unsupported proxy type');
             }
+        } else if (cluster && cluster.server && cluster.server.startsWith('http:')) {
```
cjihrig, line 548:
```suggestion
        } else if (cluster?.server?.startsWith('http:')) {
```

src/config_test.ts
mstruebing: oops, leftover.

src/config.ts
```diff
@@ -544,6 +545,8 @@ export class KubeConfig implements SecurityAuthentication {
             } else {
                 throw new Error('Unsupported proxy type');
             }
+        } else if (cluster?.server?.startsWith('http:')) {
```
brendandburns, line 548: Can we also check for `skipTlsVerify` being `true` before we allow http URLs? It's really not secure to do this.

mstruebing: Pushed a commit to check for `skipTlsVerify` as well and added a test for this.
I agree that it's not secure but I think the library should allow this as it is possible to do. 🤔

brendandburns:
/lgtm
/approve

Thanks!

PR #2356, opened 2025-04-04
Update release.yml

brendandburns:
There was a typo between the input parameter and the steps which meant that we weren't actually tagging anything.


cjihrig (approved):
/lgtm
/approve

PR #2357, opened 2025-04-04
Fix release.yml (again)

brendandburns:
So out of curiosity, I decided to ask Copilot if it could figure out what was wrong with the YAML file and it identified another issue (`events` vs `event`) and also updated the push command for clarity.


cjihrig (approved):
From a quick search, this does appear to be right.

/lgtm
/approve

Issue #2387, opened 2025-04-23
Watcher `done` callback may be invoked twice on unexpected connection loss


bverhoeven:
**Describe the bug**

When a watch connection closes abruptly, the `done` callback may be invoked multiple times because of reentrancy in the `doneCallOnce()` function.

**Client Version**
`1.1.2`

**Server Version**
`v1.32.2-gke.1182003`

**To Reproduce**

1. Use `Watch` to start a stream (e.g. watch pods in a namespace).
1. Simulate a premature close on the connection (e.g. by destroying the socket on the server side).
1. The `done` callback is called with "AbortError: The user aborted a request"
1. The `done` callback is called again, but now with "Error: Premature close"

**Expected behavior**

The `done` callback should only be called once per watch termination.

**Example Code**

The code below can be used to replicate the issue on a local machine.

```ts
import http from 'http'
import { KubeConfig, Watch } from '@kubernetes/client-node'

const MOCK_PORT = 8333
const MOCK_URL = `http://localhost:${MOCK_PORT}`

const minimalServer = http.createServer((req, res) => {
    res.writeHead(200, {
        'Content-Type': 'application/json',
        'Transfer-Encoding': 'chunked',
    })

    res.flushHeaders()
    res.destroy() // Prematurely close the connection
})

minimalServer.listen(MOCK_PORT, () => {
    const kubeConfig = new KubeConfig()

    kubeConfig.loadFromClusterAndUser(
        {
            name: 'mock-cluster',
            server: MOCK_URL,
            skipTLSVerify: true,
        },
        { name: 'mock-user' }
    )

    const watch = new Watch(kubeConfig)

    watch.watch(
        `/api/v1/namespaces/default/pods`,
        {},
        () => { },
        (err) => console.log('done()', err)
    )
})
```

Output:

```
done() AbortError: The user aborted a request.
    at abort (/home/bas/projects/watch-repro/node_modules/node-fetch/lib/index.js:1458:16)
    at AbortSignal.abortAndFinalize (/home/bas/projects/watch-repro/node_modules/node-fetch/lib/index.js:1473:4)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:827:20)
    at AbortSignal.dispatchEvent (node:internal/event_target:762:26)
    at runAbort (node:internal/abort_controller:449:10)
    at abortSignal (node:internal/abort_controller:435:3)
    at AbortController.abort (node:internal/abort_controller:468:5)
[...]

Issue #2385, opened 2025-04-22
ListWatch stops reconnecting on connection errors


bverhoeven:
**Describe the bug**

The ListWatch class stops reconnecting when a non-410 error occurs during a watch session, due to an early return in the `doneHandler()` function.

**Client Version**
`1.1.2`

**Server Version**
`v1.32.2-gke.1182003`

**To Reproduce**

1. Use `Listwatch` to watch a namespace that doesn't have any pods (to get a stream that doesn't really send anything)
1. Do not specify a `timeoutSeconds` in the request (as that seems to cause a clean disconnect)
1. Observe that the "connect" event is emitted.
1. After 5 minutes or upon a connection error, the watch ends with a set of errors:
   * AbortError: The user aborted a request.
   * Error: Premature close (code `ERR_STREAM_PREMATURE_CLOSE`)
1. No further attempts to reconnect are made (no "connect" events are emitted)

Note: I can easily reproduce this issue in our GKE cluster, but I can understand that the timeout here may be a bit tricky to reproduce in other environments.

See my script below for a mock server that simulates the issue on a local machine.

**Expected behavior**

I would expect the watch to attempt to reconnect on a connection error, not just 410 errors.

**Example Code**

I created a mock server to reproduce the issue. The server will:

* respond with an initial list of pods to non-watch requests
* cleanly close the connection for the first watch request
* prematurely close the connection for the second watch request
* cleanly close the connection for the third and subsequent watch requests
  (although we'll not reach that point in this example)

Observe that:
* When the request is cleanly closed, the watch reconnects and the "connect" event is emitted.
* When the request is closed prematurely, the watch does not reconnect and the "connect" event is not emitted, instead two errors are reported:
  * `AbortError: The user aborted a request.`
  * `Error: Premature close`

```ts
import http from 'http'
import { CoreV1Api, KubeConfig, ListWatch, Watch } from '@kubernetes/client-node'

const MOCK_PORT = 8333
const MOCK_URL = `http://localhost:${MOCK_PORT}`

let requestCount = 0

const server = http.createServer((req, res) => {
    res.writeHead(200, {
        'Content-Type': 'application/json',
        'Transfer-Encoding': 'chunked',
    })
    res.flushHeaders()

    if (req.url?.includes('watch=true')) {
        if (requestCount++ == 1) {
            console.log('[mock-server] watch: Closing connection prematurely')
            res.destroy()
        } else {
[...]

PR #2389, opened 2025-04-24
watch: prevent done callback from being called twice on connection loss

bverhoeven:
Avoids multiple invocations of the done callback when a watch connection is closed unexpectedly (e.g., a premature socket close). The fix sets `doneCalled = true` before calling `controller.abort()`, to ensure that any `AbortError` triggered by aborting does not result in another call to the done callback.

A regression test is included to simulate an abrupt connection termination. It uses a real HTTP server that flushes headers and then destroys the connection immediately. This reliably triggers the combination of `AbortError` and "Premature close" needed to reproduce the issue.

Refs: https://github.com/kubernetes-client/javascript/issues/2387

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -46,8 +46,8 @@ export class Watch {
         let doneCalled: boolean = false;
         const doneCallOnce = (err: any) => {
             if (!doneCalled) {
-                controller.abort();
                 doneCalled = true;
+                controller.abort();
                 done(err);
             }
         };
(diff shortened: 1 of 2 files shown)


bverhoeven:
Thanks! Yeah, I spent quite a bit of time trying to get this working with nock, but couldn’t get it to reliably simulate an abrupt disconnect. As mentioned in [this comment](https://github.com/kubernetes-client/javascript/issues/2385#issuecomment-2823757440), this setup seemed like the most dependable way to reproduce the issue.

Happy to adjust if maintainers prefer a different approach, but I figured it was worth using something that reliably triggers the edge case.

PR #2402, opened 2025-04-29
Move to tar-fs.

### LINKED ISSUE #2392: BlueOak-1.0.0 license is not approved by CNCF
`package.json` lists several dependencies that are licensed under the `BlueOak-1.0.0` license. That license is not listed as an [approved CNCF license](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#cncf-allowlist-license-policy).

The packages are also not listed in the [exception lists](https://github.com/cncf/foundation/tree/main/license-exceptions).

Are there any plans to resolve this?

Possible ways:
* Get the license added to the CNCF Approved List
* Apply for a CNCF exception for the specific modules
* Ask the maintainer of the modules to add a dual-license for one of the approved licenses
* Replace the dependencies with something else

brendandburns:
Fixes #2392

tar-fs only depends on MIT, ISC and Apache 2 licensed dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
         "rfc4648": "^1.3.0",
         "socks-proxy-agent": "^8.0.4",
         "stream-buffers": "^3.0.2",
-        "tar": "^7.0.0",
         "tar-fs": "^3.0.8",
         "tmp-promise": "^3.0.2",
         "ws": "^8.18.0"
--- a/package.json
+++ b/package.json
@@ -59,8 +59,6 @@
         "@types/node": "^22.0.0",
         "@types/node-fetch": "^2.6.9",
         "@types/stream-buffers": "^3.0.3",
-        "@types/tar": "^6.1.1",
-        "@types/ws": "^8.5.4",
         "form-data": "^4.0.0",
         "hpagent": "^1.2.0",
         "isomorphic-ws": "^5.0.0",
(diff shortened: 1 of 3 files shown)


src/cp.ts
```diff
@@ -1,10 +1,12 @@
-import fs from 'node:fs';
+import fs, { rmSync } from 'node:fs';
 import { WritableStreamBuffer } from 'stream-buffers';
-import * as tar from 'tar';
```
cjihrig, line 3: There are some linter errors in this file.

package.json
```diff
@@ -72,6 +72,7 @@
         "socks-proxy-agent": "^8.0.4",
         "stream-buffers": "^3.0.2",
         "tar": "^7.0.0",
```
cjihrig, line 74: `tar` should be removed from this file and `package-lock.json`

cjihrig (commented):
Left a couple comments.

I'm good with this approach, but I do think it might be worthwhile to also pursue a license exception for BlueOak. I realize JavaScript isn't the primary language of most CNCF projects, but I would be surprised if this didn't happen again in some CNCF controlled repo. My understanding is that Isaac (the author of all of the packages in #2392) intends to move all of his modules to this license over time, and a lot of those modules are heavily depended on.

src/cp.ts
brendandburns: fixed.

package.json
brendandburns: fixed.

src/cp.ts
```diff
@@ -1,7 +1,5 @@
-import fs from 'node:fs';
 import { WritableStreamBuffer } from 'stream-buffers';
-import * as tar from 'tar';
-import tmp from 'tmp-promise';
```
cjihrig, line 4: Sorry for the back and forth - I think this is the only use of this dependency in the project. If it's removed here, it can be removed from package.json and package-lock.json too.

brendandburns: Done (and removed `ws` also since it is not a direct dependency)

cjihrig (approved):
Thanks. It's great to remove dependencies.

/lgtm
/approve

Issue #2424, opened 2025-05-07
False update events from informer after upgrading from 0.22 to 1.2.0 on AKS clusters


tonirvega:
**Describe the bug**

We noticed that after migrating from version 0.22 to 1.2.0 (and also in 1.1.2), we began receiving false update events from the informer event listener. The object hasn't changed in etcd, yet these events appear to occur approximately every 40 minutes. This behavior did not occur in version 0.22—we verified this through the logs in Datadog.

datadog logs

![Image](https://github.com/user-attachments/assets/7676007d-5904-4703-832b-ffb322e3571f)

![Image](https://github.com/user-attachments/assets/18de1dfc-2d48-41cb-ab4a-bf020bac0767)

To further confirm the issue, we ran the following command:

```shell
kubectl get githubmembership juanjosevazquezgil-68512dbc-257d-48f1-a53a-055f94b8ab7f -n firestartr-github --show-managed-fields=true -o yaml --output-watch-events=true --watch
```

During the same time period, no changes to the object were shown.

We're deploying the same version of our service in an Amazon EKS cluster, where this behavior does not occur.

Additionally, we use a Kind cluster for local development, and we haven’t observed this issue there either.

**Client Version**
 `1.2.0`

**Server Version**
`1.30.11`

**Example Code**
```js
import { getConnection } from './ctl';
import Debug from 'debug';
import common from 'catalog_common';
const log = Debug('firestartr:operator:reflector');
import * as k8s from '@kubernetes/client-node';
import { Store } from './store';

/**
 * Execute the callbacks for each item in the store
 * @param {string} plural - Kind to observe
 * @param {string} namespace - Namespace to observe
 * @param {Function} onAdd - Function to call when an item is added
 * @param {Function} onChange - Function to call when an item is modified
 * @param {Function} onDelete - Function to call when an item is deleted
 * @param {Function} onRename - Function to call when an item is renamed
 */
export async function observe(
  plural: string,
  namespace: string,
  onAdd: Function,
  onChange: Function,
  onDelete: Function,
  _onRename: Function,
) {
  const store = new Store(plural);

  const { kc } = await getConnection();

  try {
    const k8sApi = kc.makeApiClient(k8s.CustomObjectsApi);

    const apiGroup = 'firestartr.dev';

    const apiVersion = 'v1';

    const apiPaths = `/apis/${apiGroup}/${apiVersion}/namespaces/${namespace}/${plural}`;

    const listFn = () => {
[...]

PR #2429, opened 2025-05-09
Cherry-pick in two watch changes from the 0.x branch, clean up tests, fix list -> resourceVersion handling

brendandburns:
@cjihrig 

This is part of #2424 I don't think it's actually going to fix all of it, but it might help and it can't hurt.


cjihrig (approved):
/lgtm
/approve

Issue #1613, opened 2024-03-12
not possible to set a client timeout


cdbkr:
**Describe the bug**
At the moment, it seems not possible to set a client timeout since node-fetch requires passing an AbortController instance to the fetch request option's signal.

** Client Version **
1.0.0-rc3

**To Reproduce**

**Expected behavior**
It would be great to set a client's timeout when creating a k8sApi instance or invoking a request

mstruebing:
I would say yes but it should definitely be an optional argument and probably an object so that it's easier to extend and adapt in the future.

@brendanburns any concerns regarding this?

brendandburns:
I think it is totally fine to add an optional parameter to the `makeApiClient` signature. Thanks for checking!

elizabeth-dev:
After starting to work on this and looking a bit more into it, I've come to the conclusion that this can not be implemented right now 😕. All the code that handles the HTTP requests is generated with the library `openapi-generator-maven-plugin`, and the client is completely coupled to that. I thought there would be some way to customize the template that generates the request (at least add some extra parameters for the timeout), but it doesn't look like it.

So 90% of the functionality for this feature needs to be done on the OpenAPI generator (and I'm assuming it'd need to be done across all the languages they support).

mstruebing:
@elizabeth-dev I did not look to deep into it, but I think this needs to be implemented in: https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/typescript

(or `typescript-fetch`/`typescript-node`, where you would be able to only change the code for a specific case and do not need to implement it for all. I may be wrong here though.

brendandburns:
I think that this is the place where you would want to add the timeout code:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/runtime.mustache#L81

In general that project is pretty good about merging PRs for specific runtimes.

As I splunked through the code though, I realized you can also make the change here:

https://github.com/kubernetes-client/javascript/blob/release-1.x/src/config.ts#L470

And I'm pretty sure that it will work. That's the `Configuration` object that flows down through to fetch ultimately (if I tracked the code right)

ylnsnv:
Has there been any progress on this issue? I'm encountering the following error: `Error: Client network socket disconnected before secure TLS connection was established` when attempting to fetch deployments from within the cluster. I suspect this might be related to an insufficient timeout setting. Could this be the case?

[...]

PR #2430, opened 2025-05-12
Add optional timeoutMs parameter to makeApiClient

jannikschaper:
Fixes https://github.com/kubernetes-client/javascript/issues/1613

I followed @brendandburns comments from the linked issue:

> I think it is totally fine to add an optional typed parameter to the makeApiClient signature. Thanks for checking!

> And I'm pretty sure that it will work. That's the Configuration object that flows down through to fetch ultimately (if I tracked the code right

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -28,14 +28,18 @@ import {
     AuthMethodsConfiguration,
     Configuration,
     createConfiguration,
+    Observable,
     SecurityAuthentication,
     ServerConfiguration,
 } from './gen/index.js';
+import { ResponseContext } from './gen/http/http.js';
 import { OpenIDConnectAuth } from './oidc_auth.js';
 import WebSocket from 'isomorphic-ws';
 import child_process from 'node:child_process';
 import { SocksProxyAgent } from 'socks-proxy-agent';
 import { HttpProxyAgent, HttpProxyAgentOptions, HttpsProxyAgent, HttpsProxyAgentOptions } from 'hpagent';
+import fetch from 'node-fetch';
+import { from } from './gen/rxjsStub.js';
 
 const SERVICEACCOUNT_ROOT: string = '/var/run/secrets/kubernetes.io/serviceaccount';
 const SERVICEACCOUNT_CA_PATH: string = SERVICEACCOUNT_ROOT + '/ca.crt';
@@ -473,7 +477,12 @@ export class KubeConfig implements SecurityAuthentication {
         );
     }
 
-    public makeApiClient<T extends ApiType>(apiClientType: ApiConstructor<T>): T {
+    public makeApiClient<T extends ApiType>(
+        apiClientType: ApiConstructor<T>,
+        options?: {
+            timeoutMs?: number;
+        },
+    ): T {
         const cluster = this.getCurrentCluster();
         if (!cluster) {
             throw new Error('No active cluster!');
@@ -485,6 +494,35 @@ export class KubeConfig implements SecurityAuthentication {
         const config: Configuration = createConfiguration({
             baseServer: baseServerConfig,
             authMethods: authConfig,
+            // This HTTP API implementation is a copy of ./gen/http/isomorphic-fetch.ts,
+            // extended with the timeout pass-through code.
+            httpApi: {
+                send(request: RequestContext): Observable<ResponseContext> {
+                    const method = request.getHttpMethod().toString();
+                    const body = request.getBody();
+
+                    const resultPromise = fetch(request.getUrl(), {
+                        method: method,
+                        body: body as any,
+                        headers: request.getHeaders(),
+                        agent: request.getAgent(),
+                        timeout: options?.timeoutMs,
+                    }).then((resp: any) => {
+                        const headers: { [name: string]: string } = {};
+                        resp.headers.forEach((value: string, name: string) => {
+                            headers[name] = value;
+                        });
+
+                        const body = {
+                            text: () => resp.text(),
+                            binary: () => resp.buffer(),
+                        };
+                        return new ResponseContext(resp.status, headers, body);
+                    });
+
+                    return from<Promise<ResponseContext>>(resultPromise);
+                },
+            },
         });
 
         const apiClient = new apiClientType(config);


src/config.ts
```diff
@@ -485,6 +494,35 @@ export class KubeConfig implements SecurityAuthentication {
...
+            // This HTTP API implementation is a copy of ./gen/http/isomorphic-fetch.ts,
+            // extended with the timeout pass-through code.
+            httpApi: {
+                send(request: RequestContext): Observable<ResponseContext> {
+                    const method = request.getHttpMethod().toString();
+                    const body = request.getBody();
+
+                    const resultPromise = fetch(request.getUrl(), {
+                        method: method,
+                        body: body as any,
```
mstruebing, line 506:
```suggestion
                        body: body as BodyInit,
```

If we do this typecast we should at least cast into the correct type.

src/config.ts
```diff
@@ -485,6 +494,35 @@ export class KubeConfig implements SecurityAuthentication {
...
+                        resp.headers.forEach((value: string, name: string) => {
+                            headers[name] = value;
+                        });
+
+                        const body = {
+                            text: () => resp.text(),
+                            binary: () => resp.buffer(),
+                        };
+                        return new ResponseContext(resp.status, headers, body);
+                    });
```
mstruebing, line 521: Not 100% sure but do we need a `catch` here as well or will this be catched by the client code?

src/config.ts
```diff
@@ -485,6 +494,35 @@ export class KubeConfig implements SecurityAuthentication {
         const config: Configuration = createConfiguration({
             baseServer: baseServerConfig,
             authMethods: authConfig,
+            // This HTTP API implementation is a copy of ./gen/http/isomorphic-fetch.ts,
+            // extended with the timeout pass-through code.
+            httpApi: {
+                send(request: RequestContext): Observable<ResponseContext> {
```
brendandburns, line 500: I don't love pulling this full implementation into this library. Is there an easier way to inject the timeout?

src/config.ts
jannikschaper: Could do that. Would lose the pure symmetry with the original code then. If you're aware that I deliberately copied the `send` implementation verbatim, trying to stay close to the original code, but still prefer to have the more correct type cast, I can do that!

src/config.ts
jannikschaper: I thought that too, it would be much nicer to inject the extra field without having to rip out the entire function. But I believe there's no other way. The auto-generated code passes the configuration object (without any `timeout` value, not even passed through) directly into `fetch`. Not even a dependency-injected `fetch` or anything that might be overwriteable, but a `fetch` directly imported from `node-fetch`.

Maybe it's still possible to reach into the code via JavaScript black magic to overwrite the `fetch` implementation and amend the `timeout` parameter, but I have a hunch such code will be even uglier than copying the 22 lines of code

jannikschaper: I wanted to avoid that seeming rabbit hole. But maybe it is a better solution, I will look into it

davidgamero:
i'm testing out a PR on the generator for passing through the signal property on the `RequestContext` type

middleware has access to the RequestContext which would allow adding a middleware that sets the abortsignal

another alternative is to add a `signal` input to the Configuration Type to expose it directly in the options. this would be a more accessible call, but it would add more complexity

https://github.com/OpenAPITools/openapi-generator/pull/21323

jannikschaper:
It works!

```ts
const makeApiClient = <T extends ApiType>(
    kubeConfig: KubeConfig,
    ApiClientType: ApiConstructor<T>,
    requestTimeoutInSeconds: number,
): T => {
    // Copied from KubeConfig.makeApiClient and adapted to add timeout
    // Suggested by @kubernetes/client-node maintainer (https://github.com/kubernetes-client/javascript/pull/2430#issuecomment-2918109624)

    const cluster = kubeConfig.getCurrentCluster();
    if (!cluster) {
        throw new Error('No active cluster!');
    }

    return new ApiClientType(createConfiguration({
        baseServer: new ServerConfiguration(cluster.server, {}),
        authMethods: {
            default: kubeConfig,
        },
        middleware: [{
            pre: (context) => {
                const controller = new AbortController();
                setTimeout(() => controller.abort(), requestTimeoutInSeconds * 1000);
                context.setSignal(controller.signal);

                return new Observable(Promise.resolve(context));
            },
            post: (context) => new Observable(Promise.resolve(context)),
        }],
    }));
};
```

Thank you so much @davidgamero for going out of your way and getting PRs merged both here and even in OpenAPI to fix my use case. Using a signal is better than a hardcoded timeout value, because you can implement a timeout and more on top of it, good idea. Thanks!

jannikschaper:
@davidgamero I have a follow-up question to the timeout approach you suggested, if you don't mind

How to clean up dangling timeout instances when requests complete normally? For example, with a 120s timeout and on average 10 successful Kubernetes API requests per second, there would be over a thousand timeout promises still hanging around in the NodeJS event loop.

Each of them also eventually calling `.abort()` on the request, which is additional mental load to make sure these delayed false-positive aborts don't cause troubles.

A naive implementation might look like this:

```ts
function timeoutMiddleware(milliseconds: number): Middleware {
    let timeoutId: NodeJS.Timeout;

    return {
        pre: async (c: RequestContext) => {
            const abortController = new AbortController();
            timeoutId = setTimeout(() => abortController.abort(), milliseconds);
            c.setSignal(abortController.signal);

            return c;
        },
        post: async (c: ResponseContext) => {
            clearTimeout(timeoutId);

            return c;
        },
    };
}
```

But this is wrong, because parallel requests would overwrite each other's `timeoutId`. You'd need to attach the `timeoutId` to the request context, but as far as I can tell, there is no way to inject a value into a `RequestContext` and get it out reliably in the `ResponseContext`. The only possibility I can think of is to inject a cookie with a timeout ID in the request and hope that the Kubernetes API server returns the cookie in its response. But this is obviously a terrible solution.

Do you have an idea how to solve the problem of dangling timeouts with this library? Is there simply no way, and the best solution is to just deal with dangling timeouts aborting each not-anymore-existent request long after it finished?

Thank you for your help!

DrauzJu:
@jannikschaper We have found a solution for that:

```ts
promiseMiddleware: [{
    pre: async (context) => {
        context.setSignal(AbortSignal.timeout(requestTimeoutInSeconds * 1000));

        return context;
    },
    post: async (context) => context,
}],
```

This way, we don't have to cleanup the timeout ourself because NodeJs internally calls [`Timeout.unref()`](https://nodejs.org/docs/latest-v20.x/api/timers.html#timeoutunref) to prevent the process from exiting if there is still such a pending timeout: https://github.com/nodejs/node/blob/3a33ee3cf09a36f6c2456b053efa8d7f20728956/lib/internal/abort_controller.js#L139-L161

PR #2442, opened 2025-05-19
chore: readd ws as a dependency

mstruebing:
It seems like people having issues due to the removal of ws https://github.com/kubernetes-client/javascript/commit/b727c7be7fc7c7a8cb5663d6d621d92027d47fc1\#r156475449

It would be good if there would be another way than to reintroduce the dependency but I don't really know how and at the same time I don't think it's good to have every consumer install ws separately.


cjihrig (approved):
Seems easier than trying to rely on `peerDependencies`.

/lgtm
/approve

PR #2520, opened 2025-06-25
fix: make yaml functions use serializer

### LINKED ISSUE #2483: CustomResourceDefinition Serialization | x-kubernetes-int-or-string Dropped
**Describe the bug**

When trying to create a CRD with `x-kubernetes-int-or-string`, its being dropped causing the server to respond with `The CustomResourceDefinition "middlewares.traefik.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[circuitBreaker].properties[checkPeriod].type: Required value: must not be empty for specified object fields`.

<details>
  <summary>Original CRD</summary>

[original.yaml](https://github.com/user-attachments/files/20659589/original.txt)

Image

</details>

<details>
  <summary>Broken CRD</summary>

[broken.yaml](https://github.com/user-attachments/files/20659591/broken.txt)

This is what is actually being passed to the server. How do I know? I print it out before sending the request in the underlying `node_module`, `node_modules/@kubernetes/client-node/dist/object.js`

Image

Image

</details>

**Client Version**
`1.0.0` to `1.3.0`

**Server Version**
`1.30.0` to `1.33.0`

**To Reproduce**

##### First option

1. Use example code

##### Second option

1. kubectl apply -f [USE HERE BROKEN CRD YAML]

Image

^^ These are the errors you get while using the first option

**Expected behavior**

The `create` call should not strip out `x-kubernetes-int-or-string`. In essence, it should replicate the behaviour of `kubectl apply -f [USE HERE ORIGINAL CRD YAML]` which successfully creates the resource

**Example Code**
```ts
import yaml from 'yaml';

kc.loadFromDefault();

const client = k8s.KubernetesObjectApi.makeApiClient(kc);
const spec = yaml.parse(USE HERE ORIGINAL CRD YAML).toJS()

await client.create(spec);
```

- OS: Linux
- Node.js version: v20.14.0

**Additional context**

feloy:
Fixes #2483 

This PR changes the implementation of the `yaml`-related functions, to make them serialize/deserialize the objects.

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -15,6 +15,7 @@ import { ObjectSerializer } from './serializer.js';
 import { KubernetesListObject, KubernetesObject } from './types.js';
 import { from, mergeMap, of } from './gen/rxjsStub.js';
 import { PatchStrategy } from './patch.js';
+import { getSerializationType } from './util.js';
 
 /** Kubernetes API verbs. */
 type KubernetesApiAction = 'create' | 'delete' | 'patch' | 'read' | 'list' | 'replace';
@@ -28,11 +29,6 @@ type KubernetesObjectHeader<T extends KubernetesObject | KubernetesObject> = Pic
         namespace?: string;
     };
 };
-
-interface GroupVersion {
-    group: string;
-    version: string;
-}
 
 /**
  * Dynamically construct Kubernetes API request URIs so client does not have to know what type of object it is acting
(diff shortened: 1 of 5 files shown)


brendandburns:
/lgtm
/approve

Thanks for the PR, and thanks for adding tests!

PR #2548, opened 2025-07-08
fix: support YAML 1.1 parsing in Kubernetes manifests by replacing js-yaml with yaml

### LINKED ISSUE #2539: YAML 1.2 only is supported
**Describe the bug**

Library uses a YAML parser compatible with YAML 1.2 only, when Kubernetes is mostly compatible with YAML 1.1 (https://kubernetes.io/docs/concepts/configuration/overview/#:~:text=YAML%201.2%20only%20recognizes%20true,cause%20unexpected%20errors%20or%20behaviors.) 

**Client Version**

main

**Server Version**

1.31

**To Reproduce**

See example code below

**Expected behavior**

The value `0644` in the manifest is correctly recognized a an octal value with Kubernetes tools, but not with the function `loadYaml` from this library

**Example Code**

```
import { EventType, State } from 'js-yaml';

const manifest = `
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: contour
  name: contour
  namespace: projectcontour
spec:
  replicas: 2
  strategy:
    type: RollingUpdate
    rollingUpdate:
      # This value of maxSurge means that during a rolling update
      # the new ReplicaSet will be created first.
      maxSurge: 50%
  selector:
    matchLabels:
      app: contour
  template:
    metadata:
      labels:
        app: contour
    spec:
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - podAffinityTerm:
              labelSelector:
                matchLabels:
                  app: contour
              topologyKey: kubernetes.io/hostname
            weight: 100
      containers:
      - args:
        - serve
        - --incluster
        - --xds-address=0.0.0.0
        - --xds-port=8001
        - --contour-cafile=/certs/ca.crt
        - --contour-cert-file=/certs/tls.crt
        - --contour-key-file=/certs/tls.key
        - --config-path=/config/contour.yaml
        command: ["contour"]
        image: ghcr.io/projectcontour/contour:v1.30.2
        imagePullPolicy: IfNotPresent
        name: contour
        ports:
        - containerPort: 8001
          name: xds
          protocol: TCP
        - containerPort: 8000
          name: metrics
          protocol: TCP
        - containerPort: 6060
          name: debug
          protocol: TCP
        livenessProbe:
          httpGet:
            path: /healthz
            port: 8000
        readinessProbe:
          tcpSocket:
            port: 8001
          periodSeconds: 10
        volumeMounts:
          - name: contourcert
            mountPath: /certs
            readOnly: true
          - name: contour-config
            mountPath: /config
            readOnly: true
        env:
        - name: CONTOUR_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
      dnsPolicy: ClusterFirst
      serviceAccountName: contour
      securityContext:
        runAsNonRoot: true
        runAsUser: 65534
        runAsGroup: 65534
      volumes:
        - name: contourcert
          secret:
            secretName: contourcert
        - name: contour-config
          configMap:
            name: contour
            defaultMode: 0644
            items:
            - key: contour.yaml
              path: contour.yaml
`;

class MyKubernetesObjectApi extends k8s.KubernetesObjectApi {
  public async getSerializationType(apiVersion?: string, kind?: string): Promise<string> {
    return super.getSerializationType(apiVersion, kind);
  }
}

kc.loadFromDefault();

const client = MyKubernetesObjectApi.makeApiClient(kc) as MyKubernetesObjectApi;

const yml = k8s.loadYaml<k8s.KubernetesObject>(manifest);

client.create(yml).then(() => {
  console.log('==> created');
}).catch((err: unknown) => {
  console.error(err);
});
```

- OS: Mac
- Node.js version 22.16

**Additional context**

Related issue on `js-yaml` library: https://github.com/nodeca/js-yaml/issues/684

Tanmayshi:
🛠️ What’s Changed

Removed: js-yaml dependency

Added: yaml library

Updated manifest parsing logic to:

parseDocument(manifest, { version: '1.1' });
Ensures values are interpreted using YAML 1.1 rules.

🐛 Problems Solved

1. ✅ Incorrect parsing of octal values
defaultMode: 0644 was being parsed as 644 in YAML 1.2 (incorrect)

In YAML 1.1, it's correctly interpreted as octal 0644 → decimal 420

This fixes configMap and secret mount permission issues in Kubernetes

2. ✅ Boolean string interpretation issues
YAML 1.2 doesn't treat yes, no, on, off as booleans

In YAML 1.1, these are properly recognized as true/false

Prevents unexpected behavior in manifest logic that depends on boolean flags

### ✅ Tested
- Applied a manifest with `defaultMode: 0644` — parsed correctly as 420
- Applied a manifest with `on`, `yes`, `no` — parsed correctly as booleans
- Resource created successfully using KubernetesObjectApi

✅ Result
Manifests parsed through this library now match Kubernetes expectations

No more silent failures or misconfigurations due to YAML parsing issues
Screenshot (312)

Fixes #2539

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
                 "form-data": "^4.0.0",
                 "hpagent": "^1.2.0",
                 "isomorphic-ws": "^5.0.0",
-                "js-yaml": "^4.1.0",
                 "jsonpath-plus": "^10.3.0",
                 "node-fetch": "^2.6.9",
                 "openid-client": "^6.1.3",
@@ -24,7 +23,8 @@
                 "socks-proxy-agent": "^8.0.4",
                 "stream-buffers": "^3.0.2",
                 "tar-fs": "^3.0.9",
-                "ws": "^8.18.2"
+                "ws": "^8.18.2",
+                "yaml": "^2.8.0"
             },
             "devDependencies": {
                 "@eslint/js": "^9.18.0",
@@ -1349,6 +1349,7 @@
             "version": "2.0.1",
             "resolved": "https://registry.npmjs.org/argparse/-/argparse-2.0.1.tgz",
             "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
+            "dev": true,
             "license": "Python-2.0"
         },
         "node_modules/asynckit": {
@@ -2659,6 +2660,7 @@
             "version": "4.1.0",
             "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
             "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+            "dev": true,
             "license": "MIT",
             "dependencies": {
                 "argparse": "^2.0.1"
@@ -4022,7 +4024,7 @@
             "version": "2.8.0",
             "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.8.0.tgz",
             "integrity": "sha512-4lLa/EcQCB0cJkyts+FpIRx5G/llPxfP6VQU5KByHEhLxY3IJCH0f0Hy1MHI8sClTvsIb8qwRJ6R/ZdlDJ/leQ==",
-            "dev": true,
+            "license": "ISC",
             "bin": {
                 "yaml": "bin.mjs"
             },
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
         "form-data": "^4.0.0",
         "hpagent": "^1.2.0",
         "isomorphic-ws": "^5.0.0",
-        "js-yaml": "^4.1.0",
         "jsonpath-plus": "^10.3.0",
         "node-fetch": "^2.6.9",
         "openid-client": "^6.1.3",
@@ -70,7 +69,8 @@
         "socks-proxy-agent": "^8.0.4",
         "stream-buffers": "^3.0.2",
         "tar-fs": "^3.0.9",
-        "ws": "^8.18.2"
+        "ws": "^8.18.2",
+        "yaml": "^2.8.0"
     },
     "devDependencies": {
         "@eslint/js": "^9.18.0",
--- a/src/yaml.ts
+++ b/src/yaml.ts
@@ -1,4 +1,4 @@
-import yaml from 'js-yaml';
+import YAML from 'yaml';
 import { getSerializationType } from './util.js';
 import { KubernetesObject } from './types.js';
 import { ObjectSerializer } from './serializer.js';
@@ -9,13 +9,12 @@ import { ObjectSerializer } from './serializer.js';
  * @param opts - Optional YAML load options.
  * @returns The deserialized Kubernetes object.
  */
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    const yml = yaml.load(data, opts) as any as KubernetesObject;
+export function loadYaml<T>(data: string): T {
+    const yml = YAML.parse(data, { version: '1.1' }) as any as KubernetesObject;
     if (!yml) {
         throw new Error('Failed to load YAML');
     }
     const type = getSerializationType(yml.apiVersion, yml.kind);
-
     return ObjectSerializer.deserialize(yml, type) as T;
 }
 
@@ -25,12 +24,12 @@ export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
  * @param opts - Optional YAML load options.
  * @returns An array of deserialized Kubernetes objects.
  */
-export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
-    const ymls = yaml.loadAll(data, undefined, opts);
-    return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+export function loadAllYaml(data: string): any[] {
+    const ymls = YAML.parseAllDocuments(data, { version: '1.1' });
+    return ymls.map((doc) => {
+        const obj = doc.toJS() as KubernetesObject;
         const type = getSerializationType(obj.apiVersion, obj.kind);
-        return ObjectSerializer.deserialize(yml, type);
+        return ObjectSerializer.deserialize(obj, type);
     });
 }
 
@@ -40,9 +39,9 @@ export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
  * @param opts - Optional YAML dump options.
  * @returns The YAML string representation of the serialized Kubernetes object.
  */
-export function dumpYaml(object: any, opts?: yaml.DumpOptions): string {
+export function dumpYaml(object: any): string {
     const kubeObject = object as KubernetesObject;
     const type = getSerializationType(kubeObject.apiVersion, kubeObject.kind);
     const serialized = ObjectSerializer.serialize(kubeObject, type);
-    return yaml.dump(serialized, opts);
+    return YAML.stringify(serialized);
 }
--- a/src/yaml_test.ts
+++ b/src/yaml_test.ts
@@ -154,4 +154,39 @@ spec:
         // not using strict equality as types are not matching
         deepEqual(actual, expected);
     });
+
+    it('should parse octal-like strings as numbers (YAML 1.1 style)', () => {
+        const yaml = `
+    defaultMode: 0644
+    fileMode: 0755
+    `;
+        const result = loadYaml<{
+            defaultMode: number;
+            fileMode: number;
+        }>(yaml);
+
+        // 0644 (octal) = 420 decimal, 0755 = 493
+        strictEqual(result.defaultMode, 420);
+        strictEqual(result.fileMode, 493);
+    });
+
+    it('should parse boolean-like strings as booleans (YAML 1.1 style)', () => {
+        const yaml = `
+    enableFeature: yes
+    debugMode: ON
+    maintenance: no
+    safeMode: off
+    `;
+        const result = loadYaml<{
+            enableFeature: boolean;
+            debugMode: boolean;
+            maintenance: boolean;
+            safeMode: boolean;
+        }>(yaml);
+
+        strictEqual(result.enableFeature, true);
+        strictEqual(result.debugMode, true);
+        strictEqual(result.maintenance, false);
+        strictEqual(result.safeMode, false);
+    });
 });


package.json
```diff
@@ -62,15 +62,15 @@
         "form-data": "^4.0.0",
         "hpagent": "^1.2.0",
         "isomorphic-ws": "^5.0.0",
-        "js-yaml": "^4.1.0",
```
cjihrig, line 65: `@types/js-yaml` should also be removed. Is there a `@types/yaml` that should be added in its place?

src/yaml.ts
```diff
@@ -1,4 +1,4 @@
-import yaml from 'js-yaml';
+import YAML from 'yaml';
```
cjihrig: Can you leave this name as it was please.

src/yaml.ts
```diff
@@ -25,12 +24,12 @@ export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
...
  * @returns An array of deserialized Kubernetes objects.
  */
-export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
-    const ymls = yaml.loadAll(data, undefined, opts);
-    return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+export function loadAllYaml(data: string): any[] {
+    const ymls = YAML.parseAllDocuments(data, { version: '1.1' });
+    return ymls.map((doc) => {
+        const obj = doc.toJS() as KubernetesObject;
```
cjihrig: Can you keep the name as `obj` to keep the diff smaller.

cjihrig (commented):
Can you please sign the CLA.

src/yaml.ts
```diff
@@ -25,12 +24,12 @@ export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
  * @param opts - Optional YAML load options.
  * @returns An array of deserialized Kubernetes objects.
  */
-export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
```
feloy, line 28: I think users may still want to pass options to the parser

src/yaml.ts
```diff
@@ -25,12 +24,12 @@ export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
...
-    const ymls = yaml.loadAll(data, undefined, opts);
-    return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+export function loadAllYaml(data: string): any[] {
+    const ymls = YAML.parseAllDocuments(data, { version: '1.1' });
```
feloy: I think this value `1.1` for `version` option should be default, but the user may be able to overwrite it

package.json
Tanmayshi: ok

Tanmay-008: Yes, I will remove @types/js-yaml since it's no longer used.

Regarding @types/yaml:
I won't be adding it because the yaml package already includes its own TypeScript definitions, so there's no need for a separate type package.

If you still think adding it is necessary for a specific reason, please let me know — happy to update accordingly.

src/yaml.ts
Tanmay-008: Yes sir  @feloy, you're absolutely right 

I've updated the function so that `version: '1.1'` is now used as the default, but users can still override it by passing their own options.

yaml.parseAllDocuments(data, { version: '1.1', ...opts }); ,

package-lock.json
```diff
@@ -16,15 +16,15 @@
                 "form-data": "^4.0.0",
                 "hpagent": "^1.2.0",
                 "isomorphic-ws": "^5.0.0",
-                "js-yaml": "^4.1.0",
```
cjihrig, line 19: I don't think `@types/js-yaml` has been removed in this file.

src/yaml.ts
```diff
@@ -1,48 +1,47 @@
...
+ * @param opts - Optional YAML parse options.
  * @returns The deserialized Kubernetes object.
  */
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    const yml = yaml.load(data, opts) as any as KubernetesObject;
+export function loadYaml<T>(data: string, opts?: yaml.ParseOptions): T {
+    const yml = yaml.parse(data, { version: '1.1', ...opts }) as any as KubernetesObject;
     if (!yml) {
-        throw new Error('Failed to load YAML');
+        throw new Error('Failed to load yaml');
```
cjihrig, line 15: Please undo this change.

src/yaml.ts
```diff
@@ -1,48 +1,47 @@
...
-    const ymls = yaml.loadAll(data, undefined, opts);
-    return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+export function loadAllYaml(data: string, opts?: yaml.ParseOptions): KubernetesObject[] {
+    const ymls = yaml.parseAllDocuments(data, { version: '1.1', ...opts });
+    return ymls.map((doc) => {
+        const obj = doc.toJSON() as KubernetesObject;
         const type = getSerializationType(obj.apiVersion, obj.kind);
-        return ObjectSerializer.deserialize(yml, type);
+        return ObjectSerializer.deserialize(obj, type);
```
cjihrig, line 35: If you don't rename things and make unrelated changes, the diff stays smaller and easier to review. It also makes `git blame` easier to use too.

cjihrig:
I started the CI. It looks like there are relevant failures.

src/yaml_test.ts
```diff
@@ -84,22 +84,27 @@ spec:
...
         strictEqual(objects.length, 3);
-        strictEqual(objects[0].kind, 'Namespace');
-        strictEqual(objects[1].kind, 'Pod');
-        strictEqual(objects[0].metadata.name, 'some-namespace');
-        strictEqual(objects[1].metadata.name, 'some-pod');
-        strictEqual(objects[1].metadata.namespace, 'some-ns');
-        strictEqual(objects[2].apiVersion, 'apiextensions.k8s.io/v1');
-        strictEqual(objects[2].kind, 'CustomResourceDefinition');
-        strictEqual(objects[2].metadata!.name, 'my-crd.example.com');
+        // Assert specific types for each object
```
cjihrig: The fact that an existing test broke implies to me that this may be a subtly breaking change.

src/yaml.ts
```diff
@@ -1,48 +1,47 @@
...
- * Load a Kubernetes object from YAML.
+ * Load a single Kubernetes object from YAML.
  * @param data - The YAML string to load.
- * @param opts - Optional YAML load options.
+ * @param opts - Optional YAML parse options.
  * @returns The deserialized Kubernetes object.
  */
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    const yml = yaml.load(data, opts) as any as KubernetesObject;
+export function loadYaml<T>(data: string, opts?: yaml.ParseOptions): T {
```
feloy: `yaml.parse` accepts `options` of type `ParseOptions & DocumentOptions & SchemaOptions & ToJSOptions`. Let's the user have access to all the options, not only `ParseOptions` ones (especially the `version` one, which is in the `DocumentOptions`)

src/yaml.ts
```diff
@@ -1,48 +1,47 @@
...
  * @param data - The YAML string to load.
- * @param opts - Optional YAML load options.
+ * @param opts - Optional YAML parse options.
  * @returns An array of deserialized Kubernetes objects.
  */
-export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
-    const ymls = yaml.loadAll(data, undefined, opts);
-    return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+export function loadAllYaml(data: string, opts?: yaml.ParseOptions): KubernetesObject[] {
```
feloy: type of opts would be `ParseOptions & DocumentOptions & SchemaOptions` here

feloy: I'm not sure we want to change the type of the return value, as it will be a breaking change for user (and is what makes the tests fail)

src/yaml_test.ts
Tanmay-008: The tests themselves were not failing — the issue was purely related to TypeScript type-checking. TypeScript was warning that some properties might be undefined or missing. I've addressed those issues, and CI is now passing 

If you'd prefer a more strictly type-safe solution, please let me knowd.

src/yaml.ts
Tanmay-008: Thanks! Updated the type to ParseOptions & DocumentOptions so version: '1.1' and similar options work correctly now. Let me know if more should be added.

src/yaml.ts
Tanmay-008: i do it

mstruebing: We should not. We should try to keep backwards compatibility as much as possible.

Tanmay-008: We initially used KubernetesObject[] for better type safety and IntelliSense — especially when accessing standard fields like apiVersion, kind, or metadata.

But you're right — changing it from any[] would be a breaking change for users. So we’ve reverted back to any[] to preserve backward compatibility.

src/yaml.ts
feloy: I cannot be sure some users won't have an interest on using options from the other ones, I would set all of them: `ParseOptions & DocumentOptions & SchemaOptions & ToJSOptions`

src/yaml.ts
cjihrig: I think we may need to define our own types for the options passed to `loadYaml()`, `loadAllYaml()`, and `dumpYaml()`. These are part of the public API, and changing the underlying YAML library like this changes the actual options that are supported.

cjihrig:
@Tanmayshi there are still a few pending comments here in case you missed them.

Tanmay-008:
Hello @cjihrig 
I’ve carefully reviewed all the comments again and made sure that every single one has been addressed — either through code updates or direct replies. As far as I can see, there are no remaining suggestions that are unhandled.
If you still feel something is pending, could you please point out which specific comment or section I might’ve missed? I’ll be happy to fix it right away.

cjihrig:
Please take a look [here](https://github.com/kubernetes-client/javascript/pull/2548/files#diff-ddb39cb67821bb3b65c196c0721385fda0fd8ffa8d0f2b6a345d8f965cf24332). I count three outstanding threads. Two are very minor fixes, but one of them could require a bit of work.

Tanmay-008:
hello @cjihrig  can you gave me more details

brendandburns:
@Tanmayshi I believe you need to address the following comments:
* https://github.com/kubernetes-client/javascript/pull/2548/files#r2193043462
* https://github.com/kubernetes-client/javascript/pull/2548/files#r2195300625
* https://github.com/kubernetes-client/javascript/pull/2548/files#r2193046225

Tanmay-008:
@brendandburns Yes sir , I have paid attention to these comments, but for the second comment I needed some help. Could you please explain what needs to be done there?

brendandburns:
In the second comment, on that line you changed `yml` to `obj`, @cjihrig is requesting that you not change variable names when it is not necessary. Adding such changes that don't have impact make it harder to read and review code, so please revert that particular line changes.

Tanmay-008:
Yes sir @brendandburns , I have corrected this mistake but @cjihrig  say to write your own types ( define our own types for the options passed to loadYaml(), loadAllYaml(), and dumpYaml(). ) I am not able to understand how should I do this, can you tell me something in this.

brendandburns:
Ah, for that comment I believe that @cjihrig is suggesting that instead of using `yaml.Foo` types as parameters to `loadYaml` we define our own type in `yaml.ts` such that we are insulated from future changes in the `yaml` libraries types.

Basically define our own `LoadOptions` or something like that, and then map from that type to the `yaml.Foo` types.

mstruebing:
@Tanmayshi any news on this? Will you continue working on this PR?

Tanmay-008:
“Hi @mstruebing , I’m still working on this PR and need a little more time. I’ll update it soon. Thanks!”

Tanmay-008:
i well create a new pr for this

Issue #2573, opened 2025-07-28
End of life for v0.x release line


cjihrig:
We haven't cut a release from the 0.x release line since last November, and the only changes that have been landing on that branch are dependabot updates (which are now starting to fail). The 1.x release line has finally taken over in terms of npm downloads, and Node core has recently evolved such that ESM-only makes a lot more sense.

I propose discontinuing the 0.x release line, effective immediately. I volunteer to do the work of removing the remaining 0.x related code in the repo.

brendandburns:
Agreed.

jeevanroy1434:
https://api.etherscan.io/v2/api
   ?chainid=1
   &module=proxy
   &action=eth_getTransactionCount
   &address=0x4bd5900Cb274ef15b153066D736bf3e83A9ba44e
   &tag=latest
   &apikey=YourApiKeyToken

PR #2576, opened 2025-07-29
remove support for v0.x release line

cjihrig:
This commit disables everything related to the old v0.x release line (the master branch in git). The v0.x git branch and old release artifacts will be left in tact.

Fixes: https://github.com/kubernetes-client/javascript/issues/2573

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,29 +1,29 @@
 name: Build and Deploy Docs
 on:
-  push:
-    branches:
-      - master
+    push:
+        branches:
+            - main
 jobs:
-  build-and-deploy-docs:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4.2.2
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '20'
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -1,29 +1,29 @@
 name: Build and Deploy Docs
 on:
-  push:
-    branches:
-      - master
+    push:
```
cjihrig, line 3: The whitespace changes in this file were done automatically by my editor.

brendandburns:
Let's leave `codeql` and `test` so that if we get PRs we don't accidentally merge broken, malicious code without CI/CD.

It will only run if people send PRs.

mstruebing:
Nice LGTM to me but I will leave the final approval to @brendandburns.

Should we update the readme in the `master` branch with a big note that this is a deprecated branch as well?

cjihrig:
> Should we update the readme in the master branch with a big note that this is a deprecated branch as well?

No strong opinion on that, but it would have to be done as a separate PR.

brendandburns:
/lgtm
/approve

Thanks for cleaning this up. I'm not sure that we need to modify the `master` branch. We've already switched the default branch to `main` so in my opinion `master` is just another branch like any other.

But if someone wants to send a PR, I'm happy to merge it.

Issue #2582, opened 2025-08-01
New Release Needed.


Venkatesh-j-koppar:
Hi maintainers,

Thank you for your work on this project!

I’m wondering if there’s a release planned in the coming days that would include [PR #2514](https://github.com/kubernetes-client/javascript/pull/2514). I’ve been waiting for this fix and would really appreciate if it could be part of the next tagged version — even a patch release would be great.

Thanks again.

cjihrig:
+1 to cutting a new release, it's been a couple months. I can do one early next week if no one beats me to it. I don't really want to cut a new release on a Friday in case it creates unexpected work for myself or others.

brendandburns:
We will do a new 1.4.x release when the new Kubernetes version (1.34.x) releases (estimated 8/25/25).

Alternately, if you send a couple PRs to the `release-1.3` branch:
* cherry-pick the commits from #2514 into the `release-1.3` branch
* update the `package.json` and `package-lock.json` to be version `1.3.1`

We could do a patch release, which might be slightly faster than waiting for the k8s release.

brendandburns:
@cjihrig 

fwiw I think we need to discuss our versioning policy. In the Java client we chose to increase the major version whenever we move to a new Kubernetes release. In this library we have sort-of used the minor version for a new API, but we've also rev'd the minor version when we've made significant client changes.

I propose that we do the following:
* Major version increment aligned with k8s version increment (but not the version #) so after 1.34 releases, we would rev to `2.0.0` instead of `1.4.0`
* Minor versions used for standard periodic updates from `main` (we should maybe automate this?)
* Patch versions (used sparingly for cherry-picks of security and bug fixes into older versions of the client, standard 3 version support window)

Thoughts?

cjihrig:
I'm not opposed to that proposal. The only thing I'd point out is that most of the JavaScript ecosystem has settled on SemVer for versioning. Using a different versioning scheme could lead to confusion for users.

brendandburns:
My intent was that we should follow semvar as well, I'm going to rephrase to be precise:

* Major version increments aligned with k8s version increment. K8S version changes are breaking API changes, we will release from `main` when we perform this release, every ~3-4 months.
* Minor version increments when there are significant changes in `main` that someone wants in the current active version via cherry-pick into the currently released branch `release-X` eg. `release-2` these cherry picks are _only_ applied to the latest live major version, not into previous versions.
[...]

PR #2583, opened 2025-08-03
abort signal support

davidgamero:
- update openapigenerator commit
- regenerate with added abort signal support


davidgamero:
> Can you remove this part from the `tsconfig-with-tests.json` and `tsconfig.json` files?
> 
> ```js
>         "paths": {
>             "node-fetch/externals": ["./node_modules/@types/node-fetch/externals.d.ts"]
>         }
> ```
> 
> https://github.com/kubernetes-client/javascript/blob/main/tsconfig-with-tests.json#L4-L8 https://github.com/kubernetes-client/javascript/blob/main/tsconfig.json#L20-L22
> 
> I think it should work without now.

yes done

mstruebing:
/lgtm
/approve

Thanks 🚀

jannikschaper:
Can we expect a 1.4.0 release soon with this new feature included? Would be useful to us (no pressure though!) :)

mstruebing:
We want to release a new version with the new kubernetes update, maybe even v2.0.0 based on the discussion here: https://github.com/kubernetes-client/javascript/issues/2582

I've tried to prepare a release and regenerate the client with release-1.34 as it's finally there but it failed as there is nothing to commit here: https://github.com/kubernetes-client/javascript/actions/runs/17399794124

@davidgamero did you already regenerate the client with release-1.34?
I've looked up the dates and it doesn't really add up, it looks like 1.34 wasn't out there when you've regenerate but just asking to confirm.

Issue #2189, opened 2025-01-31
Type Error. ObjectSerializer expects a date but I want to pass it a string


BenSaxon:
**Describe the bug**
Type Error. ObjectSerializer expects a date but gets a string.

```
TypeError: data.toISOString is not a function
    at ObjectSerializer.serialize (file:///Users/bensaxon/Documents/code/monorepo/src/private/console/node_modules/.pnpm/@kubernetes+client-node@1.0.0/node_modules/@kubernetes/client-node/dist/gen/models/ObjectSerializer.js:2024:29)
```

The issue is happening [here](https://github.com/kubernetes-client/javascript/blob/main/src/gen/models/ObjectSerializer.ts#L2031).

The code expects `data` to be a `Date` object with a `toISOString` method.

Instead, when I log the result, I actually get a string, already in ISO format, such as:

`2025-01-30T09:58:20.000Z`

This is because the resource that I pass into the client has a `metadata.creationTimestamp` field that is a string. I would rather not have to convert it into a Date. Since I receive it as a string from API requests. I like it as a string. I know it's "supposed" to be a Date according to the type, but it's just impractical. I think it's a bad developer experience. Also, this error does not occur in v0.

**Client Version**
`1.0.0`

**To Reproduce**
Steps to reproduce the behaviour:

The error is thrown when trying to make a PUT request.
Ensure that `resource.metadata.creationTimestamp` is a string, not a date object.

Make a Patch request:

```
    const response = await client.patch(
      resource as KubernetesObject,
      undefined,
      undefined,
      undefined,
      undefined,
      PatchStrategy.MergePatch
    );
```

**Expected behaviour**

I expect the error not to throw. I suggest putting a conditional inside the [serialize](https://github.com/kubernetes-client/javascript/blob/main/src/gen/models/ObjectSerializer.ts#L2031) function that checks if `data` is of type string before assuming it's a `Date` object. If `data` is a string, then return it as is, else return `data.toISOstring()`.

brendandburns:
The decision about how to handle this is part of the code generator, specifically the files here:

https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/resources/typescript-fetch/modelGeneric.mustache


We rely on code generation because the surface area of the Kubernetes API is too large to have hand-written code maintained by a small number of maintainers.

[...]

PR #2599, opened 2025-08-13
regenerate client

mstruebing:
This updates the openapi generator regarding: https://github.com/kubernetes-client/javascript/issues/2189

~~as mentioned here: https://github.com/kubernetes-client/javascript/issues/2189#issuecomment-3182734858~~

~~There is an issue with the AbortSignal, I know it's not needed anymore in newer node versions but that means we need to adjust the generator first if we would want to remove it (or, but that would be a bad solution, patch the output of the generator in the kubernetes-client/javascript repository).~~

~~Maybe someone else knows how to fix it?
If not, I will go ahead and adjust the generator soonish.~~

~~`AbortSignal` was added in NodeJS 15.0.0 so it should be fine to use~~

EDIT: Just got reminded by a notification that this is in flight: https://github.com/kubernetes-client/javascript/pull/2583

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -16,7 +16,10 @@
         "importHelpers": true,
         "skipLibCheck": true,
         "esModuleInterop": true,
-        "useDefineForClassFields": false
+        "useDefineForClassFields": false,
+        "paths": {
+            "node-fetch/externals": ["./node_modules/@types/node-fetch/externals.d.ts"]
+        }
         // enable this in the future
         // "declarationMap": true
     },


0x5457:
Without modifying the openapi-generator, I tested two approaches to resolve the AbortSignal type import in the generated code:

1. add a paths mapping in tsconfig.json.
```json
{
  "compilerOptions": {
    "paths": {
      "node-fetch/externals": ["./node_modules/@types/node-fetch/externals.d.ts"]
    }
  }
}
```

2. change moduleResolution back to "node"(same as [openapi-generator (tsconfig.json)](https://github.com/OpenAPITools/openapi-generator/blob/aacbdf88c0ff3f10cb5227e608fafc23d4f8f058/samples/client/echo_api/typescript/build/tsconfig.json#L6)).   This also seems to fix the resolution, although I’m not entirely sure about possible side effects—especially around the differences between ESM and CommonJS exports.

mstruebing:
@0x5457 thanks, I think this would be fine to use until https://github.com/kubernetes-client/javascript/pull/2583 is ready.

I think we should merge this PR even though we will regenerate the client, but as this pins the generator to a way more fresh commit we will not regenerate with an old generator by accident.

cjihrig (approved):
/lgtm
/approve

Issue #2601, opened 2025-08-17
NODE_TLS_REJECT_UNAUTHORIZED environment variable not respected in applyToFetchOptions


0x5457:
**Describe the bug**
I'm using `@kubernetes/client-node` to send HTTP requests to an apiserver with a self-signed certificate. I can't provide the CA, so I'm trying to use NODE_TLS_REJECT_UNAUTHORIZED=0 to skip the tls verification. However, this doesn't work.

**Client Version**
1.3.0

**Server Version**
N/A

**To Reproduce**
```
import * as k8s from '@kubernetes/client-node';
import fetch from 'node-fetch';

const url = "https://localhost:443";  // server with self-signed certificate
const kc = new k8s.KubeConfig();

const res = await fetch(url, await kc.applyToFetchOptions({}));
console.log(res.status);
```
**Expected behavior**
Print status code 200

**Environment (please complete the following information):**

- OS:  Ubuntu 24.04
- Node.js version v24.4.1

abdulfarid-dev:
Hi, I was able to reproduce this issue on my local setup (Node.js v24.4.1, Ubuntu 24.04).  
I’d like to work on fixing this . Could you please assign it to me?  

Thanks!

abdulfarid-dev:
“Hi 👋 I commented a f
One day ago and I’m still interested in contributing to this issue. Could you please let me know if it’s okay for me to work on it? Thanks!”

0x5457:
Hi @abdulfarid-dev ,

I'm already working on fixing this issue in PR #2602. I'll continue handling it until the fix is complete.

Thanks for your willingness to help!

PR #2602, opened 2025-08-17
fix: respect NODE_TLS_REJECT_UNAUTHORIZED environment variable

0x5457:
fix: #2601

After testing, I found that the fetch behavior is as follows:
- Fetch only uses the NODE_TLS_REJECT_UNAUTHORIZED environment variable when the rejectUnauthorized property is completely absent from agent options
- If rejectUnauthorized is explicitly set to any value (including undefined), the environment variable is ignored

Also should I can create test certificates in the test-data directory to avoid introducing the selfsigned library


src/config_test.ts
```diff
@@ -491,6 +493,28 @@ describe('KubeConfig', () => {
...
+            const { server, host, port } = await createTestHttpsServer();
+            const originalValue = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+            after(() => {
+                server.close();
+                process.env.NODE_TLS_REJECT_UNAUTHORIZED = originalValue;
+            });
+
+            const kc = new KubeConfig();
+            const rc = new RequestContext(`https://${host}:${port}`, HttpMethod.GET);
```
brendandburns: I'd prefer this test used an actual client (with mock data) instead of making it's own request.

0x5457: @brendandburns Thanks for the review! I understand the preference for using an actual client with mock data. However, I'm facing a challenge with this specific test case.

The whole point here is to make sure that `NODE_TLS_REJECT_UNAUTHORIZED` actually works when we're hitting HTTPS servers with self-signed certs. The tricky part is that mocking libraries like `nock` just intercept the requests and return fake responses - they never actually make the real HTTPS connection, so there's no way to test if the TLS validation is getting bypassed properly.

What do you think about modifying the [createTestHttpsServer](https://github.com/kubernetes-client/javascript/pull/2602/files#diff-f7d16ccc3d41b45b1511b3e65724c67009d49174a7a290c4034487e89e13141dR1856) to make it  returns Kubernetes API responses and use the real client to connect to it

brendandburns: Thanks for the detail. I think that plan is fine, feel free to just return some hard-coded JSON data that looks like a valid k8s response.

0x5457: I've updated the test to use the real Kubernetes client as requested

I kept both the `CoreV1Api.listNamespace()` test (main case) and the `kc.applyToFetchOptions()` test because `applyToFetchOptions` is a public API method that users may call directly, and the high-level API clients don't actually use this method internally.

brendandburns:
/lgtm
/approve

Thanks for updating the test.

brendandburns:
Tests are failing with some error related to generating the certificate.

package.json
```diff
@@ -83,6 +83,7 @@
         "nock": "^14.0.5",
         "prettier": "^3.0.0",
         "pretty-quick": "^4.0.0",
+        "selfsigned": "^3.0.1",
```
cjihrig: So, [`selfsigned`](https://pkg-size.dev/selfsigned) looks to be about 1.7MB, and it is only used in one test. It might be simpler to use a fixture certificate.

src/config.ts
```diff
@@ -202,7 +202,9 @@ export class KubeConfig implements SecurityAuthentication {
         agentOptions.key = opts.key;
         agentOptions.pfx = opts.pfx;
         agentOptions.passphrase = opts.passphrase;
-        agentOptions.rejectUnauthorized = opts.rejectUnauthorized;
+        if (opts.rejectUnauthorized !== undefined) {
```
cjihrig, line 206: The changes in this file are very likely to be refactored in the future by someone missing the context of this PR.

Can you either add comments to the two code blocks in this file, or rewrite as something along the lines of `agentOptions.rejectUnauthorized = opts.rejectUnauthorized ?? process.env.NODE_TLS_REJECT_UNAUTHORIZED !== '0';`. If you go with the second option, it probably makes sense to put the logic in a function, [similar to what Node core does](https://github.com/nodejs/node/blob/0a46aa9e5dafe31e5318b4e7b3d740d62f6f5e0c/lib/internal/options.js#L148).

src/config_test.ts
```diff
@@ -491,6 +493,63 @@ describe('KubeConfig', () => {
...
+
+            strictEqual(namespaceList.kind, 'NamespaceList');
+            strictEqual(namespaceList.items.length, 1);
+            strictEqual(namespaceList.items[0].metadata?.name, 'default');
+
+            const res2 = await fetch(`https://${host}:${port}`, await kc.applyToFetchOptions({}));
+            strictEqual(res2.status, 200);
+            strictEqual(await res2.text(), 'ok');
+
+            delete process.env.NODE_TLS_REJECT_UNAUTHORIZED;
```
cjihrig: Is this line necessary since the `after()` hook is assigning a value to `process.env.NODE_TLS_REJECT_UNAUTHORIZED`?

src/config_test.ts
```diff
@@ -491,6 +493,63 @@ describe('KubeConfig', () => {
...
+                    res.writeHead(200);
+                    res.end('ok');
+                }
+            });
+
+            const originalValue = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+            after(() => {
+                server.close();
+                process.env.NODE_TLS_REJECT_UNAUTHORIZED = originalValue;
```
cjihrig, line 526: I would reverse the order of these two lines in case `server.close()` throws for some reason.

mstruebing:
> Tests are failing with some error related to generating the certificate.

@0x5457 seems like you would need to pass in an option to define the key size: https://github.com/jfromaniello/selfsigned?tab=readme-ov-file#options 2048 should be sufficient.

Having a fixture certificate seems better to me as well but as this is only a dev dependency I would be fine with it.

src/config.ts
0x5457: I added comments

package.json
0x5457: Replaced `selfsigned` dependency with static test certificates 
`testdata/certs/test-cert.pem`

0x5457:
Replaced `selfsigned` dependency with static test certificates

Generated using OpenSSL:
```sh
mkdir -p testdata/certs
openssl genrsa -out testdata/certs/test-key.pem 2048
openssl req -new -x509 -key testdata/certs/test-key.pem \
    -out testdata/certs/test-cert.pem -days 999999 -subj "/CN=localhost"
```

PR #2635, opened 2025-09-26
feat: allow to set the cwd in copyFromPod for use in tar

dherbolt:
It is just a cherry-pick of the original [commit](https://github.com/kubernetes-client/javascript/commit/2bb4ab70ce92caca1c58fd411e1042625a533fa8)


dherbolt:
Looks like v1.3 uses "tar-fs" instead of the "tar" package, so I'm not sure how many changes it will require.

cjihrig:
I think we would want this to target the `main` branch (otherwise, it will get lost again). Also, can you please sign the CLA.

dherbolt:
> I think we would want this to target the `main` branch (otherwise, it will get lost again). Also, can you please sign the CLA.

Yes, it should also be in the main branch, but as I wrote above, there is a change of npm `tar` package in the main branch, so it will probably require more changes/testing.

cjihrig:
> there is a change of npm tar package in the main branch

Have you verified that it requires extra changes? It looks like the cwd is only added to a `tar` system command, and not actually passed to the imported module.

dherbolt:
> > there is a change of npm tar package in the main branch
> 
> Have you verified that it requires extra changes? It looks like the cwd is only added to a `tar` system command, and not actually passed to the imported module.

You are right, it does not depend on the `tar` package. I got somehow confused. sorry.

mstruebing:
Hey @dherbolt thanks for the PR there are conflicts in the code which needs to be resolved, also there is the `cpToPod` function which had the optional argument `cwd` argument as well, I think we should add it there as well.

What would be extra nice would be to have a simple integration test that makes sure it works: https://github.com/kubernetes-client/javascript/blob/main/src/test/integration/index.ts

If you need help or further guidance feel free to ask

dherbolt:
> Hey @dherbolt thanks for the PR there are conflicts in the code which needs to be resolved, also there is the `cpToPod` function which had the optional argument `cwd` argument as well, I think we should add it there as well.
> 
> What would be extra nice would be to have a simple integration test that makes sure it works: https://github.com/kubernetes-client/javascript/blob/main/src/test/integration/index.ts
> 
> If you need help or further guidance feel free to ask

I agree, the `copyToPod` function should also support `cwd` argument. Unfornutelly, there is a change in the library used to pack the files ("tar" => "tar-fs"), so I will need more time to find how to pass the `cwd` arg into it.

.github/workflows/generate-javascript.yml
```diff
@@ -1,53 +1,52 @@
 name: Generate
 
 on:
-  workflow_dispatch:
```
dherbolt, line 4: oh, I have no idea why there are changes in whitespaces after rebase... sorry.

cjihrig: Can you back out the unrelated changes in this file and README.md.

dherbolt: > oh, I have no idea why there are changes in whitespaces after rebase... sorry.

For some unknown  reason pre-commit hook reformated those files.

dherbolt:
Hi @cjihrig is there a plan to add this fix to version 1.1? I can see that it is now part of v1.4, but it would be nice to have it available in older versions because we use Kube 1.32, which is fully compatible with JS client v1.1.
Thanks in advance

cjihrig:
> is there a plan to add this fix to version 1.1?

Not that I am aware of. @brendandburns might be a better person to answer that question. Although, according to the [compatibility table](https://github.com/kubernetes-client/javascript?tab=readme-ov-file#compatibility), 1.32 and v1.4 should be compatible (the JS client may have additional features, but that shouldn't be a problem).

dherbolt:
Hmm... with version 1.4 and Kube 1.32, I'm getting an error:
```
node_modules/@kubernetes/client-node/node_modules/tar-stream/extract.js:292
    cb(this._finished ? null : new Error('Unexpected end of data'))
```

Looks like the WS connection is closed before all data is sent. I receive only 316 bytes.

mstruebing:
We have a `release-1.1` branch where we could merge commits into it and publish a new 1.1 version i.e. 1.1.3.

Can you try the same from `main`? There was this PR in between 1.4.0 and now: https://github.com/kubernetes-client/javascript/pull/2664

dherbolt:
@mstruebing It would be nice to backport the commits into the 1.1 branch and release it as a patch.
I can confirm the `copyFromPod` works from `main` branch.

Thanks in advance.

mstruebing:
I just tried it and it was relatively easy: https://github.com/kubernetes-client/javascript/pull/2701

PR #2636, opened 2025-09-26
Fix syntax for environment variable exports

brendandburns:


.github/workflows/generate-javascript.yml
```diff
@@ -26,8 +26,8 @@ jobs:
           node-version: '20'
       - name: Generate Openapi
         run: |
-          echo "export KUBERNETES_BRANCH=${{ github.event.inputs.kubernetesBranch }} >> ./settings"
-          echo "export GEN_COMMIT="${{ github.event.inputs.genCommit }}" >> ./settings"
+          echo "export KUBERNETES_BRANCH="${{ github.event.inputs.kubernetesBranch }}"" >> ./settings
```
cjihrig: Is it possible to mix single and double quotes on these lines? Or, maybe some of the quotes could be escaped? Right now, it's kind of confusing. I think you're intending to add quotes around the value of `KUBERNETES_BRANCH`, but it seems like the first added quote would match the quote before `export` and the second set of quotes would be an empty string. Maybe I'm misunderstanding the intention though?

mstruebing: Single quotes for the branch name would be better as this wouldn't result in shell expanding when we would have some fancy branch names.

I'm not 100% sure if GitHub-actions would replace that value in single quotes, but I guess so:

```suggestion
          echo "export KUBERNETES_BRANCH='${{ github.event.inputs.kubernetesBranch }}'" >> ./settings
```

brendandburns: Actually the real thing I'm trying to do is get rid of the end quote at `settings"` which was actually what was causing the issue.  I think I could eliminate the interior quotes all together unless version is going to have spaces in it, which seems unlikely.

cjihrig: Seems fine to me.

mstruebing: Go ahead, sounds fine to me as well.

brendandburns:
Addressed comment, please take another look.

PR #2646, opened 2025-10-01
fix: readd error callback on informer error and connect events

### LINKED ISSUE #2448: Version 1.0.0+ expected Informer regression?
**Describe the bug**
Starting client 1.0.0, I noticed two things which have been reverted to a 5/6 years old:
- the informer interface doesn't "support" anymore ErrorCallback  for on/off("connect" and on/off("error"
- the makeInformer function returns an Informer<T> instead of Informer<T> & ObjectCache<T>

Both things above have been introduced 3 years ago.
So, i was wondering if it's expected to have reverted these to an old version?

Thank you

**Client Version**
1.0.0+

**Server Version**
N/A

**To Reproduce**
N/A

**Expected behavior**

**Example Code**
Code snippet for what you are doing

- OS: [e.g. Windows, Linux]
- Node.js version [eg. 20]
- Cloud runtime [e.g. Azure Functions, Lambda]

**Additional context**

mstruebing:
This should finally fix #2448


WeetA34:
thank you

PR #2660, opened 2025-10-14
fix(yaml): add support for YAML 1.1 parsing in Kubernetes manifests

### LINKED ISSUE #2539: YAML 1.2 only is supported

Tanmay-008:
##Summary
This PR fixes YAML parsing inconsistencies by replacing js-yaml with the modern yaml library, which supports both YAML 1.1 and 1.2 specifications.

##What’s Changed
Replaced js-yaml with yaml parser.

Added support for parsing multiple YAML documents using both YAML 1.1 and 1.2 versions.

Updated loadYaml and loadAllYaml functions to accept YamlParseOptions with version selection.

Fixes #2539

### DIFF
--- a/src/yaml.ts
+++ b/src/yaml.ts
@@ -1,6 +1,6 @@
-import yaml from 'js-yaml';
+import yaml from 'yaml';
 import { getSerializationType } from './util.js';
-import { KubernetesObject } from './types.js';
+import { KubernetesObject, YamlParseOptions } from './types.js';
 import { ObjectSerializer } from './serializer.js';
 
 /**
@@ -9,8 +9,8 @@ import { ObjectSerializer } from './serializer.js';
  * @param opts - Optional YAML load options.
  * @returns The deserialized Kubernetes object.
  */
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    const yml = yaml.load(data, opts) as any as KubernetesObject;
+export function loadYaml<T>(data: string, opts?: YamlParseOptions): T {
+    const yml = yaml.parse(data, { version: '1.1', ...opts }) as any as KubernetesObject;
     if (!yml) {
         throw new Error('Failed to load YAML');
     }
@@ -25,12 +25,12 @@ export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
  * @param opts - Optional YAML load options.
  * @returns An array of deserialized Kubernetes objects.
  */
-export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
-    const ymls = yaml.loadAll(data, undefined, opts);
+export function loadAllYaml(data: string, opts?: YamlParseOptions): any[] {
+    const ymls = yaml.parseAllDocuments(data, { version: '1.1', ...opts });
     return ymls.map((yml) => {
-        const obj = yml as KubernetesObject;
+        const obj = yml.toJS() as KubernetesObject;
         const type = getSerializationType(obj.apiVersion, obj.kind);
-        return ObjectSerializer.deserialize(yml, type);
+        return ObjectSerializer.deserialize(obj, type);
     });
 }
 
@@ -40,9 +40,9 @@ export function loadAllYaml(data: string, opts?: yaml.LoadOptions): any[] {
  * @param opts - Optional YAML dump options.
  * @returns The YAML string representation of the serialized Kubernetes object.
  */
-export function dumpYaml(object: any, opts?: yaml.DumpOptions): string {
+export function dumpYaml(object: any, opts?: YamlParseOptions): string {
     const kubeObject = object as KubernetesObject;
     const type = getSerializationType(kubeObject.apiVersion, kubeObject.kind);
     const serialized = ObjectSerializer.serialize(kubeObject, type);
-    return yaml.dump(serialized, opts);
+    return yaml.stringify(serialized, { version: '1.1', ...opts });
 }
--- a/src/yaml_test.ts
+++ b/src/yaml_test.ts
@@ -154,4 +154,55 @@ spec:
         // not using strict equality as types are not matching
         deepEqual(actual, expected);
     });
+
+    it('should parse octal values correctly using default YAML 1.1', () => {
+        const yamlStr = `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: test
+data:
+  mode: 0644
+`;
+        const obj = loadYaml<{
+            data: { mode: number };
+            metadata: { name: string };
+            apiVersion: string;
+            kind: string;
+        }>(yamlStr);
+        strictEqual(obj.data.mode, 420);
+    });
+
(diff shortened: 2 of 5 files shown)


src/yaml_test.ts
```diff
@@ -154,4 +154,55 @@ spec:
         // not using strict equality as types are not matching
         deepEqual(actual, expected);
     });
+
+    it('should parse octal values correctly using default YAML 1.1', () => {
```
cjihrig, line 158:
```suggestion
    it('should parse octal values as numbers using default YAML 1.1', () => {
```

src/yaml_test.ts
```diff
@@ -154,4 +154,55 @@ spec:
...
+        const obj = loadYaml<{
+            data: { mode: number };
+            metadata: { name: string };
+            apiVersion: string;
+            kind: string;
+        }>(yamlStr);
+        strictEqual(obj.data.mode, 420);
+    });
+
+    it('should treat octal as string if version 1.2 is provided', () => {
```
cjihrig, line 176: I'm curious why this test is so much different from the one before it. Couldn't it be the exact same setup, but with a different YAML version specified and a different expected value in the assertion?

src/yaml.ts
```diff
@@ -9,8 +9,8 @@ import { ObjectSerializer } from './serializer.js';
  * @param opts - Optional YAML load options.
  * @returns The deserialized Kubernetes object.
  */
-export function loadYaml<T>(data: string, opts?: yaml.LoadOptions): T {
-    const yml = yaml.load(data, opts) as any as KubernetesObject;
+export function loadYaml<T>(data: string, opts?: YamlParseOptions): T {
```
cjihrig, line 12: I believe this is still a breaking change to end users. Prior to this change, they could pass in `LoadOptions` from the `js-yaml` library. After this change, the available options are different and there is no logic to support the old options.

cjihrig:
It looks like the CLA still needs to be signed.

src/yaml.ts
Tanmay-008: @cjihrig  ,I have reviewed the documentation for both libraries and compared their options. I found that most of the options have the same functionality, but a few methods differ in behavior. Do you think we could use the Adapter Design Pattern to address this issue and maintain compatibility while migrating to the new library?

cjihrig: I think we need to decide what options we want to support and go from there. If we want to support the options from the previous library, then we will need an adapter. If we want to support the options from the new library, then we'll need to hold this PR until the next major release. I think no matter which route we go, we should define our own type for the options (as you've done here) so we can help insulate ourselves from this sort of problem in the future.

cc: @brendandburns @mstruebing

Tanmay-008: Thanks, that’s clear. Let’s hold off until we get feedback from @brendandburns and @mstruebing about which option set we want to support going forward.

mstruebing: I definitely think we should define our own options as @cjihrig already said.
I would think we should make this PR fine and hold it until the next release except someone thinks it's worth to put in the effort to make this mergeable without a breaking change, but I would be fine with waiting.

Tanmay-008: Thanks @cjihrig @mstruebing, I understand your point. It might take me some time, but I’d like to work on modifying this PR to avoid the breaking change. I’ll try to achieve that by adding an adapter layer to maintain backward compatibility.

Issue #2663, opened 2025-10-17
cpFromPod results in `Error: Unexpected end of data`


sachiniyer:
**Describe the bug**
`cpFromPod` results in errors when I r

**Client Version**
the latest commit on main 

**Server Version**
e.g. `v1.34.0`

**To Reproduce**
run `cpFromPod` on any file in the container.

**Expected behavior**
I expect to see that there is a file copied to my local filesystem when I use `cpFromPod`

**Example Code**
```
    await cp.cpFromPod(
        'default', 
        'test-pod', 
        'test-pod',
        'test.txt',
        targetDir,
        '/tmp'
    );
```

**Environment (please complete the following information):**

- OS: Linux
- Node.js: v22.13.0

**Additional context**
N/A

cjihrig:
This should be fixed by https://github.com/kubernetes-client/javascript/pull/2664

PR #2664, opened 2025-10-17
fix: cp from pod was broken

sachiniyer:
I believe https://github.com/kubernetes-client/javascript/commit/3a452ed7f6981ae4a6169075caba156809e55eff broke cp from pod (I can't get it to work and get `Error: Unexpected end of data` messages). I created an issue with more info here - https://github.com/kubernetes-client/javascript/issues/2663

The [tar-fs readme](https://github.com/mafintosh/tar-fs?tab=readme-ov-file#usage) says that it does not gunzip automatically (like the previously used `tar` package did). 

Apologies if I missed something here.

### DIFF
--- a/src/test/integration/cpFromPod.ts
+++ b/src/test/integration/cpFromPod.ts
@@ -2,6 +2,7 @@ import assert from 'node:assert';
 import fs from 'node:fs';
 import path from 'node:path';
 import os from 'node:os';
+import { setTimeout } from 'node:timers/promises';
 import { CoreV1Api, KubeConfig, V1Pod } from '../../index.js';
 import { Cp } from '../../cp.js';
 import { generateName } from './name.js';
@@ -40,7 +41,7 @@ export default async function cpFromPod() {
             podReady = true;
             break;
         }
-        await new Promise((resolve) => setTimeout(resolve, 1000));
+        await setTimeout(1000);
     }
 
     assert.strictEqual(podReady, true, 'Pod did not become ready in time');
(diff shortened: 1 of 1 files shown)


cjihrig:
I don't know how much effort it would be, but is it possible to write a regression test (beyond checking the `tar` flags)? For example, we have this [integration test](https://github.com/kubernetes-client/javascript/blob/main/src/test/integration/index.ts) that will run against an actual cluster.

sachiniyer:
Yup @cjihrig ! I've added a previously failing integration test (if you make the flags `zcf`, you will see that this test fails as well).

Lmk, if there is anything unsatisfactory about the way I have written it.

src/test/integration/cpFromPod.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    console.log('Waiting for pod to be ready...');
+    let podReady = false;
+    for (let i = 0; i < 30; i++) {
+        const currentPod = await coreV1Client.readNamespacedPod({ name: testPodName, namespace });
+        if (currentPod.status?.phase === 'Running') {
+            podReady = true;
+            break;
+        }
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
cjihrig: Here and on line 64, can you use `setTimeout()` from `node:timers/promises`. Then, you can `await setTimeout()` directly without needing to create an additional `Promise`.

sachiniyer: Yup, I've made that change.

PR #2665, opened 2025-10-17
fix: timeoutSignal is lost after being overriden with controller.signal

sachiniyer:
I believe that the watch timeout abort signal was getting overwritten, and I think https://github.com/kubernetes-client/javascript/commit/0ae5ddfbdd5ea9f31b40c458b936f1dcdba94f59 just forgot to delete line 44.

I wrote a one-off test to make sure this was really happening, but please let me know if I should be adding a unit test for this functionality as part of this PR.

### DIFF
--- a/src/watch_test.ts
+++ b/src/watch_test.ts
@@ -459,7 +459,7 @@ describe('Watch', () => {
         const watch = new Watch(kc);
 
         // NOTE: Hack around the type system to make the timeout shorter
-        (watch as any).requestTimeoutMs = 1;
+        (watch as any).requestTimeoutMs = 10;
 
         let doneErr: any;
 


cjihrig:
> I wrote a one-off test to make sure this was really happening, but please let me know if I should be adding a unit test for this functionality as part of this PR.

That would be very helpful!

sachiniyer:
Hi @cjihrig !

I've added a previously failing test. If you add back the line that I deleted, you will see that the test hangs/times out.

I also made the timeout value a private instance property to make it more easily testable - hopefully that's alright.

src/watch_test.ts
```diff
@@ -452,6 +452,42 @@ describe('Watch', () => {
...
+            (err: any) => {
+                doneCalled = true;
+                doneErr = err;
+                doneResolve();
+            },
+        );
+
+        await donePromise;
+
+        strictEqual(doneCalled, true);
```
cjihrig: I don't think we need `doneCalled` at all since we are already validating `doneErr`.

src/watch_test.ts
```diff
@@ -452,6 +452,42 @@ describe('Watch', () => {
...
     });
 
+    it('should timeout when server takes too long to respond', async (t) => {
+        const kc = await setupMockSystem(t, (_req: any, _res: any) => {
+            // Don't respond - simulate a hanging server
+        });
+        const watch = new Watch(kc);
+
+        // NOTE: Hack around the type system to make the timeout shorter
+        (watch as any).requestTimeoutMs = 100;
```
cjihrig: Does the test still pass with a much smaller timeout (like 1ms)? Looking at the test run in the CI, this test takes a good bit longer than most of the other tests.

cjihrig (commented):
Mostly LGTM, thanks.

sachiniyer:
Made both adjustments @cjihrig 

Lmk if anything else looks off.

cjihrig:
The code looks good to me, but in CI it's crashing the Node.js process in C code somehow (impressive):

```
node: ../deps/uv/src/unix/stream.c:446: uv__stream_destroy: Assertion `!uv__io_active(&stream->io_watcher, POLLIN | POLLOUT)' failed.
✖ /home/runner/work/javascript/javascript/dist/watch_test.js (1764.106097ms)
```

sachiniyer:
Hey @cjihrig , I have no idea why this is happening (or how to debug it really).

I bumped the timeout to 10 ms for the test, and the actions are passing on a test pr that I created on my fork of this repo - https://github.com/sachiniyer/kubernetes-javascript-client/pull/1

If there is a way to be doing better due diligence for this change, please lmk.

cjihrig:
Before merging, do @brendandburns or @mstruebing have any concerns given https://github.com/kubernetes-client/javascript/pull/2665#issuecomment-3418642911

mstruebing:
I would run the pipeline a couple of times - I just restarted it a couple of times and it seems stable.
That's probably some bug in NodeJS itself, so I'm fine with merging when it runs kinda stable

sachiniyer:
Hey @cjihrig and @mstruebing, thanks for reviewing this PR.

Please lmk if there is anything I can do from my side to move this along. If we are worried about the test failing, I can attempt to write it even cleaner (or maybe remove the test and just merge the fix?)

cjihrig (approved):
/lgtm
/approve

PR #2675, opened 2025-10-23
Update watch-example.ts to reflect `done` error is conditionally passed

loucadufault:
simple example change

There is very little documentation about this class online, so this example is pretty much the only usage (aside from looking at code) that explains the role of error param

### DIFF
--- a/examples/typescript/watch/watch-example.ts
+++ b/examples/typescript/watch/watch-example.ts
@@ -29,9 +29,9 @@ const req = await watch
         // done callback is called if the watch terminates either normally or with an error
         (err) => {
             if (err) {
-                console.log(err);
+                console.error(err);
             } else {
-                console.log('watch done')
+                console.log('watch finished normally')
             }
         },
     )


examples/typescript/watch/watch-example.ts
```diff
@@ -26,9 +26,13 @@ const req = await watch
             }
             console.log(apiObj);
         },
-        // done callback is called if the watch terminates normally
+        // done callback is called if the watch terminates either normally or with an error
         (err) => {
-            console.log(err);
+            if (err) {
+                console.log(err);
```
cjihrig:
```suggestion
                console.error(err);
```

examples/typescript/watch/watch-example.ts
```diff
@@ -26,9 +26,13 @@ const req = await watch
...
+            } else {
+                console.log('watch done')
```
cjihrig:
```suggestion
                console.log('watch finished normally')
```

cjihrig (approved):
Left two minor suggestions. Thanks for the improvement.

examples/typescript/watch/watch-example.ts
loucadufault: thanks, made the change

PR #2701, opened 2025-11-25
fix: cp from pod (backport to release-1.1)

mstruebing:
I'm not sure for how long we want to support older release branches, but as this was mostly a quick one I've just did it.
Is there any specification in the other clients? Do we need to clarify how long we support older branches or is it just fine if we don't have any guidance on it and just do it on a case by case basis?

@dherbolt can you confirm this works?

I've ran the integration test and it seems to work just fine but want to make sure it works for you as well.

Once this is in, we need to increase the version in the release-1.1 branch and publish it to npm.

ref: https://github.com/kubernetes-client/javascript/pull/2635#issuecomment-3575013458

### DIFF
--- a/src/cp.ts
+++ b/src/cp.ts
@@ -30,7 +30,7 @@ export class Cp {
     ): Promise<void> {
         const tmpFile = tmp.fileSync();
         const tmpFileName = tmpFile.name;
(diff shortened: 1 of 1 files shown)


dherbolt:
Yes, it works! Thanks.

PR #2702, opened 2025-11-25
chore: rev new version

mstruebing:
Rev new version to be able to publish: https://github.com/kubernetes-client/javascript/pull/2701

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@kubernetes/client-node",
-    "version": "1.1.2",
+    "version": "1.1.3",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@kubernetes/client-node",
-            "version": "1.1.2",
+            "version": "1.1.3",
             "license": "Apache-2.0",
             "dependencies": {
                 "@types/js-yaml": "^4.0.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@kubernetes/client-node",
-    "version": "1.1.2",
+    "version": "1.1.3",
     "description": "NodeJS client for kubernetes",
     "type": "module",
     "repository": {


mstruebing:
I can not release a new version - there are multiple restrictions.
1. I think we have some kind of environment protection for the `gh-pages` branch (should we remove the branch and the associated actions?)
2. The `build-package.sh` script executes `pre-check.sh` which makes sure the script is only run from the `main` branch.

I've tried to circumvent 1. by temporary adding the branch in the allowed branches for the production environment - which kinda worked. 
Then I've got hit 2 (I think I have a Déjà vu right now) and I'm not how to proceed.

My proposal:

We change the script `pre-check.sh` to be allowed to run on `main` and `release-*` branches and we include the release branches also in the allowed branches for the production environment.

Otherwise we would need always manually release fixes for older versions.

What do you think?

cc @brendanburns @cjihrig  

Here are the runs btw:
https://github.com/kubernetes-client/javascript/actions/runs/19671313188
https://github.com/kubernetes-client/javascript/actions/runs/19671291175

dherbolt:
Any update here? I'm curious how previous versions 1.1 and 1.2 were released...
Thanks.

EDIT: Ah, I see... They were released before 1.2, so probably from the `main` branch.

mstruebing:
Exactly!
The only one who could release manually would be @brendanburns right now as he has control over the npm package.

brendandburns:
Yeah, we've never actually released a branched release, so yes, I think we do need to update the pre-check script.

To clarify, there's an NPM token stored in the GH action so anyone with access to that action can perform a release.

dherbolt:
Do you have any updates here, please? I can see that it should now be possible to release older versions from the `release-*` branch, but v1.1.3 has not been released yet.

Thanks for your time.

mstruebing:
Tried it today but it seems like the npm token is expired. @brendandburns can you update the token?

```
npm notice Access token expired or revoked. Please try logging in again.
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@kubernetes%2fclient-node - Not found
npm error 404
npm error 404  '@kubernetes/client-node@1.1.3' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
```

https://github.com/kubernetes-client/javascript/actions/runs/20072811449/job/57579276703

dherbolt:
@mstruebing Not sure what type of token is used here but NPM tokens, but classic tokens are revoled from today https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/

mstruebing:
I'm not sure what kind of token was used but we need to have an access token that has access to the npm package in our CI.
As of now, as far as I know only @brendandburns has access to the npm package and thus can maintain this token.

brendanburns:
So I dug into this some more and it looks like there's a way to permanently trust the GitHub action on the npm side without requiring constant token renewal.

I will try to get this set up when I have a free moment.


________________________________

brendandburns:
Ok, I set up the connection between our github action and npm, please try this again.

mstruebing:
@brendanburns the exact same issue is stated in the logs. I've run it round about 1 hour after you've announced the changes

```
npm notice Access token expired or revoked. Please try logging in again.
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@kubernetes%2fclient-node - Not found
npm error 404
npm error 404  '@kubernetes/client-node@1.1.3' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2025-12-10T19_42_23_712Z-debug-0.log
```

https://github.com/kubernetes-client/javascript/actions/runs/20111101932/job/57708322447

mstruebing:
Update: I've deleted our Production `NPM_AUTH_TOKEN` as I've thought it's expired anyway, so if we want to get that back up working we need to replace it anyway. Getting a new error now. I've read the npm docs and spotted that we need this one permission (`id-token: write`) in order to create oidc tokens: https://docs.npmjs.com/trusted-publishers#github-actions-configuration

Creating a new PR right now.

```
npm notice
npm error code ENEEDAUTH
npm error need auth This command requires you to be logged in to https://registry.npmjs.org/
npm error need auth You need to authorize this machine using `npm adduser`
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2025-12-11T12_31_49_509Z-debug-0.log
```

https://github.com/kubernetes-client/javascript/actions/runs/20133131631/job/57779361806

mstruebing:
Still the same issue, not sure what could be wrong otherwise. When clicking on the `view workflow file` button in the context menu of the failed run I can see that it seems to use the workflow with the adjusted permissions.

@brendanburns can you double check you've pasted the correct data into the npm connection?

If I'm not mistaken it should be:
* `Organization or user*`: `kubernetes-client`
* `Repository*`: `javascript`
* `Workflow filename*`: `release.yml`
* `Environment name`: `production`

https://github.com/kubernetes-client/javascript/actions/runs/20133784506/job/57781582724

brendandburns:
This is what I see in NPM:

image

mstruebing:
That looks correct though.
I will play around with a package I own in the next few days when I have time to figure this out.

mstruebing:
There is one issue left, we need to define a tag as npm defined it this way and it can not be a semver version.

`npm error Cannot implicitly apply the "latest" tag because previously published version 1.4.0 is higher than the new version 1.1.3. You must specify a tag using --tag.` (https://github.com/kubernetes-client/javascript/actions/runs/20178469809/job/57932550345)

https://github.com/npm/npm/issues/8531

I've tried to do a couple of publishes in my test repo and it seems to work with a tag set: https://www.npmjs.com/package/@mstruebing/npm-publish-test?activeTab=versions

I think we could still use `latest` tag on `main` and `patch-<VERSION>` in other cases.

Will bring up a PR.

mstruebing:
@dherbolt I have an early christmas present for you - 1.1.3 is on npm: https://www.npmjs.com/package/@kubernetes/client-node/v/1.1.3

dherbolt:
@mstruebing 
Nice 🎄🎁—thanks for the early Christmas gift! I see 1.1.3 finally made it to npm. I’ve already tested it on my side and everything looks good 👍.

PR #2716, opened 2025-12-09
feat: add timeout support to RequestContext and Isomorphic-fetch

### LINKED ISSUE #1613: not possible to set a client timeout

dahaque:
This PR finally adds proper timeout support to the Node.js client.

Fixes #1613 

Screenshot 2025-12-09 151328
Screenshot 2025-12-09 151348

First-time contributor, be gentle! 😅


cjihrig:
Thanks for contributing. A few notes:

- Changes in `src/gen` should be made in the upstream repo: https://github.com/OpenAPITools/openapi-generator.
- The test in this PR should follow the style and naming convention of existing tests.
- At a quick glance, there at least seem to be some linter issues.

mstruebing:
As @cjihrig already said, the code in `src/gen` would be overwritten once we regenerate the client.
So if any changes are needed in that part of the code it needs to go through the upstream repo and we need to regenerate the code.

PR #2734, opened 2026-01-06
update settings rev to match 1.35 release

davidgamero:
main was generated with the release-1.35 branch, which is not a fixed tag

this means when someone opens a PR after the first 1.35.x patch release their PR will require those changes to be included.

if this was the intended behavior i will change `export KUBERNETES_BRANCH="v1.35.0"` to `export KUBERNETES_BRANCH="release-1.35"` 

either way, this PR is to align the settings file and src/gen folder.

the GEN_COMMIT variable is updated to match what was used in https://github.com/kubernetes-client/javascript/pull/2732 based on the inclusion of enum changes i noticed here https://github.com/kubernetes-client/javascript/pull/2732/changes#diff-53a4fc8b730eb7a699aab421052cfa52ddf8d6eb973bd03e58bc1c96d1dd4c96


cjihrig (approved):
/lgtm
/approve

mstruebing:
While we are at it - does anything speak against releasing a new version?
I've noticed the other clients didn't do this as well yet, is there a reason why we would want to wait?

If so, I can do it but I didn't want to waste time/effort.

cc @brendanburns

PR #2737, opened 2026-01-06
make headers comparisons case-insensitive

### LINKED ISSUE #2690: setHeaderOptions should normalize header keys
**Describe the bug**
Currently to override any header that's already set by the library you need to match the case. If the case changes for whatever reason the code breaks.
This causes hard to debug [issues](https://github.com/kubernetes-client/javascript/issues/2160#issuecomment-2620169494).

I suggest lowercasing all headers so they correctly override regardless of case used anywhere. HTTP1 defines header keys as case insensitive, HTTP2 requires lowercase headers. https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2

If you don't decide to lowercase header keys right now, `setHeaderOptions` should at least compare the new keys to existing ones case insensitively and overwrite existing values instead of adding new entries

davidgamero:
bump generator commit to https://github.com/OpenAPITools/openapi-generator/pull/22393 making header key comparisons case-insensitive for updates

resolves #2690

gen docs were moved from src/gen to src/gen/docs due to https://github.com/OpenAPITools/openapi-generator/pull/22124

this does not impact our typedoc generated docs


cjihrig (approved):
/lgtm
/approve

PR #2744, opened 2026-01-13
add useragent with package version

davidgamero:
make debugging easier by including a user agent with package version

following similar pattern to go client https://github.com/kubernetes/client-go/blob/0b06cf5bf0e19c0f026fd5eef7eecb5a346ca95f/rest/config.go#L524

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -47,7 +47,7 @@ const USER_AGENT_KEY = 'User-Agent';
 
 function getUserAgent(): string {
     const version = packagejson.version ?? '';
-    return `kubernetes-javascript-client/${version}`;
+    return `kubernetes-client-javascript/${version}`;
 }
 
 // fs.existsSync was removed in node 10
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -1691,6 +1691,11 @@ describe('KubeConfig', () => {
                     }),
                 );
             });
+            t.after(async () => {
+                await new Promise<Error | undefined>((resolve) => {
+                    server.close(resolve);
+                });
+            });
 
             const kc = new KubeConfig();
             kc.loadFromClusterAndUser(
@@ -1720,7 +1725,7 @@ describe('KubeConfig', () => {
             // Verify User-Agent header contains client name and version
             strictEqual(typeof capturedUserAgent, 'string');
             strictEqual(
-                capturedUserAgent!.startsWith('kubernetes-javascript-client/'),
+                capturedUserAgent!.startsWith('kubernetes-client-javascript/'),
                 true,
                 'capturedUserAgent should start with "kubernetes-javascript-client/"',
             );
@@ -1729,11 +1734,6 @@ describe('KubeConfig', () => {
                 true,
                 `User-Agent should include version ${expectedVersion}`,
             );
-            t.after(async () => {
-                await new Promise<Error | undefined>((resolve) => {
-                    server.close(resolve);
-                });
-            });
         });
     });
 


src/config_test.ts
```diff
@@ -1674,6 +1674,68 @@ describe('KubeConfig', () => {
...
+                    capturedUserAgent?.startsWith('kubernetes-javascript-client/'),
+                    true,
+                    'capturedUserAgent should start with "kubernetes-javascript-client/"',
+                );
+                strictEqual(
+                    capturedUserAgent?.includes(expectedVersion),
+                    true,
+                    `User-Agent should include version ${expectedVersion}`,
+                );
+            } finally {
```
cjihrig: You should be able to use [`t.after()`](https://nodejs.org/api/test.html#contextafterfn-options) to create a hook once the test starts running. We can avoid the `try...finally` that way.

src/config_test.ts
```diff
@@ -1674,6 +1674,68 @@ describe('KubeConfig', () => {
...
+                await new Promise<void>((resolve) => {
+                    server.close(() => resolve());
```
cjihrig:
```suggestion
                    server.close(resolve);
```

src/config_test.ts
```diff
@@ -1674,6 +1674,68 @@ describe('KubeConfig', () => {
...
+                const expectedVersion = packageJson.version;
+
+                // Verify version is not blank
+                strictEqual(typeof expectedVersion, 'string');
+                strictEqual(expectedVersion.length > 0, true, 'package.json version should not be blank');
+
+                // Verify User-Agent header contains client name and version
+                strictEqual(typeof capturedUserAgent, 'string');
+                strictEqual(
+                    capturedUserAgent?.startsWith('kubernetes-javascript-client/'),
```
cjihrig: Since we have already asserted that it is a string:

```suggestion
                    capturedUserAgent!.startsWith('kubernetes-javascript-client/'),
```

src/config_test.ts
```diff
@@ -1674,6 +1674,68 @@ describe('KubeConfig', () => {
...
+                    true,
+                    'capturedUserAgent should start with "kubernetes-javascript-client/"',
+                );
+                strictEqual(
+                    capturedUserAgent?.includes(expectedVersion),
```
cjihrig: Same thing here. This can also probably use `endsWith()` instead of `includes()` for a better match.

```suggestion
                    capturedUserAgent!.includes(expectedVersion),
```

src/config.ts
```diff
@@ -36,11 +36,19 @@ import WebSocket from 'isomorphic-ws';
...
+import { setHeaderMiddleware } from './middleware.js';
 
 const SERVICEACCOUNT_ROOT: string = '/var/run/secrets/kubernetes.io/serviceaccount';
 const SERVICEACCOUNT_CA_PATH: string = SERVICEACCOUNT_ROOT + '/ca.crt';
 const SERVICEACCOUNT_TOKEN_PATH: string = SERVICEACCOUNT_ROOT + '/token';
 const SERVICEACCOUNT_NAMESPACE_PATH: string = SERVICEACCOUNT_ROOT + '/namespace';
+const USER_AGENT_KEY = 'User-Agent';
+
+function getUserAgent(): string {
+    const version = packagejson.version || '';
```
cjihrig:
```suggestion
    const version = packagejson.version ?? '';
```

davidgamero:
thank you i've updated to include the suggestions

src/config.ts
```diff
@@ -36,11 +36,19 @@ import WebSocket from 'isomorphic-ws';
...
 
 const SERVICEACCOUNT_ROOT: string = '/var/run/secrets/kubernetes.io/serviceaccount';
 const SERVICEACCOUNT_CA_PATH: string = SERVICEACCOUNT_ROOT + '/ca.crt';
 const SERVICEACCOUNT_TOKEN_PATH: string = SERVICEACCOUNT_ROOT + '/token';
 const SERVICEACCOUNT_NAMESPACE_PATH: string = SERVICEACCOUNT_ROOT + '/namespace';
+const USER_AGENT_KEY = 'User-Agent';
+
+function getUserAgent(): string {
+    const version = packagejson.version ?? '';
+    return `kubernetes-javascript-client/${version}`;
```
mstruebing: Should we name it `kubernetes-client-javascript` to better match the org/repo name?

src/config_test.ts
```diff
@@ -1674,6 +1674,67 @@ describe('KubeConfig', () => {
...
+                capturedUserAgent!.startsWith('kubernetes-javascript-client/'),
+                true,
+                'capturedUserAgent should start with "kubernetes-javascript-client/"',
+            );
+            strictEqual(
+                capturedUserAgent!.endsWith(expectedVersion),
+                true,
+                `User-Agent should include version ${expectedVersion}`,
+            );
+            t.after(async () => {
```
cjihrig: This should be moved to right after the server is started. Otherwise, if any assertions fail, this will never be added.

src/config.ts
davidgamero: makes sense to me

cjihrig (approved):
/lgtm
/approve

PR #2749, opened 2026-01-17
feat: allow port forward to deployment and service

### LINKED ISSUE #2746: Port Forward currently only allows forwarding to pod resource type
Currently the API reference for port-forward is hardcoded to the /pods/ route
https://github.com/kubernetes-client/javascript/blob/main/src/portforward.ts#L43

As per the Kubernetes docs, port-forward has functionality for any Resource Type (Service, Pod, Deployment, etc)
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_port-forward/

Could there be an added parameter to change this dynamically?

mstruebing:
fixes #2746

### DIFF
--- a/src/test/integration/portForward.ts
+++ b/src/test/integration/portForward.ts
@@ -13,185 +13,142 @@ export default async function portForwardIntegration() {
     const kc = new KubeConfig();
     kc.loadFromDefault();
 
-    const coreV1Client = kc.makeApiClient(CoreV1Api);
-    const appsV1Client = kc.makeApiClient(AppsV1Api);
-    const portForward = new PortForward(kc);
+    console.log(`\n=== Port Forward Integration Test ===`);
+
+    await deploymentSuccessTest(kc);
+    await serviceSuccessTest(kc);
+    await serviceNoSelectorTest(kc);
+
+    console.log('\nAll Port Forward integration tests passed!');
+}
+
+async function deploymentSuccessTest(kc: KubeConfig): Promise<void> {
+    console.log('Deployment Port Forward Success Test...');
+    let deploymentTestPassed = false;
 
     const namespace = 'default';
-    const appName = generateName('pf-test');
+    const appName = generateName('pf-dp-test');
+    const deploymentName = `${appName}-deployment`;
+    const containerPort = 8080;
+    const localPort = 18080;
+
+    const portForward = new PortForward(kc);
+
+    await createDeployment(appName, deploymentName, namespace, containerPort, kc);
+
+    const deploymentServer = net.createServer(async (socket) => {
+        try {
+            await portForward.portForwardDeployment(
+                namespace,
+                deploymentName,
+                [containerPort],
+                socket,
+                null,
+                socket,
+            );
+        } catch (error) {
+            console.error('Deployment port forward error:', error);
+            socket.destroy();
+        }
+    });
+
+    await new Promise<void>((resolve) => {
+        deploymentServer.listen(localPort, '127.0.0.1', () => {
+            console.log(`Local server listening on port ${localPort} for deployment test`);
+            resolve();
+        });
+    });
+
+    deploymentTestPassed = await testPortForwardConnection(localPort, 'deployment');
+    deploymentServer.close();
+    assert.strictEqual(deploymentTestPassed, true, 'Failed to connect to deployment via port-forward');
+
+    console.log('Cleaning up deployment...');
+    const appsV1Client = kc.makeApiClient(AppsV1Api);
+    await appsV1Client.deleteNamespacedDeployment({ name: deploymentName, namespace });
+}
+
+async function serviceSuccessTest(kc: KubeConfig): Promise<void> {
+    console.log('Service Port Forward Success Test...');
+    let serviceTestPassed = false;
+
+    const namespace = 'default';
+    const appName = generateName('pf-svc-test');
     const serviceName = `${appName}-service`;
     const deploymentName = `${appName}-deployment`;
     const containerPort = 8080;
-    const testPort = 18080; // Local port for testing
+    const localPort = 18081;
 
-    console.log(`\n=== Port Forward Integration Test ===`);
-    console.log(`Creating test resources in namespace: ${namespace}`);
+    const portForward = new PortForward(kc);
 
-    // Create a deployment with a simple HTTP server
-    const deployment = new V1Deployment();
-    deployment.metadata = { name: deploymentName, namespace };
-    deployment.spec = {
-        replicas: 2,
-        selector: { matchLabels: { app: appName } },
-        template: {
-            metadata: { labels: { app: appName } },
-            spec: {
-                containers: [
-                    {
-                        name: 'test-server',
-                        image: 'busybox',
-                        command: [
-                            'sh',
-                            '-c',
-                            `while true; do echo -e "HTTP/1.1 200 OK\\r\\nContent-Length: 13\\r\\n\\r\\nHello World\\n" | nc -l -p ${containerPort} || true; done`,
-                        ],
-                        ports: [{ containerPort }],
-                    },
-                ],
-            },
-        },
-    };
+    await createDeployment(appName, deploymentName, namespace, containerPort, kc);
+    await createService(appName, serviceName, namespace, containerPort, kc);
 
-    console.log(`Creating deployment ${deploymentName}...`);
-    await appsV1Client.createNamespacedDeployment({ namespace, body: deployment });
-
-    // Create a service for the deployment
-    const service = new V1Service();
-    service.metadata = { name: serviceName, namespace };
-    service.spec = {
-        selector: { app: appName },
-        ports: [{ port: containerPort, targetPort: containerPort, protocol: 'TCP' }],
-        type: 'ClusterIP',
-    };
-
-    console.log(`Creating service ${serviceName}...`);
-    await coreV1Client.createNamespacedService({ namespace, body: service });
-
-    // Wait for pods to be ready
-    console.log('Waiting for deployment pods to be ready...');
-    let podsReady = false;
-    for (let i = 0; i < MAX_POD_READY_CHECKS; i++) {
-        const deployment = await appsV1Client.readNamespacedDeployment({ name: deploymentName, namespace });
-        if (
-            deployment.status?.readyReplicas === deployment.spec?.replicas &&
-            (deployment.status?.readyReplicas ?? 0) > 0
-        ) {
-            podsReady = true;
-            console.log(`Deployment is ready with ${deployment.status?.readyReplicas} replicas`);
-            break;
+    const serviceServer = net.createServer(async (socket) => {
+        try {
+            await portForward.portForwardService(
+                namespace,
+                serviceName,
+                [containerPort],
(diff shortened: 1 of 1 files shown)


examples/typescript/port-forward/port-forward-deployment.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const forward = new k8s.PortForward(kc);
+
+const namespace = process.argv[2] || 'default';
+const deploymentName = process.argv[3] || 'demo-deployment';
+const localPort = parseInt(process.argv[4] || '8080', 10);
+const remotePort = parseInt(process.argv[5] || '8080', 10);
```
cjihrig:
```suggestion
const namespace = process.argv[2] ?? 'default';
const deploymentName = process.argv[3] ?? 'demo-deployment';
const localPort = parseInt(process.argv[4] ?? '8080', 10);
const remotePort = parseInt(process.argv[5] ?? '8080', 10);
```

examples/typescript/port-forward/port-forward-deployment.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// This creates a local server that forwards traffic to a deployment in Kubernetes
+// by resolving the deployment to its first ready pod and port-forwarding to that pod.
+// Usage: npx ts-node port-forward-deployment.ts [namespace] [deploymentName] [localPort] [remotePort]
```
cjihrig: I don't think we should promote using `ts-node`. I believe it's unmaintained at this point. There is `tsx`, but also Node can run TypeScript directly now.

examples/typescript/port-forward/port-forward-service.ts
```diff
@@ -0,0 +1,43 @@
+import * as k8s from '@kubernetes/client-node';
+import net from 'node:net';
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const forward = new k8s.PortForward(kc);
+
+const namespace = process.argv[2] || 'default';
```
cjihrig: Same comment on these lines about using `??` instead of `||`.

examples/typescript/port-forward/port-forward-service.ts
```diff
@@ -0,0 +1,43 @@
...
+const serviceName = process.argv[3] || 'demo-service';
+const localPort = parseInt(process.argv[4] || '8080', 10);
+const remotePort = parseInt(process.argv[5] || '8080', 10);
+
+// This creates a local server that forwards traffic to a service in Kubernetes
+// by resolving the service to its first ready pod and port-forwarding to that pod.
+// Usage: npx ts-node port-forward-service.ts [namespace] [serviceName] [localPort] [remotePort]
```
cjihrig: Same comment about ts-node here.

examples/port-forward-deployment.js
```diff
@@ -0,0 +1,43 @@
...
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const forward = new k8s.PortForward(kc);
+
+const namespace = process.argv[2] || 'default';
+const deploymentName = process.argv[3] || 'demo-deployment';
+const localPort = parseInt(process.argv[4] || '8080', 10);
+const remotePort = parseInt(process.argv[5] || '8080', 10);
```
cjihrig: `??` here as well please.

examples/port-forward-service.js
```diff
@@ -0,0 +1,43 @@
...
+
+const kc = new k8s.KubeConfig();
+kc.loadFromDefault();
+
+const forward = new k8s.PortForward(kc);
+
+const namespace = process.argv[2] || 'default';
+const serviceName = process.argv[3] || 'demo-service';
+const localPort = parseInt(process.argv[4] || '8080', 10);
+const remotePort = parseInt(process.argv[5] || '8080', 10);
```
cjihrig: And here 😄

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,280 @@
...
+        console.log(`\nTest 1: Port forwarding to deployment ${deploymentName}...`);
+        let deploymentTestPassed = false;
+
+        const deploymentServer = net.createServer((socket) => {
+            portForward
+                .portForwardDeployment(namespace, deploymentName, [containerPort], socket, null, socket)
+                .catch((error) => {
+                    console.error('Deployment port forward error:', error.message);
+                    socket.destroy();
+                });
```
cjihrig: Can we use `async-await` and `try...catch` to be more consistent.

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,280 @@
...
+
+        await new Promise<void>((resolve) => {
+            deploymentServer.listen(testPort, '127.0.0.1', () => {
+                console.log(`Local server listening on port ${testPort} for deployment test`);
+                resolve();
+            });
+        });
+
+        // Give the server a moment to start
+        await setTimeout(500);
```
cjihrig: Is this necessary? Once the `listen()` Promise above resolves, the server should be ready to serve traffic.

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,280 @@
...
+        assert.strictEqual(deploymentTestPassed, true, 'Failed to connect to deployment via port-forward');
+
+        // Test 2: Port forward to service
+        console.log(`\nTest 2: Port forwarding to service ${serviceName}...`);
+        let serviceTestPassed = false;
+
+        const serviceServer = net.createServer((socket) => {
+            portForward
+                .portForwardService(namespace, serviceName, [containerPort], socket, null, socket)
+                .catch((error) => {
```
cjihrig: Same comment here about using `async-await` and `try...catch`.

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,280 @@
...
+
+        await new Promise<void>((resolve) => {
+            serviceServer.listen(testPort, '127.0.0.1', () => {
+                console.log(`Local server listening on port ${testPort} for service test`);
+                resolve();
+            });
+        });
+
+        // Give the server a moment to start
+        await setTimeout(500);
```
cjihrig: Same question about this being necessary.

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,280 @@
...
+
+        // Test connection to service via port-forward
+        for (let i = 0; i < 5; i++) {
+            try {
+                const response = await new Promise<string>((resolve, reject) => {
```
cjihrig: Is this logic all the same as above? If so, maybe it should be placed in its own function.

examples/typescript/port-forward/port-forward-service.ts
mstruebing: I don't necessarily disagree, but why `??` instead of `||`?

cjihrig: I think in general we should prefer `??` unless we explicitly want to default on all falsy values. In this case, `process.argv[x]` will be `undefined` if not set, so they should behave identically. I haven't benchmarked it, but I'd also expect `??` to be faster since it does slightly less work under the hood.

examples/typescript/port-forward/port-forward-deployment.ts
mstruebing: Today I learned! Thank you I didn't knew NodeJS is capable of running typescript direclty.
In case anyone else stumbles upon this comment: https://nodejs.org/en/learn/typescript/run-natively

mstruebing:
@cjihrig thanks for your review. I think I've addressed all your comments. I've made them in separate commits to easier keep track of what I already did and what I still need to do so it also be easy to review.

examples/typescript/watch/watch-example.ts
```diff
@@ -1,40 +1,40 @@
-import * as k8s from '@kubernetes/client-node';
+// import * as k8s from '@kubernetes/client-node';
+import * as k8s from '../../../dist/index.js';
```
cjihrig: This looks like maybe it was left over from local testing.

mstruebing: I've removed it, good catch.
Prettier autoformatted the file so I would leave it in here.

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,240 @@
...
+            console.log(`Deployment is ready with ${deployment.status?.readyReplicas} replicas`);
+            break;
+        }
+        await setTimeout(1000);
+    }
+
+    assert.strictEqual(podsReady, true, 'Deployment pods did not become ready in time');
+
+    try {
+        // Test 1: Port forward to deployment
```
davidgamero: LGTM
can we move test 1, test2, test3 into named functions to organize main test func body?
```ts
deploymentSuccessTest(...)
serviceSuccessTest(...)
serviceNoSelectorTest(...)
```

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,240 @@
...
+        type: 'ClusterIP',
+    };
+
+    console.log(`Creating service ${serviceName}...`);
+    await coreV1Client.createNamespacedService({ namespace, body: service });
+
+    // Wait for pods to be ready
+    console.log('Waiting for deployment pods to be ready...');
+    let podsReady = false;
+    for (let i = 0; i < 60; i++) {
```
davidgamero: nit the cycles/timouts are easier to understand when they are extracted into const
```suggestion
    const MAX_POD_READY_CHECKS = 60;
    const POD_READY_CHECK_POLL_DELAY_MS = 1000;
    for (let i = 0; i < MAX_POD_READY_CHECKS; i++) {
```

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,240 @@
...
+        const deployment = await appsV1Client.readNamespacedDeployment({ name: deploymentName, namespace });
+        if (
+            deployment.status?.readyReplicas === deployment.spec?.replicas &&
+            (deployment.status?.readyReplicas ?? 0) > 0
+        ) {
+            podsReady = true;
+            console.log(`Deployment is ready with ${deployment.status?.readyReplicas} replicas`);
+            break;
+        }
+        await setTimeout(1000);
```
davidgamero:
```suggestion
        await setTimeout(POD_READY_CHECK_POLL_DELAY_MS);
```

src/test/integration/portForward.ts
```diff
@@ -0,0 +1,240 @@
...
+            await appsV1Client.deleteNamespacedDeployment({ name: deploymentName, namespace });
+            await coreV1Client.deleteNamespacedService({ name: serviceName, namespace });
+        } catch (error) {
+            console.warn('Cleanup warning:', (error as Error).message);
+        }
+    }
+}
+
+async function testPortForwardConnection(testPort: number, label: string): Promise<boolean> {
+    for (let i = 0; i < 5; i++) {
```
davidgamero: nit similar magic number or comment here

mstruebing:
Thanks @davidgamero I've adjusted the tests as you mentioned to be more modular :)

cjihrig (approved):
/lgtm
/approve

PR #2813, opened 2026-03-26
Undici migration p2

davidgamero:
node-fetch removals
- rips out node-fetch and @types/node-fetch, uses native fetch (undici) everywhere
- drops nock, migrates all test mocking to undici MockAgent
- rewrite health, log, watch to use fetch() + Readable.fromWeb() for streaming
- rename integration_test.ts → fullrequest_test.ts since it's a unit test

auth
- removes applyToFetchOptions from kubeconfig (dead after migration)
- cleans up dead tls config in applySecurityAuthentication (was set then immediately overwritten)

watch
- treat watch TimeoutError as reconnectable in informer, add exponential backoff on reconnect
- normalize timeout errors in watch.ts (fixes Node 23 mid-stream AbortError vs TimeoutError inconsistency)
- add integration test for watch

informer 
- add informer reconnect integration test


cjihrig:
This is really great progress on modernizing the project. Thanks for working on it.

davidgamero:
> This is really great progress on modernizing the project. Thanks for working on it.

thanks! i was trying to remove nock so we can consolidate testing into undici mockagent and it snowballed

src/cache_test.ts
```diff
@@ -1518,7 +1523,134 @@ describe('ListWatchCache', () => {
...
+                fakeWatch.watch(mock.anything(), mock.anything(), mock.anything(), mock.anything()),
+            ).thenCall(() => {
+                watchCalls++;
+                resolve(new AbortController());
+                return Promise.resolve(new AbortController());
+            });
+        });
+
+        const cache = new ListWatch('/some/path', mock.instance(fakeWatch), listFn);
+        (cache as any).delayFn = (ms: number) => {
```
davidgamero: Will look for a better way than this `as any`

src/test/integration/watchPods.ts
```diff
@@ -0,0 +1,92 @@
...
+                setTimeout(() => reject(new Error('Timed out waiting for DELETED event')), 15000),
+            ),
+        ]);
+
+        const deleteEvent = receivedEvents.find((e) => e.type === 'DELETED' && e.name === podName);
+        assert.ok(deleteEvent, 'Should have received DELETED event for pod');
+        console.log('✓ Received DELETED event');
+    } finally {
+        controller.abort();
+        try {
```
davidgamero, line 75: Todo: extract this cleanup logic across integration tests

src/test/integration/informerReconnect.ts
```diff
@@ -0,0 +1,129 @@
...
+        console.log(`Creating configmap ${cm1Name}`);
+        await coreV1Client.createNamespacedConfigMap({
+            namespace,
+            body: {
+                metadata: { name: cm1Name, labels: { [labelKey]: labelValue } },
+                data: { key: 'value1' },
+            },
+        });
+
+        for (let i = 0; i < 30; i++) {
```
brendandburns: I don't love this since prone to flakiness, I'm not sure if there is a better way to do it in Javascript though. Ideally we would use an event/await here.

src/test/integration/informerReconnect.ts
```diff
@@ -0,0 +1,129 @@
...
+            if (addedNames.includes(cm1Name)) break;
+            await setTimeout(500);
+        }
+        assert.ok(addedNames.includes(cm1Name), 'Should have received add event for cm1');
+        console.log('✓ Received add event for cm1');
+
+        const initialConnects = connectCount;
+        console.log(`Waiting for watch reconnection (up to 45s)...`);
+        for (let i = 0; i < 90; i++) {
```
brendandburns: As above wrt flakiness from sleep

src/test/integration/informerReconnect.ts
```diff
@@ -0,0 +1,129 @@
...
+        console.log(`Creating configmap ${cm2Name}`);
+        await coreV1Client.createNamespacedConfigMap({
+            namespace,
+            body: {
+                metadata: { name: cm2Name, labels: { [labelKey]: labelValue } },
+                data: { key: 'value2' },
+            },
+        });
+
+        for (let i = 0; i < 30; i++) {
```
brendandburns: Here too.

src/config.ts
```diff
@@ -308,14 +284,6 @@ export class KubeConfig implements SecurityAuthentication {
 
         await this.applyOptions(httpsOptions);
 
-        if (cluster && cluster.skipTLSVerify) {
```
brendandburns, line 311: Should we be deleting this?

src/watch_test.ts
```diff
@@ -181,91 +190,6 @@ describe('Watch', () => {
         strictEqual(doneCalled, 1);
     });
 
-    it('should call setKeepAlive on the socket to extend the default of 5 mins', async (t) => {
```
brendandburns, line 184: Does this test no longer apply?

brendandburns:
Generally looks good, some comments on sleep() usage in one of the tests and a couple of things that I'm not sure why they were removed.

src/test/integration/informerReconnect.ts
cjihrig: You could create a Promise that you await here. The Promise would be resolved in the informer's add event. Alternatively, I believe you could await the add event directly as long as it's only supposed to fire for the data we care about.

davidgamero: added promise and abstracted withTimeout() to handle timing

```ts
// helper
export async function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promise<T> {
    return Promise.race([
        promise,
        new Promise<T>((_, reject) => setTimeout(() => reject(new Error(message)), ms)),
    ]);
}

// before
for (let i = 0; i < 30; i++) {
    if (addedNames.includes(cm1Name)) break;
    await setTimeout(500);
}

// refactored
await withTimeout(cm1AddPromise, 15000, 'Timed out waiting for cm1 add event');
```

src/test/integration/watchPods.ts
davidgamero: not sure an abstraction to save 4 lines is worth it here. closing for now

src/cache_test.ts
davidgamero: refactored to promise

src/watch_test.ts
davidgamero: it appears undici default keepalive is 60s which could still cause a race condition on some load balancers.

reintroduced updated test in config to keep 30s keepalive

src/config.ts
davidgamero: my reasoning is that  the `applyOptions` call above invokes `applyHTTPSOptions`, which sets httpsOptions.rejectUnauthorized

this is mirrored into `agentOptions` below by `agentOptions.rejectUnauthorized = httpsOptions.rejectUnauthorized;`

servername is similarly applied in applyOptions > applyHTTPSOptions, but was missing the `httpsOptions -> agentOptions` conversion in this function which i've now added below along with a test calling the whole chain to assert the TLS propagation

happy to reasses my interpretation or refactor to improve readability

src/cache.ts
```diff
@@ -23,32 +23,45 @@ export interface ObjectCache<T> {
...
         labelSelector?: string,
         fieldSelector?: string,
+        options?: ListWatchOptions,
     ) {
         this.path = path;
         this.watch = watch;
         this.listFn = listFn;
         this.labelSelector = labelSelector;
         this.fieldSelector = fieldSelector;
+        this.delayFn = options?.delayFn ?? ((ms) => new Promise((resolve) => setTimeout(resolve, ms)));
```
cjihrig: My only real comment is that you could use the `node:timers/promises` version of `setTimeout()` in a few places in this PR if you want, but it's optional.

cjihrig (approved):
This LGTM. I'll hold off on merging in case others want to weigh in first.

src/util.ts
```diff
@@ -150,10 +150,8 @@ export function totalForResource(pod: V1Pod, resource: string): ResourceStatus {
     return new ResourceStatus(reqTotal, limitTotal, resource);
 }
 
-// There is a disconnect between the ApiException headers and the response headers from fetch
+// There is a disconnect between the ApiException headers and the response headers from fetch.
 // ApiException expects { [key: string]: string } whereas some fetch implementations provide: { [key: string]: string[] }
-// https://github.com/node-fetch/node-fetch/issues/783
-// https://github.com/node-fetch/node-fetch/pull/1757
 export function normalizeResponseHeaders(response: { headers: { entries(): Iterable<[string, string]> } }): {
```
mstruebing, line 155: Is this function still needed with undici? I remember adding it specifically for `node-fetch`.

davidgamero: we still need to convert from [fetch Headers type ](https://github.com/nodejs/undici/blob/6824eedbba0602d26860990acb187258e9c88a78/types/fetch.d.ts#L93)

this is only use in the hand written code parts. updating generator to use fetch Headers type would close the need for this, and improve compat. updated comment to specify

mstruebing:
@davidgamero can/should we merge this?

davidgamero:
> @davidgamero can/should we merge this?

yes all comments are addressed

in my view it is good to merge @mstruebing

mstruebing (commented):
/lgtm
/approve

PR #2814, opened 2026-03-27
pin actions to sha

davidgamero:
to follow guidance from https://github.com/kubernetes/community/blob/main/github-management/github-actions-policy.md

added [zizmor](https://github.com/zizmorcore/zizmor) lint step validating SHAs exist and match

if people would prefer to hold off on the lint step i can reduce PR scope to just the pinned SHAs, i added linting to give some way to validate the SHAs are correct

> All GitHub Actions MUST be referenced using commit SHA hashes.

using zizmor in the meantime to lint. this will likely be applied via repo/org settings in the future and we can drop that dependency from the test workflow

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,15 +5,21 @@ updates:
       schedule:
           interval: 'daily'
           time: '02:00'
+      cooldown:
+          default-days: 7
       target-branch: 'main'
     - package-ecosystem: 'github-actions'
       directory: '/'
       schedule:
           interval: 'daily'
           time: '03:00'
+      cooldown:
+          default-days: 7
       target-branch: 'main'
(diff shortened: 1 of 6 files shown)


brendandburns:
This looks good to me, thanks for doing it!

I'll hold off merging for another day or two to give others a chance to look

mstruebing:
Looks good to me as well, I will not trigger prow right now for the same reason as Brendan.

cjihrig (approved):
LGTM as well. I'll merge it as I'm not sure there is anyone else to wait on reviewing 😄 

/lgtm
/approve

PR #2816, opened 2026-04-05
fix: merge global agent CA certificates with cluster-specific ones

### LINKED ISSUE #2546: System certificates ignored
**Describe the bug**

System certificates configured through global agent are ignored by the kubernetes-client.

In Podman Desktop, as nodejs is not using the system certificates, we load them and make them available to various https libraries through the global agent.
But the kubernetes-client is providing its own agent to the node-fetch library and everything configured through the global agent is not taken.
So if we have to connect to a cluster with a certificate generated by a private certificate authority, it is failing but working fine in kubectl.

I can see 2 solutions:

- merge of global agent is done in the kubernetes-client library
- merge of global agent is done outside the kubernetes-client library (overriding applyToHTTPSOptions ?)

**Client Version**
`1.3.0`

**Server Version**
N/A

**To Reproduce**

**Expected behavior**

Connect to a cluster with private certificate (kind ?)

**Example Code**

N/A

- OS: Ubuntu
- Node.js version 22
- Cloud runtime N/A

**Additional context**

cyphercodes:
## Description

When system certificates are configured through the global https.agent, they were being ignored by the kubernetes-client because it provides its own dispatcher to undici. This PR merges the global agent's CA certificates with any cluster-specific CA certificates, ensuring that custom CAs configured via the global agent are respected.

## Changes

- Modified `createDispatcherOptions` in `src/config.ts` to:
  - Read CA certificates from `https.globalAgent.options.ca`
  - Merge global CA with cluster-specific CA by concatenating them
  - Handle both Buffer and string CA formats, as well as arrays

## Testing

Added two new test cases:
1. `should merge global agent CA with cluster CA` - verifies that when both global and cluster CAs are present, they are concatenated
2. `should use global CA when no cluster CA is provided` - verifies that global CA is used when no cluster CA is specified

## Related Issue

Fixes #2546

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -648,7 +648,23 @@ export class KubeConfig implements SecurityAuthentication {
         agentOptions: https.AgentOptions,
     ): DispatcherOptions {
         const tlsOptions: tls.ConnectionOptions = {};
-        if (agentOptions.ca !== undefined) tlsOptions.ca = agentOptions.ca;
+        // Merge global agent CA certificates with cluster-specific ones
+        const globalCA = (https.globalAgent as https.Agent).options?.ca;
+        if (globalCA !== undefined) {
+            tlsOptions.ca = globalCA;
+        }
+        if (agentOptions.ca !== undefined) {
+            // If both global and cluster CA exist, concatenate them
+            if (tlsOptions.ca !== undefined) {
+                const globalCAs = Array.isArray(tlsOptions.ca) ? tlsOptions.ca : [tlsOptions.ca];
+                const clusterCAs = Array.isArray(agentOptions.ca) ? agentOptions.ca : [agentOptions.ca];
+                tlsOptions.ca = Buffer.concat(
+                    [...globalCAs, ...clusterCAs].map((ca) => (Buffer.isBuffer(ca) ? ca : Buffer.from(ca))),
+                );
+            } else {
+                tlsOptions.ca = agentOptions.ca;
+            }
+        }
         if (agentOptions.cert !== undefined) tlsOptions.cert = agentOptions.cert;
         if (agentOptions.key !== undefined) tlsOptions.key = agentOptions.key;
         if (agentOptions.pfx !== undefined) tlsOptions.pfx = agentOptions.pfx;
--- a/src/config_test.ts
+++ b/src/config_test.ts
@@ -565,6 +565,64 @@ describe('KubeConfig', () => {
                 message: 'Unsupported proxy type',
             });
         });
+        it('should merge global agent CA with cluster CA', async () => {
+            const kc = new KubeConfig();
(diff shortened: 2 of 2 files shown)


src/config.ts
```diff
@@ -648,7 +648,25 @@ export class KubeConfig implements SecurityAuthentication {
         agentOptions: https.AgentOptions,
     ): DispatcherOptions {
         const tlsOptions: tls.ConnectionOptions = {};
-        if (agentOptions.ca !== undefined) tlsOptions.ca = agentOptions.ca;
+        // Merge global agent CA certificates with cluster-specific ones
+        const globalCA = (https.globalAgent as https.Agent).options?.ca;
+        if (globalCA !== undefined) {
+            tlsOptions.ca = globalCA;
```
brendandburns, line 654: What is the behavior if tlsOptions.ca is null/undefined? Wouldn't it fall back on the global CA anyway? Why is this necessary?

src/config.ts
```diff
@@ -648,7 +648,25 @@ export class KubeConfig implements SecurityAuthentication {
...
+        }
+        if (agentOptions.ca !== undefined) {
+            // If both global and cluster CA exist, concatenate them
+            if (tlsOptions.ca !== undefined) {
+                const globalCAs = Array.isArray(tlsOptions.ca) ? tlsOptions.ca : [tlsOptions.ca];
```
brendandburns, line 659: What is the use case for this? If there is a CA in the kubeconfig, that should be the CA for the server, the global CAs shouldn't be necessary.

cyphercodes:
I updated the commit message to remove the forbidden closing keyword/# mention. On the CA question: this client builds its own undici dispatcher, so it does not inherit `https.globalAgent.options.ca`; leaving `tlsOptions.ca` unset falls back to Node defaults rather than app-provided global CAs. Merging preserves kubeconfig CAs while also allowing private/system CAs loaded by an embedding app. Remaining blocker: EasyCLA still reports my CLA is not authorized, so I need to complete the EasyCLA signing link.

cyphercodes:
I pushed a small follow-up that only applies Prettier to `src/config.ts` so lint passes. I verified `npm run lint`, `npm run build`, `npm test`, and the focused `src/config_test.ts` run locally.

On the CA question: this path builds an undici dispatcher, so it does not inherit `https.globalAgent.options.ca`; merging keeps kubeconfig CAs while also honoring private/system CAs an embedding app has loaded globally. The remaining blocker is EasyCLA still reporting my CLA authorization as missing, so I need to complete the EasyCLA signing link.

brendandburns:
I'm ok with the first part of the change, but I don't think that we should merge CAs. If there is a CA in the kubeconfig file, that's the only CA we should load, if there is no CA in the kubeconfig then it seems ok to fall back to the node-default or OS default CA.

PR #2840, opened 2026-04-25
Drop Node 20 support

mstruebing:
Drop Node 20 support as it's EOL in a couple of days and undici also dropped support already (#2835, https://github.com/nodejs/undici/blob/main/package.json#L135-L137)

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -25,7 +25,7 @@ jobs:
             - name: Setup Node.js
               uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
               with:
-                  node-version: '20'
+                  node-version: '22'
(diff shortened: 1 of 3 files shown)


brendandburns:
Let's add Node 26 and drop Node 25 and 23?

mstruebing:
Node 26 is not yet released, we would need to wait for a couple of days it should be released in April 26 as of: https://nodejs.org/en/blog/announcements/evolving-the-nodejs-release-schedule

I've removed Node 23.

I will wait until 26 is released to drop Node 25 and add Node 26.
I can do it either as a followup or we wait with the merge.

EDIT: actually seems like it will be released tomorrow: https://github.com/nodejs/node/issues/61832

brendandburns:
/lgtm
/approve

Looks like node 26 got delayed til 5/4 let's pick it up in a different PR.

Issue #2845, opened 2026-05-05
Release 2.0.0 planning


davidgamero:
proposal to release a client version 2.0.0-rc1 to gather initial feedback as the next release will the the first after full migration to undici from node-fetch

need someone with perms once we are aligned on plan to trigger workflows 

k8s v1.36 [is out](https://kubernetes.io/releases/1.36/), and we don't have a version for k8s v1.35 release yet.

proposing an rc generated on k8s v1.35, then a full release of client v2.0.0 (barring any major rc issues) before advancing to a k8s v1.36 compatible version (client v2.1.0) in the future.

looking for input to move forward

ref #2650

@mstruebing @cjihrig @brendandburns

cjihrig:
This seems like a fine plan to me.

> need someone with perms once we are aligned on plan to trigger workflows

You've been doing a ton of work lately. Maybe you should be onboarded as a maintainer again?

brendandburns:
Works for me.

mstruebing:
Sounds good.
I think everyone can do releases.

davidgamero:
sure i've opened https://github.com/kubernetes/org/pull/6369 to onboard

davidgamero:
prerelease is out https://www.npmjs.com/package/@kubernetes/client-node/v/2.0.0-rc.1

Emut:
Is there already a change log for 2.0.0?
Want to give it a try, but would be nice to know what to expect 😬

davidgamero:
@Emut

the most important is the node-fetch to undici migration and major node version changes

current 2.0.0 draft changelog:
```

### Breaking Changes

- **HTTP backend migrated from `node-fetch` to `undici`.** `undici` is now the
  underlying `fetch` implementation. The `node-fetch` dependency has been
  removed. Code relying on `node-fetch`-specific `Agent`/request internals or
  proxy wiring may need updates; the client now configures an undici
  `Dispatcher` instead of a `node-fetch` agent. (#2306)
- **Dropped Node.js 18 support.** Node 18 removed from the test matrix (EOL Apr 30, 2025).
- **Dropped Node.js 20 support.**
- **Dropped Node.js 23 support.**
- **`socks-proxy-agent` upgraded 8.x → 10.x** and `socks` added as a direct
  dependency for the undici SOCKS proxy path.
- **`undici` pinned to `^8.0.0`.**

### Features

- Add SOCKS proxy support on the undici dispatcher path via a `SocksClient` connector.
[...]

PR #2856, opened 2026-05-15
allow workflow to write PRs

mstruebing:
I've tried to regen with kubernetes release-1.35 but it seems like we can not create a PR from the workflow:
https://github.com/kubernetes-client/javascript/actions/runs/25906285452/job/76140710609#step:7:14

We may need to additionally check this box, but I don't know how
Screenshot 2026-05-15 at 09 45 01

This was done in preparation of #2845


cjihrig (approved):
/lgtm
/approve

PR #2857, opened 2026-05-15
Sync gen-commits

mstruebing:
Noticed a difference in the gen-commits used.

https://github.com/kubernetes-client/javascript/blob/main/settings#L18


cjihrig (approved):
/lgtm
/approve

Issue #2908, opened 2026-06-23
Build Docs is failing; published docs are out of date


jarofgreen:
https://www.npmjs.com/package/@kubernetes/client-node links to https://kubernetes-client.github.io/javascript/

These docs are now old, and there are some big changes - for instance https://kubernetes-client.github.io/javascript/classes/CustomObjectsApi.html#listNamespacedCustomObject tells you to pass the required parameters separately, but now in the latest version of the library you must pass them as one options object.

The badge reports the builds as failing: https://github.com/kubernetes-client/javascript/actions/workflows/deploy-docs.yml The error seems to be "Required status check "EasyCLA" is expected."

Can the docs be updated?

Thanks,

jarofgreen:
I've realised that's not the best example after building the doc's locally and seeing that the parameters are still shown separately, but I'm pretty sure the general point still stands.

mstruebing:
This is definitely an issue. 
I've tried to fix it a while ago but if I remember correctly some global applied branch protections and/or easy CLA were in the way.
We didn't wanted to remove the job back then but didn't fix it yet.

KumarNirupam1:
hi @jarofgreen  @mstruebing I'd like to work on this issue

mstruebing:
@KumarNirupam1 This is currently already in progress in #2911

davidgamero:
updated docs are live

docs now include category api labels, and links from examples to mentioned types

always welcome to contributions to enhance them @KumarNirupam1 they use docusaurus now

jarofgreen:
Thank you!

PR #2911, opened 2026-06-24
Upgrade Docs to Docusaurus

davidgamero:
- move generated docs to lead with validated examples
- improved search by verb and api
- categorize and link between doc pages by API
- automatically detect and link relevant example snippets in docs

~~will attach generated output to test on my fork~~
deployed live version on fork https://davidgamero.github.io/javascript/

the current publish pipeline uses the gh-pages branch which gets blocked by CLA check. we can change repo to use "GH Action" method for gh pages publishing to resolve this issue

addressing #2908

recognizing line count is still large. package lock is 19k making diff look much bigger

note: I used opencode for chunks of this migration


.github/workflows/deploy-docs-v2.yml
```diff
@@ -0,0 +1,36 @@
...
+        branches:
+            - main
+permissions:
+    contents: write
+jobs:
+    build-and-deploy-docs:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v6.0.1
```
brendandburns: Use SHA not version (new k8s policy)

brendandburns:
Looks like there are some bad links, so CI/CD is failing.

davidgamero:
updated to link generated docs dir so absolute links can be validated

mstruebing:
> the current publish pipeline uses the gh-pages branch which gets blocked by CLA check. we can change repo to use "GH Action" method for gh pages publishing to resolve this issue

@davidgamero can you elaborate how this works? What is the "GH Action" method?

davidgamero:
> @davidgamero can you elaborate how this works? What is the "GH Action" method?

there are 2 ways to publish to github pages static sites:
- from a branch: the contents of a particular branch is served as static site content. we do this now
- from a github workflow: a github workflow publishes using the [actions/deploy-pages](https://github.com/actions/deploy-pages) github action

[Configuring a publishing source for your GitHub Pages site](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)

current flow is as follows
1. PR with docs changes merges to main, triggering [.github/workflows/deploy-docs.yml](https://github.com/kubernetes-client/javascript/blob/38186ddc9984d90211912690e67126e38516817b/.github/workflows/deploy-docs.yml)
2. deploy-docs.yml workflow updates the contents of the `gh-pages` branch, which is reflected in live github.io hosting content

step 2 is blocked by the CLA check or branch protections mentioned in #2908

the new flow replaces this step with the [actions/deploy-pages](https://github.com/actions/deploy-pages) github action to publish without using a branch to store the contents, instead a github artifact is uploaded with the static site content.

as no branch push is required, the CLA and branch rules no longer block the doc publishing. additionally, the docs are generated from the types during the workflow run, so the full docs content is not pushed in source control every time, it instead exists as a function of the source, and in an artifact

this requires toggling a setting on the repo for under "Pages" 
image

davidgamero:
@mstruebing i can break this into a smaller poc to show the doc gen capability to make reviewing easier, but most of the diff is the packagelock as is

mstruebing (commented):
In general this LGTM to me, I'm just a bit afraid of all the extra code we would need to maintain (the one in `scripts`).
But as our current approach isn't working at all and this is in general a good solution I'm totally fine with it. Will add the lgtm for now so others could still have a look.
If no one does I will approve tomorrow morning.

/lgtm

PR #2941, opened 2026-07-15
group codeql actions for dependabot

mstruebing:
We've had this a couple of times already that dependabot updates all three code ql actions in separate PRs but they only work together. This should let dependabot group them together in a single pr.


cjihrig (approved):
/lgtm
/approve

Issue #3007, opened 2026-08-12
allow copilot via CLA


davidgamero:
gathering consensus on enabling the linux foundation CLA checks to allow GitHub Copilot

currently the github copilot UI is shown on this repo for "making edits" and "assign tasks" yet all copilot PRs are blocked by CLA and require force push to remove copilot from commit history, as many PRs have encountered

example: https://github.com/kubernetes-client/javascript/pull/2904

following guidance from https://docs.linuxfoundation.org/lfx/easycla/v2-current/allowing_bots_to_skip_cla

i'm prepared to coordinate the request with EasyCLA, would like to gather views on this from maintainers/contributors

cjihrig:
+1 on enabling it.

mstruebing:
+1 for sure

brendandburns:
+1

davidgamero:
fixed in #3010 the solution was to remove copilot commit attribution to comply with CLA

PR #3010, opened 2026-08-14
disable copilot coauthor commit info following steering committee guidance

davidgamero:
disable copilot coauthor commit info in alignment with steering committee guidelines on AI:
>Listing AI tooling as a co-author, co-signing commits using an AI tool, or using the assisted-by, co-developed or similar commit trailer is not allowed.

https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance

this change is intended so that the author of the PR, who must have signed the CLA, takes full commit credit for their work even when using copilot tools. this does not address PRs initiated by an agent, in which there isn't an author, as those will still fail CLA and are being discussed by steering committee for potential policy changes

this should unblock PRs where someone uses copilot to apply changes where it previously caused the CLA to trigger for the "copilot" committer

ref #3007

### DIFF
--- a/.github/copilot/settings.json
+++ b/.github/copilot/settings.json
@@ -1,0 +1,3 @@
+{
+    "includeCoAuthoredBy": false
+}


mstruebing (commented):
/lgtm
/approve

PR #3022, opened 2026-08-24
fix: use inactivity timeout for watch instead of whole-request timeout

### LINKED ISSUE #3020: Watch aborts healthy connections after 30s (2.0.0)
**Describe the bug**

`Watch` kills a working watch after 30 seconds, even while events are still coming in.

The relevant lines are [`src/watch.ts#L42-L44`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/watch.ts#L42-L44), with the timeout value at [`src/watch.ts#L11`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/watch.ts#L11):

```ts
const controller = new AbortController();
const timeoutSignal = AbortSignal.timeout(this.requestTimeoutMs); // 30_000
const signal = AbortSignal.any([controller.signal, timeoutSignal]);
```

and that signal is passed to the fetch that streams the watch body, [`src/watch.ts#L63-L68`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/watch.ts#L63-L68).

`AbortSignal.timeout` starts counting the moment the watch starts, and `fetch` keeps that signal active for as long as it's reading the response body. So it fires 30 seconds in regardless of whether anything is wrong with the connection, and `done` is called with a `TimeoutError`.

I might be reading the intent wrong. #2367 added this 30 seconds as `requestInit.timeout`, which in node-fetch meant a timer that only fired when nothing was being received. I expected that meaning to have carried over, so a healthy watch would be left alone.

**Client Version**

`2.0.0`

**Server Version**

`1.35` (GKE)

**To Reproduce**

Start a watch on any resource and keep the connection healthy. `done` is called after 30 seconds with `TimeoutError: The operation was aborted due to timeout`, and the connection is closed from the client side.

Against a real cluster, watching `/api/v1/pods` on a busy namespace set — the event counter is still climbing when the client hangs up:

```
   0.26s watch started, stream is open
   0.26s event ADDED argo-events/argo-events-controller-manager-...
   0.35s event ADDED argo-rollouts/argo-rollouts-...
   ...
  10.01s ... 329 events received so far
  20.01s ... 340 events received so far
  30.02s DONE  TimeoutError: The operation was aborted due to timeout
  30.02s ... 349 events received so far
  40.02s ... 349 events received so far
  ...
  80.03s ... 349 events received so far
```

Nothing on my side aborted it, and the event count is frozen from 30.02s on because there is no longer a watch.

I also checked it against a fake server that sends one event and then deliberately never ends the response, so that nothing except the client is able to close the connection:

```
 0.01s server: watch connection opened
 0.02s client: event ADDED p1
30.02s client: done(TimeoutError: The operation was aborted due to timeout)
30.02s server: connection closed by client
```

On `1.4.0` the same test keeps the connection for as long as I leave it running and `done` is never called — but that isn't the behaviour I'm asking for either, since 1.4.0 never notices a genuinely dead connection. I'm expecting something in between: leave a working connection alone, hang up on a silent one.

**Expected behavior**

A watch on a working connection stays open until the server closes it or I abort it myself.

**Example Code**

The `done` callback is what shows it:

```js
const t0 = Date.now();
const at = () => ((Date.now() - t0) / 1000).toFixed(2) + 's';

const kc = new KubeConfig();
kc.loadFromDefault();

await new Watch(kc).watch(
  '/api/v1/pods',
  {},
  (phase, obj) => console.log(at(), 'event', phase, obj.metadata.name),
  // fires at ~30s with a TimeoutError, even on a connection that is fine
  (err) => console.log(at(), 'done', err ? `${err.name}: ${err.message}` : '(no error)'),
);
```

**Environment**

- OS: macOS (host) and Linux (`node:24-alpine` container)
- Node.js version: 26.7.0 against the real cluster, 24.19.0 for the fake-server runs
- Cloud runtime: none — plain process / container

**Additional context**

This also affects `ListWatch`. Its retry delay is only reset when an event arrives, in [`src/cache.ts#L267`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/cache.ts#L267) inside `watchHandler`, while the delay itself is applied on every reconnect in [`src/cache.ts#L207-L215`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/cache.ts#L207-L215). So on a resource where nothing ever happens, the delay keeps doubling until it reaches [`MAX_RECONNECT_DELAY_MS`](https://github.com/kubernetes-client/javascript/blob/f72cc23ed378cb8e7f09129ee6e55aa531a2b9ba/src/cache.ts#L33). Time between watch connections, measured against the fake server with no events:

```
30s → 31s → 32s → 34s → 38s → 46s → 60s
```

It settles at roughly 30 seconds watching, then 30 seconds not watching. With an event every 10 seconds it stays at a steady 30 seconds. This matters for quiet resources — in my case a `ListWatch` over CRDs, which can go a long time with no events.

I'm happy to send a PR if this is real. For the timeout, I think it needs to become a timer that's reset each time a line is received and cleared when the watch finishes, rather than one that runs for the whole request — that's a few lines and wants a test, so I didn't want to guess at it here. Separately, resetting `reconnectDelayMs` once the connection is established, instead of when the first event arrives, would stop quiet resources from backing off.

duizabojul:
Fixes #3020

**What / why**

`Watch` passes `AbortSignal.timeout(requestTimeoutMs)` to the fetch that streams the watch body, so the signal fires 30s after the request starts regardless of connection health, and every watch — including one actively receiving events — is aborted with a `TimeoutError`.

Per the discussion in #3020, this PR keeps timeout signaling but changes what it measures:

- **`src/watch.ts`** — replaces the whole-request `AbortSignal.timeout` with a resettable inactivity timer. The timer is armed before the fetch (connect timeout, same as today), then re-armed on the 200 response and on every body data chunk. A watch that keeps receiving data is left alone; a connection that goes silent for `requestTimeoutMs` (or never connects) is still aborted with the same `TimeoutError` as before. The timer is `unref()`ed to match `AbortSignal.timeout`'s event-loop behavior, and is cleared once the watch finishes. No public API change; `requestTimeoutMs` keeps its name and 30s default, its meaning becomes "max time with no data received".
- **`src/cache.ts`** — `ListWatch` now resets `reconnectDelayMs` in the existing `TimeoutError` branch of `doneHandler`. A client-side timeout means the client already waited the full `requestTimeoutMs`, so an immediate reconnect cannot tight-loop — the timeout itself is the throttle. Without this, a watch on a quiet resource (which now times out every `requestTimeoutMs` by design) accumulates exponential backoff and ends up alternating ~30s watching / ~30s not watching. Backoff is unchanged for real errors and server-side disconnects.

**Tests**

- watch: a stream that keeps sending events outlives the timeout (lives ~2.4× `requestTimeoutMs`, then clean server close → `done(null)`)
- watch: a connection that goes idle after connecting (one event, then silence) still ends with `TimeoutError`
- watch: the pre-existing connect-timeout test passes unchanged
- cache: repeated `TimeoutError` reconnects apply no delay, while a subsequent non-timeout reconnect still backs off (via the injectable `delayFn`)

Both new watch tests and the cache test fail against the previous implementation.

`npm test` (356/356), `npm run lint`, `npm run build` all pass.

### DIFF
--- a/src/watch.ts
+++ b/src/watch.ts
@@ -40,8 +40,25 @@ export class Watch {
         }
 
         const controller = new AbortController();
-        const timeoutSignal = AbortSignal.timeout(this.requestTimeoutMs);
-        const signal = AbortSignal.any([controller.signal, timeoutSignal]);
+
+        let timedOut: boolean = false;
+        let timer: NodeJS.Timeout | undefined;
+        const clearTimer = () => {
+            if (timer !== undefined) {
+                clearTimeout(timer);
+                timer = undefined;
+            }
+        };
+        const resetTimer = () => {
+            clearTimer();
+            timer = setTimeout(() => {
+                timedOut = true;
+                controller.abort(
+                    new DOMException('The operation was aborted due to timeout', 'TimeoutError'),
+                );
+            }, this.requestTimeoutMs);
+            timer.unref();
+        };
 
         const ctx = new RequestContext(watchURL.toString(), HttpMethod.GET);
         await this.config.applySecurityAuthentication(ctx);
@@ -50,8 +67,9 @@ export class Watch {
         const doneCallOnce = (err: any) => {
             if (!doneCalled) {
                 doneCalled = true;
+                clearTimer();
                 controller.abort();
-                if (err && timeoutSignal.aborted) {
+                if (err && timedOut) {
                     done(new DOMException('The operation was aborted due to timeout', 'TimeoutError'));
                 } else {
                     done(err);
@@ -60,14 +78,17 @@ export class Watch {
         };
 
         try {
+            resetTimer();
             const response = await fetch(watchURL, {
                 method: 'GET',
                 headers: ctx.getHeaders(),
                 dispatcher: ctx.getDispatcher(),
-                signal,
+                signal: controller.signal,
             });
 
             if (response.status === 200) {
+                // The connect timeout is over; from here on it becomes an inactivity timeout.
+                resetTimer();
                 const body = Readable.fromWeb(response.body! as any);
 
                 body.on('error', doneCallOnce);
@@ -86,6 +107,7 @@ export class Watch {
                         // ignore parse errors
                     }
                 });
+                body.on('data', resetTimer);
             } else {
                 const statusText =
                     response.statusText || STATUS_CODES[response.status] || 'Internal Server Error';
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -168,7 +168,11 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
         ) {
             this.resourceVersion = '';
         } else if (err && (err as { name?: string }).name === 'TimeoutError') {
-            // Watch client-side timeout — reconnect from last known resourceVersion
+            // Watch client-side timeout — reconnect from last known resourceVersion.
+            // The timeout itself already throttled us for the full request timeout, so
+            // reconnecting immediately cannot tight-loop and backing off would leave
+            // quiet resources unwatched for up to MAX_RECONNECT_DELAY_MS.
+            this.reconnectDelayMs = 0;
         } else if (err) {
             this.callbackCache[ERROR].forEach((elt: ErrorCallback) => elt(err));
             return;
--- a/src/cache_test.ts
+++ b/src/cache_test.ts
@@ -1639,7 +1639,7 @@ describe('ListWatchCache', () => {
         deepStrictEqual(delayValues, [1000]);
     });
 
-    it('should reconnect with backoff on TimeoutError', async () => {
+    it('should reconnect on TimeoutError', async () => {
         const fakeWatch = mock.mock(Watch);
         const listObj = {
             metadata: { resourceVersion: '12345' } as V1ListMeta,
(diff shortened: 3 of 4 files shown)


davidgamero:
this looks good to me. this exposed a couple other issues to address with watches like lack of jitter on the reconnects which i'll handle in a separate PR.
/lgtm

duizabojul:
Worth also looking at what client-go does here: a randomized timeoutSeconds sent to the server so it ends the watch on its own schedule instead of the client guessing, and watch bookmarks to keep the resourceVersion fresh across restarts.

src/watch.ts
```diff
@@ -40,8 +40,25 @@ export class Watch {
...
+                clearTimeout(timer);
+                timer = undefined;
+            }
+        };
+        const resetTimer = () => {
+            clearTimer();
+            timer = setTimeout(() => {
+                timedOut = true;
+                controller.abort(
+                    new DOMException('The operation was aborted due to timeout', 'TimeoutError'),
```
mstruebing: Should this really be a `DOMException`?

mstruebing (commented):
/lgtm
/approve

PR #3027, opened 2026-08-30
ci: pin the Node 26 matrix entry to 26.7

davidgamero:
mock-fs broke on node 26.8 https://github.com/tschaub/mock-fs/issues/447

proposing to pin tests to 26.7 while working on longer-term fix options

[SERVICEACCOUNT_ROOT](https://github.com/kubernetes-client/javascript/blob/f994a0f923f868b155b63d053d8308c4d681c01e/src/config.ts#L64) tests rely on mock-fs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,10 @@ jobs:
             contents: read
         strategy:
             matrix:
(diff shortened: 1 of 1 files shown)


mstruebing (commented):
/lgtm
/approve

PR #3030, opened 2026-09-01
Add client-go gap analysis and implementation roadmap

brendandburns:
## Summary

- compare the JavaScript client's capabilities with `client-go`
- use the Java client as an additional source of cross-language API designs
- prioritize implementation gaps from P0 through P3
- propose phased delivery, explicit non-goals, and success criteria

The analysis emphasizes controller-runtime foundations such as reliable LIST/WATCH behavior, discovery and REST mapping, dynamic clients, informers, workqueues, retries, and test fakes rather than generated API coverage.

## Validation

- `npx --yes prettier@3.9.6 --check CLIENT_GO_GAP_ANALYSIS.md`
- `git diff --check`

### DIFF
--- a/CLIENT_GO_GAP_ANALYSIS.md
+++ b/CLIENT_GO_GAP_ANALYSIS.md
@@ -1,0 +1,441 @@
+# Client-go Gap Analysis and Implementation Roadmap
+
+## Purpose
+
+This document compares the Kubernetes JavaScript client with
+[`kubernetes/client-go`](https://github.com/kubernetes/client-go) and uses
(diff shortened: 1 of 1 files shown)


cjihrig:
Is the idea to check this generated file into git, or is this intended more as a tracking issue?

FWIW, I have ported some of these things from Go in the past, such as [leader election](https://github.com/cjihrig/kubenode/blob/main/controller-runtime/lib/leaderelection/leaderelection.js) and other controller-runtime things. Feel free to take any of that stuff - although in 2026, I guess it's easier to just tell an AI to write one from scratch.

PR #3032, opened 2026-09-01
Fix pretty option for pod log requests

brendandburns:
## Summary

- use `LogOptions.pretty` when setting the `pretty` query parameter for pod log requests
- add regression coverage where `follow` is false and `pretty` is true, ensuring the options are handled independently

This closes the pod logs parity gap identified in the client-go gap analysis.

## Testing

- `npm run build`
- `node --test --test-reporter=spec --import tsx src/log_test.ts`
- repository pre-commit test hooks


cjihrig (approved):
/lgtm
/approve

PR #3033, opened 2026-09-01
Fix pod memory request aggregation

brendandburns:
## Summary

- accumulate each container's memory request from the pod request total rather than the pod limit total
- add regression coverage using multiple containers with distinct memory requests and limits

This addresses the pod memory-request aggregation correctness gap identified in the client-go gap analysis.

## Testing

- `npm run build`
- `node --test --test-reporter=spec --import tsx src/top_test.ts`
- repository push hooks


cjihrig (approved):
/lgtm
/approve

PR #3034, opened 2026-09-02
Retry transient informer failures

brendandburns:
## Summary

- retry transient informer LIST and WATCH failures instead of terminating the informer
- classify network errors, HTTP `408`, `429`, and `5xx` responses as retryable while preserving terminal behavior for other `4xx` errors
- reuse the existing client-go-inspired exponential jittered backoff and preserve timeout, `410 Gone`, and explicit-stop behavior
- add deterministic regression coverage for retryable LIST/WATCH failures and clean informer shutdown

This addresses the P0 informer recovery gap identified in the client-go gap analysis.

## Testing

- `npm run build`
- `node --test --test-reporter=spec --import tsx src/cache_test.ts`
- full repository test suite: 361 passed, 0 failed
- repository push hooks

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -197,8 +197,31 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
         return ms + this.randFn() * ListWatch.BACKOFF_JITTER * ms;
     }
 
+    private errorStatusCode(err: any): number | undefined {
+        const statusCode = err?.statusCode ?? err?.code;
+        return typeof statusCode === 'number' ? statusCode : undefined;
+    }
+
+    private isRetryableError(err: any): boolean {
+        const statusCode = this.errorStatusCode(err);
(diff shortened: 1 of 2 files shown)


src/cache.ts
```diff
@@ -212,25 +235,36 @@ export class ListWatch<T extends KubernetesObject> implements ObjectCache<T>, In
...
-                list = await promise;
-            } catch (err) {
-                this.callbackCache[ERROR].forEach((elt: ErrorCallback) => elt(err));
-                return;
+            let listed = false;
+            while (!listed) {
+                let list: KubernetesListObject<T>;
+                try {
+                    const promise = this.listFn();
+                    list = await promise;
```
davidgamero: was this done for a test?

```suggestion
                    const list = await this.listFn();
```

davidgamero (approved):

PR #3052, opened 2026-09-09
fix: invalidate stale informer retries on stop and restart

davidgamero:
build on #3034

add preventative check to cover the following case:
1. retryable call A fails, starting wait
2. call `stop()` for informer
3. call `start()` for informer, opening watch B
4. A's pending retry finishes wait, checks `stopped === false`  and retries, opening a duplicate watch A

- prevent pending informer retries from surviving a stop/restart cycle and creating duplicate watches
- discards LIST results that arrive after shutdown, preventing cache updates and unwanted reconnections
- use a generation counter to ignore stale callbacks and abort watch connections that complete too late
- add tests covering the race

ai disclosure: opencode + gpt6

### DIFF
--- a/src/cache_test.ts
+++ b/src/cache_test.ts
@@ -1,4 +1,4 @@
-import { describe, it } from 'node:test';
+import { describe, it, type TestContext } from 'node:test';
 import { deepStrictEqual, notStrictEqual, ok, strictEqual, throws } from 'node:assert';
 import mock from 'ts-mockito';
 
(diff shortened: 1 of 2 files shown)


src/cache_test.ts
```diff
@@ -1779,6 +1716,147 @@ describe('ListWatchCache', () => {
         deepStrictEqual(errors, [error]);
     });
 
+    function deferred<T = void>() {
```
cjihrig, line 1719: Can `Promise.withResolvers()` be used here?

cjihrig (approved):
One comment, but LGTM

/lgtm

PR #3056, opened 2026-09-10
test: abstract utility to use Promise.withResolvers for deferred promises

davidgamero:
followup for #3052

implement feedback from @cjihrig

clean up repeated deferred testing function

adopt Promise.withResolvers<T>() [doc link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)

```ts
// old wrapper
function deferred<T = void>() {
        let resolve!: (value: T | PromiseLike<T>) => void;
        const promise = new Promise<T>((done) => (resolve = done));
        return { promise, resolve };
}
```

```ts
// new wrapper
export function deferred<T = void>() {
    return Promise.withResolvers<T>();
}
```

### DIFF
--- a/src/cache_test.ts
+++ b/src/cache_test.ts
@@ -10,6 +10,7 @@ import { ListPromise } from './informer.js';
 
 import { MockAgent, setGlobalDispatcher, getGlobalDispatcher } from 'undici';
 import { Watch } from './watch.js';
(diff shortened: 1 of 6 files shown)


mstruebing (commented):
/lgtm
/approve

